PR #356, opened 2018-08-29
Span detection

pitrack:
@iftenney if we want to do pruning, we would probably change the `index_array` part so that it is not a square matrix.

currently only conll-2003-ner is using "detect-spans", but it would be a small change to switch all the tasks to it. But maybe we want to make the detect-span versions separate tasks (e.g have both "ontonotes-ner" and "ontonotes-ner-detect-spans").


src/tasks.py
```diff
@@ -558,6 +562,47 @@ def get_metrics(self, reset=False):
...
+    Same as EdgeProbingTasks but with span detection
+    '''
+    def __init__(self, path: str, max_seq_len: int,
+                 name: str,
+                 label_file: str=None,
+                 files_by_split: Dict[str,str]=None,
+                 is_symmetric: bool=False,
+                 single_sided: bool=False,
+    ):
+        max_seq_len = 200
```
iftenney: Just make `max_seq_len` just an argument in to the `@register_task` macro, instead of adding boilerplate here?

For efficiency, it might be worth running the [edge_data_stats.py](https://github.com/jsalt18-sentence-repl/jiant/blob/master/probing/edge_data_stats.py) script to get the max length for each task, then setting this explicitly.

PR #396, opened 2018-09-07
SPR1 data & edge-probing task

iftenney:
Script to convert data from JSON files provided by Rachel as used in https://arxiv.org/pdf/1804.07976.pdf. TODO before release to check in a full processing pipeline.

### DIFF
--- a/config/edgeprobe_cove.conf
+++ b/config/edgeprobe_cove.conf
@@ -33,3 +33,4 @@ elmo_chars_only = 1
 // Use no-op encoder (no params).
 sent_enc = "null"
 skip_embs = 1  // forward embeddings from lower level.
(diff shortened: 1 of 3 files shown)


leo-liuzy:
Any access to the SPR1 data?

sleepinyourhat:
This data isn't ours, and I don't think we're allowed to distribute it. The readmes should offer some guidance on where to look.

iftenney:
Hi Leo,

Sorry for the delay - this is a long-closed PR, so notifications don't always show up. Please see documentation here: https://github.com/nyu-mll/jiant/tree/master/probing/data#semantic-proto-roles-spr and reach out via email for the processed data.

- Ian

PR #398, opened 2018-09-07
Add OpenAI Transformer model

iftenney:
Add support for the OpenAI transformer language model (https://blog.openai.com/language-unsupervised/) as a source of representations. Optionally, can fine-tune the model as in their paper.

To enable, set `openai_transformer = 1`. See examples in `scripts/edges/exp_fns.h`.

Use the PyTorch implementation from https://github.com/huggingface/pytorch-openai-transformer-lm; the included notebook under src/openai_transformer_lm runs both the original TF version and this PyTorch version to verify that they produce the same representations.

Note that this model uses a BPE tokenization and so isn't compatible with the Moses tokenization used by the rest of our models. There are a couple asserts that prevent you from using this in conjunction with other embedders that expect a different tokenization, since these would result in different length sequences that are difficult to align. But, no guarantees that these are airtight - it's quite possible to run with the wrong data and get bogus results.

For edge probing, there's a script that projects spans from the original tokenization onto the wordpieces, so that this data (the `edges-*-openai` tasks) is safe to use.


src/preprocess.py
```diff
@@ -277,6 +271,10 @@ def _build_vocab(args, tasks, vocab_path: str):
     vocab = get_vocab(word2freq, char2freq, max_v_sizes)
     for task in tasks:  # add custom label namespaces
         add_task_label_vocab(vocab, task)
+    if args.openai_transformer:
+        # Add pre-computed BPE vocabulary
+        add_openai_bpe_vocab(vocab)
```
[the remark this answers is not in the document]
iftenney: Fixed.

config/defaults.conf
```diff
@@ -177,7 +177,13 @@ cove = 0  // If true, use CoVe.
...
 char_embs = 0  // If true, train char embeddings. This is separate from the ELMo char component, and the two
                // usually aren't used together.
-
+openai_transformer = 0  // If true, use OpenAI Transformer model for
+                        // representations. Not compatible with other
+                        // embeddings due to different tokenization.
+openai_transformer_fine_tune = 0  // If true, fine-tune the OpenAI transformer
+                                  // weights. Note that this might require a
+                                  // high amount of GPU memory, so consider
+                                  // lowering the batch size.
```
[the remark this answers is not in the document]
iftenney: It's an embedding layer, in place of ELMo (see models.py).

iftenney:
Cleaned up a few things, and tested more thoroughly.

sleepinyourhat:
Worth another review?

iftenney:
Yes, please!

sleepinyourhat:
Sure! Running around for the next bit, but should get to it tomorrow (if no
one beats me to it).

config/edgeprobe_openai.conf
```diff
@@ -0,0 +1,35 @@
...
+allow_untrained_encoder_parameters = 1  // allow skipping training phase
+allow_missing_task_map = 1  // ignore missing classifier_task_map.json
+train_for_eval = 1  // train using eval task params
+do_eval = 1
+write_preds = "val,test"
+
+task_patience = 5  // vals until LR decay
+patience = 20      // vals until early-stopping
+
+cove = 0
```
W4ngatang: "Run edge probing models over CoVe" but `cove = 0`?

src/models.py
```diff
@@ -58,7 +58,15 @@ def build_model(args, vocab, pretrained_embs, tasks):
     '''Build model according to args '''
 
     # Build embeddings.
-    d_emb, embedder, cove_layer = build_embeddings(args, vocab, tasks, pretrained_embs)
+    if args.openai_transformer:
+        from .openai_transformer_lm.utils import OpenAIEmbedderModule
+        log.info("Using OpenAI transformer model; skipping other embedders.")
+        cove_layer = None
```
W4ngatang: we should probably refactor the model construction at some point

src/models.py
```diff
@@ -58,7 +58,15 @@ def build_model(args, vocab, pretrained_embs, tasks):
     '''Build model according to args '''
 
     # Build embeddings.
-    d_emb, embedder, cove_layer = build_embeddings(args, vocab, tasks, pretrained_embs)
+    if args.openai_transformer:
+        from .openai_transformer_lm.utils import OpenAIEmbedderModule
+        log.info("Using OpenAI transformer model; skipping other embedders.")
```
W4ngatang: worth adding a warning if using OpenAI transformer but other embeddings are set?

src/modules.py
```diff
@@ -162,9 +162,20 @@ def forward(self, sent, task, reset=True):
             sent_embs = self._dropout(sent_embs)
         if task_sent_embs is not None:
             task_sent_embs = self._dropout(task_sent_embs)
+        max_seq_length = (sent_embs.shape[1]
+                          if sent_embs is not None
+                          else task_sent_embs.shape[1])
 
         # The rest of the model
         sent_mask = util.get_text_field_mask(sent).float()
+        if sent_mask.shape[1] < max_seq_length:
```
W4ngatang: I'm not sure this will ever trigger since `sent_mask` is the same shape as `sent_embs` so its `shape[1]` will be the same as `sent_embs.shape[1]`.

iftenney:
Fixed the clunky padding code. PTAL!

src/preprocess.py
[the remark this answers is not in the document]
W4ngatang: not really an issue to address in this PR, but when running experiments just be careful that the first run in an experiment directory will create the vocabulary, and subsequent runs will just load that vocab, even if it's missing its task-specific vocab

config/defaults.conf
```diff
@@ -413,3 +421,13 @@ edges-constituent-ontonotes = ${edges-tmpl} {
 // These tasks are still very slow. TODO: Debug.
 edges-ccg-tag = ${edges-tmpl}
 edges-ccg-parse = ${edges-tmpl}
+
+// OpenAI transformer versions.
+edges-srl-conll2012-openai = ${edges-srl-conll2012}
```
W4ngatang: just checking: these should all be defined in `user_config.sh` or wherever people are setting their paths?

src/tasks.py
```diff
@@ -559,6 +559,24 @@ def get_metrics(self, reset=False):
...
 
 
+class OpenAIEdgeProbingTask(EdgeProbingTask):
+    """Version of EdgeProbingTask that loads BPE-tokenized data."""
+    @property
+    def tokenizer_name(self):
+        return "OpenAI.BPE"
+
+
+# Register copies of all edge probing tasks.
```
W4ngatang: it's a little weird to me to have executed code this deep into a file that otherwise just looks like class definitions, but seems harmless enough

config/defaults.conf
iftenney: Sorry, HOCON overloads that syntax - these aren't shell variables, they're just inheriting other tuples. So in this case, it's just setting the `edges-srl-conll2012-openai` config to a copy of the `edges-srl-conll2012` config.

src/tasks.py
iftenney: Yeah, I don't love this either but it's just adding entries to the registry. I implemented it as a loop to avoid a (very) large block of boilerplate. Unfortunately, it does have to go at the end of the file since it requires the relevant classes to be defined.

iftenney:
Comment added. Thanks for the review!

PR #406, opened 2018-09-12
updated location of DNC data in README

azpoliak:
Adding the link to download the DNC data that is now publically available


azpoliak:
good catch on that. Let me instead make this change on a new branch directly from master instead of the `docs` branch

PR #417, opened 2018-09-17
rename config args

shuningjin:
Need not be merged immediately

1. see Issue: https://github.com/jsalt18-sentence-repl/jiant/issues/315

renaming 6 arguments
- task_patience -> lr_patience
- do_train-> do_pretrain
- train_for_eval -> do_target_task_training
- do_eval -> do_full_eval
- train_tasks -> pretrain_tasks
- eval_tasks -> target_tasks

2. done by searching through full project and replace (in Atom editor, command+shift+f), so the names get changed whenever they appear; demo run is successful


sleepinyourhat:
Still working on this?

shuningjin:
I have added src/update_params.py for updating old params.conf.
But in the main, it seems to rewrite params.conf every time. When will we load old parameters?
For instance, if I run demo.sh with master halfway, and then reload it with this branch where the params are renamed, it works out.

sleepinyourhat:
The new script looks good, but I'm not sure I understand your question.

If you attempt to restore with an old params.conf, then nothing will crash, but the old parameter values won't be used. (The model will have both do_train and do_pretrain, and it'll ignore do_train and set do_pretrain to default.) Right?

sleepinyourhat:
Thanks! One last thing: Move the script to scripts/ (since it's not part of the main source tree), and add some instructions to the readme explaining both when and how to use it.

main.py
```diff
@@ -94,6 +94,20 @@ def main(cl_arguments):
     cl_args = handle_arguments(cl_arguments)
     args = config.params_from_file(cl_args.config_file, cl_args.overrides)
 
+    # prevent loading old config arg names
```
iftenney: Can you move this block to a helper function?

Also might be simpler to just loop through `name_dict.items()`, like:
```
for old_name, new_name in name_dict.items():
    assert_for_log(old_name not in args, "Error ...")
```

shuningjin:
Moved the assert from main to utils; added README.
Just curious, will the reviewers be automatically notified when the code under review is updated?

scripts/update_config.py
```diff
@@ -0,0 +1,24 @@
...
+# Mapping - key: old name, value: new name
+name_dict = {'task_patience':'lr_patience',\
+               'do_train': 'do_pretrain',\
+               'train_for_eval':'do_target_task_training',\
+               'do_eval': 'do_full_eval',\
+               'train_tasks':'pretrain_tasks',\
+               'eval_tasks':'target_tasks'}
+
+path = sys.argv[1]
+assert '.conf' in path, "Error: .conf not found."
```
iftenney: Is this necessary?

scripts/update_config.py
```diff
@@ -0,0 +1,24 @@
+'''
+Overwrite a config file by renaming args.
+Require one argument: path_to_file.
+'''
+
+import sys
+
+# Mapping - key: old name, value: new name
+name_dict = {'task_patience':'lr_patience',\
+               'do_train': 'do_pretrain',\
```
iftenney: FYI, don't need the backslashes since each line ends with a comma. This should work:
```
name_dict = {'task_patience': 'lr_patience',
             'do_train': 'do_pretrain',
             ...
}
```

scripts/update_config.py
```diff
@@ -0,0 +1,24 @@
...
+               'train_tasks':'pretrain_tasks',\
+               'eval_tasks':'target_tasks'}
+
+path = sys.argv[1]
+assert '.conf' in path, "Error: .conf not found."
+with open(path, 'r+') as f:
+    param = f.read()
+    for old_name, new_name in name_dict.items():
+        # avoid substrings, won't replace first line
+        param = param.replace('\n' + old_name, '\n' + new_name)
```
iftenney: This is somewhat dangerous since it assumes well-behaved line breaks, and that the names aren't found in any strings values (like file paths).

It would be safer to use the utils in `src/config.py`, in particular `params_from_file()` and `write_params()`; you could do something like:
```
params = params_from_file(path)
for old_name, new_name in name_dict.items():
    if old_name in params:
        params[new_name] = params[old_name]
        del params[old_name]
write_params(params, path)
```

iftenney: (FYI, I just added support for `del params[key]` to the Params object)

iftenney:
LGTM, but will let @sleepinyourhat give final approval.

sleepinyourhat:
LGTM—merging.

FYI, we do see notifications for new commits, but it's better to add a comment as well so we know when the PR is totally ready to review.

PR #427, opened 2018-10-11
Load Transformer weights from a TensorFlow checkpoint

iftenney:
Turns out this is fairly straightforward, and we can use this to load custom models trained on BWB without converting to a brittle .npy format. 

Verified in the notebook that a checkpoint created from the original TF code loads to the same weight values as the huggingface PyTorch loader.


iftenney:
Is this hacky? It's pretty straightforward, I think - just a name mapping really. (You should see what the original loader looks like from the .npy files...)

Anyway, thanks!

PR #428, opened 2018-10-20
Update readme for recommendation on how to add tasks

Iwontbecreative:
AllenNLP's shuffler handling makes it so that ti will shuffle per data loaded into memory.
So e.g if you have a file where the first 1m lines are labelled 1 and the next 1m are labelled 0, it will first go through the 1m first examples before showing any example where the label is 0.
In practice, this makes models converge to predicting 1 and not generalizing.

Easiest fix is to recommend that people shuffle their training (and val I guess) data to add new tasks.


README.md
```diff
@@ -161,7 +161,7 @@ Note: if you want to train and evaluate on a task, that task must be in both ``t
 
 To add new tasks, you should:
 
-1. Add your data to the ``data_dir`` you intend to use. When constructing your task class (see next bullet), make sure you specify the correct subfolder containing your data. 
+1. Add your data to the ``data_dir`` you intend to use. When constructing your task class (see next bullet), make sure you specify the correct subfolder containing your data. It is recommended that you shuffle your training data if there are many rows (>10k) to avoid training artefacts.
```
sleepinyourhat: Add a bit more detail here: What happens if you don't?

sleepinyourhat:
Sorry to dwell on this, but be clear about what the main training code does. Does it never shuffle the data? Does it shuffle it only within buckets of a certain size? Does it shuffle it after the first epoch?

Iwontbecreative:
My understanding (might be flawed), is that `BucketIterator` it will first load `max_instances_in_memory` instances in memory (these are the `_memory_sized_lists`, for which there is no shuffle. In our case I believe this is set to be 10k.

Once that is done, it will bucket by length those 10k and shuffle these batches. So the shuffling only happens within loaded data AFAICT.

I ran into this issue while adding a large file (10m+ rows) sorted by label and seeing that all my batches only had positive instances.

sleepinyourhat:
I'd believe it. Mind updating the note when you have a chance?

Iwontbecreative:
Tried to make it clearer without being too verbose, thanks for the feedback

README.md
```diff
@@ -161,7 +161,7 @@ Note: if you want to train and evaluate on a task, that task must be in both ``t
 
 To add new tasks, you should:
 
-1. Add your data to the ``data_dir`` you intend to use. When constructing your task class (see next bullet), make sure you specify the correct subfolder containing your data. 
+1. Add your data to the ``data_dir`` you intend to use. When constructing your task class (see next bullet), make sure you specify the correct subfolder containing your data. It is recommended that you shuffle your training and validation data if there are many rows (>10k) to avoid training artefacts. Indeed, the framework loads 10k examples at a time in memory then shuffles those. E.g: in a binary classification setting, if your data is sorted by label, the model would go through all the positive examples before finding a negative, so would learn to always predict 1. If you reach 100% accuracy in training before one epoch, this is likely the case.
```
iftenney: Might be better to add this note (sentences 2:end) as a separate paragraph for emphasis?

iftenney: Also maybe add a TODO to the code to have a better solution to this, even something as simple as running `shuf` on the instance-record file emitted by the indexing step (this should work since each line is one b64-encoded example).

sleepinyourhat:
T: Mind finishing this soonish?

Iwontbecreative:
Updated for new README, let me know if further changes are required.

sleepinyourhat:
Thanks! @iftenney: Back to you.

Issue #376, opened 2018-09-02
Minor ELMo mixing weight cleanup


sleepinyourhat:
TODO here:

`    # TODO: rename variables to be more descriptive (mix_id -> task_id, mixer -> task_weights)
`
https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/utils.py#L119

Very minor, but could help with general sanity-checking.

sleepinyourhat:
@pitrack  - Fix or or close or unassign yourself?

PR #429, opened 2018-10-22
resolve variable naming TODO

pitrack:
fixing a TODO (#376)


pitrack:
Yes I did. Stopped it after it went through one or two train/eval loops (~10 minutes on the cpu machines) and both with/without look fine.

PR #434, opened 2018-11-16
Edge probing: UD data code & instructions

iftenney:
Cherry-picked from `edgeprobe_docs` to include the UD code only. Relative to that branch, also refactored argument handing & wrapper script for consistency with other datasets.


iftenney:
Once this is merged, everything for the paper should be on master. I'll continue to do some cleanup.

Issue #420, opened 2018-09-18
Clean up modules.py


iftenney:
I won't get to this until after the ICLR deadline, but I'd like to clean up the `src/` directory so dependencies are a bit cleaner. Also, `tasks.py` is massive and we should really split it up.

Proposed structure:
```
src/
    allennlp_mods/
    cove/ [submodule]
    __init__.py
    evaluate.py
    models.py
    modules/
        cnns/
        __init__.py
        <several .py files with old contents of modules.py>
        edge_probing.py
        seq2seq_decoder.py
        generate_s2s.py
    preprocess.py
    tasks/
        __init__.py
        tasks.py    # core definitions and registry
        edge_probing.py
        <other .py files containing various groups of tasks>
    trainer.py
    utils/    # or "misc/" or "lib/"
        __init__.py
        config.py
        emails.py
        gcp.py
        hocon_writer.py
        retokenize.py
        serialize.py
        utils.py
```

sleepinyourhat:
SGTM!

W4ngatang:
I'm all for it

iftenney:
Most of this is done now, but we might want to still split up `modules.py`. I shied away from doing this because of potential bugs this close to the NAACL deadline, but if someone (@W4ngatang maybe?) who knows this code better than I wants to take a stab at it, we could merge later in December.

sleepinyourhat:
What's left to do, @iftenney?

PR #438, opened 2018-11-16
Split up tasks.py

iftenney:
`tasks.py` is too big to fail!

Split out task registry into its own file, and separate edge probing tasks into their own file. Add package-level definitions in `src/tasks/__init__.py` for common imports & constants.

Note that `tasks.REGISTRY` is a global object and that tasks are added to it at import time by calling `register_task(...)`, so we need to make sure that any file with task definitions is imported before it can be used. We do this at the top of `src/tasks/__init__.py`, so that other code should need only do something like: `from .tasks import REGISTRY` to get access to everything.


iftenney:
Second part of addressing #420

sleepinyourhat:
Nudge.

iftenney:
Sorry! On vacation last week.

@W4ngatang : as crazy as you want to be. Since we can import in the `__init__.py` file it's easy to refactor this package without changing too much code in `src/`.

PR #450, opened 2019-01-05
Updated dependencies

pruksmhc:
mosestokenizer got separated out of nltk and updated, and last_dim_softmax is no longer in allennlp codebase

### DIFF
--- a/src/utils/utils.py
+++ b/src/utils/utils.py
@@ -11,7 +11,7 @@ import logging
 import codecs
 import time
 
-from nltk.tokenize.moses import MosesTokenizer, MosesDetokenizer
+from mosestokenizer import MosesTokenizer, MosesDetokenizer
 
 import numpy as np
 import torch
@@ -21,7 +21,6 @@ from torch.nn import Parameter
 from torch.nn import init
 
 from allennlp.common.checks import ConfigurationError
-from allennlp.nn.util import last_dim_softmax
 from allennlp.modules.seq2seq_encoders.seq2seq_encoder import Seq2SeqEncoder
 from allennlp.common.params import Params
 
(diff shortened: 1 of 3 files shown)


src/utils/utils.py
```diff
@@ -21,7 +21,6 @@
 from torch.nn import init
 
 from allennlp.common.checks import ConfigurationError
-from allennlp.nn.util import last_dim_softmax
```
iftenney: Is this used anywhere?

src/utils/utils.py
```diff
@@ -11,7 +11,7 @@
 import codecs
 import time
 
-from nltk.tokenize.moses import MosesTokenizer, MosesDetokenizer
+from mosestokenizer import MosesTokenizer, MosesDetokenizer
```
iftenney: If this is a separate package from the nltk download, can you add it to the Dockerfile as well?
https://github.com/jsalt18-sentence-repl/jiant/blob/master/Dockerfile

src/utils/utils.py
pruksmhc: It is but there's also a definition in the utils.py file (line 659)- actually, it seems like the definition in the utils.py file is last_dim_softmax in allennlp's version. Why was this definition added in?
We'll have to deprecate this line anyways with the new AllenNLP update.

src/utils/utils.py
```diff
@@ -163,7 +162,7 @@ def process_sentence(sent, max_seq_len, sos_tok=SOS_TOK, eos_tok=EOS_TOK):
     max_seq_len -= 2
     assert max_seq_len > 0, "Max sequence length should be at least 2!"
     if isinstance(sent, str):
-        return [sos_tok] + TOKENIZER.tokenize(sent)[:max_seq_len] + [eos_tok]
+        return [sos_tok] + TOKENIZER(sent)[:max_seq_len] + [eos_tok]
```
iftenney: Does the new package change the interface? If so, we need to change this everywhere else the tokenizer is called. Can you also verify that the new tokenizer exactly matches the old one?

src/utils/utils.py
W4ngatang: I think because I was ripping something out of AllenNLP early on before I wanted to invest in their eco-system...should be safe to pick one of using theirs or this clone (probably the former).

pruksmhc:
These changes are in the bigger one on Pytorch/Allen upgrades so closing this one for now

Issue #442, opened 2018-12-16
Update to PyTorch v1.0 and AllenNLP 0.8


W4ngatang:
New PyTorch version is out, evidently with big speed ups. Would be good to switch to (as well as possibly updating AllenNLP version).

iftenney:
This seems like a great project for the new SWE :)

sleepinyourhat:
Updated the title: AllenNLP now supports PyTorch 1.0, and we'll presumably have to upgrade both at once.

pruksmhc:
Have gone through the release docs and have a few thoughts on what I'm implementing now. 

@iftenney @sleepinyourhat 
-AllenNLP does have a TrainerBase class (below) but it's possible to not use it, and since the places where trainer.py  should be refactorized is not something that inheriting from TrainerBase class would help,, I think we should not worry about making our SamplingTaskTrainer inherit from this class. 
https://github.com/allenai/allennlp/blob/e08ade81ff6684869883ce3da66941b9d3a37503/allennlp/training/trainer_base.py 
-There's no need to implement from_params if you inherit from torch.Model if the from_param is a simple 1-1 mapping from JSON to the model.

pruksmhc:
So actually there might not be that many changes in terms of upgrading/using some new developments from more recent AllenNLP releases to our code.

sleepinyourhat:
Sounds good to me.

W4ngatang:
SGTM

Is it worth using their BERT implementation over ours? Or is it close enough that it's not worth the extra effort?

pruksmhc:
I think it's close enough that we can just stick to ours.

PR #451, opened 2019-01-07
Docker updates

iftenney:
Update Dockerfile and instructions, and add a `run_demo.sh` script that builds the container and runs `demo.conf`.

Also added a number of TODOs for where we could clean up paths and remove dependencies on our particular NFS directory structure. For now, `run_demo.sh` makes a temp dir to alias nfs inside the container, so it _should_ run anywhere. (TODO to test this on a blank machine...)


iftenney:
@pruksmhc this might be helpful for testing for #442. As currently configured, it uses the local copy of `jiant/` but uses the python packages from the docker image.

iftenney:
Force merge since this won't break anything except my experiments for now, and I've verified that those still work.

Issue #441, opened 2018-12-14
Incorporate PyTorch-BERT


sleepinyourhat:
https://twitter.com/Thom_Wolf/status/1073588652985466880

iftenney:
I'll look in to this (most likely after the holidays) - would like to run edge probing on BERT. @W4ngatang , any others want to help out?

iftenney:
Multi-GPU support would be very useful for BERT, since the models are huge.

sleepinyourhat:
@Iwontbecreative @zphang Did you wind up getting far with this?

sleepinyourhat:
There are now mentions of BERT in AllenNLP. Before we get too far, we should see what's already there.

Here's the issue, though it's not very helpful—I think something was committed without being mentioned here: https://github.com/allenai/allennlp/issues/1901

sleepinyourhat:
It'd also be good to support BERT-style _unsupervised_ training, rather than just fine-tuning from pre-trained BERT weights. I'm not sure if HuggingFace or AllenNLP support that as-is.

iftenney:
There's this: https://github.com/huggingface/pytorch-pretrained-BERT#2-bertforpretraining
But otherwise, they state that they don't intend to support this until PyTorch supports TPUs: https://github.com/huggingface/pytorch-pretrained-BERT#tpu

sleepinyourhat:
Safe to close this?

PR #452, opened 2019-01-07
BERT encoder module

iftenney:
BERT support (#441). Implementation modeled after that of the OpenAI transformer model, albeit somewhat cleaner using the excellent `pytorch_pretrained_bert` package (https://github.com/huggingface/pytorch-pretrained-BERT).

Currently works for edge probing experiments; see #460. For other tasks, you'll need to apply the BERT WPM tokenization either offline or at data loading time.


iftenney:
Note: this also depends on #451

config/defaults.conf
```diff
@@ -198,6 +198,26 @@ openai_transformer_fine_tune = 0  // If false, OpenAI Transformer weights will
                                   // Note that this might require a
                                   // high amount of GPU memory, so consider
                                   // lowering the batch size if set to true.
+bert_model_name = ""  // If nonempty, use this BERT model for representations.
+                      // Available values are:
+                      // bert-base-uncased, bert-large-cased, ...
+                      // For more, see:
+                      // https://github.com/huggingface/pytorch-pretrained-BERT#loading-google-ais-pre-trained-weigths-and-pytorch-dump
+                      // Not compatible with other embeddings due to different
```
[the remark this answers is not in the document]
iftenney: Yup.

pitrack:
I'm confused by tokenization. Earlier, each model corresponded to a tokenization scheme (e.g. openai-> openai-bpe). However, now it's broken down by task because bert cases for ner.

Then the order of priority is something like: default is MosesTokenizer(), then use the one specified by the model, then by the task? Eventually, we would want to update the readme or add comments somewhere which suggests this. It also feels a bit redundant (having both tasks have a class variable and another one in args).

src/bert/utils.py
```diff
@@ -0,0 +1,106 @@
+import os
+import random
+import copy
+import logging as log
+
+from typing import List, Dict
+
+import numpy as np
```
pitrack: [nit] some of these imports aren't used -- e.g. `os`, `copy`, `List`, `numpy`, or `random`. We probably have many more cases of this in the codebase. Perhaps the one to be most wary of is where we call/import `random`

iftenney: Yeah, some of this was copypasta from the openai model wrapper, which requires a lot more preprocessing. Fixed.

src/bert/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Returns:
+            h: [batch_size, seq_len, d_emb]
+        """
+        assert "bert_wpm_pretokenized" in sent
+        # <int32> [batch_size, var_seq_len]
+        var_ids = sent["bert_wpm_pretokenized"]
+
+        # Model has fixed, learned positional component, so we must pass a
+        # block of exactly seq_len length.
```
W4ngatang: Since (I think) the model will always load positional embeddings for 512(?) positions, does `seq_len` need to be set to 512 when using a BERT model?

src/bert/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+        mask = (ids == 0)
+        # "Correct" ids to account for different indexing between BERT and
+        # AllenNLP.
+        # The AllenNLP indexer adds a '@@UNKNOWN@@' token to the
+        # beginning of the vocabulary, *and* treats that as index 1 (index 0 is
+        # reserved for padding).
+        FILL_ID = 0  # [PAD] for BERT models.
+        ids[ids == 0] = FILL_ID + 2
+        ids -= 2
```
W4ngatang: this will map elements of `ids` that are 1 to -1, potentially causing indexing errors; is there a guard against this or a reason it won't happen?

src/bert/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+        if self.embeddings_mode != "none":
+            # Extract lexical embeddings; see
+            # https://github.com/huggingface/pytorch-pretrained-BERT/blob/master/pytorch_pretrained_bert/modeling.py#L186
+            h_lex = self.model.embeddings.word_embeddings(ids)
+            h_lex = self.model.embeddings.LayerNorm(h_lex)
+            # following our use of the OpenAI model, don't use dropout for
+            # probing.
+            #  h_lex = self.model.embeddings.dropout(embeddings)
```
W4ngatang: Maybe make it an option whether or not the model is being used in a probing context? I think we'll (read: I will want to) use this in a non-probing context.
Also, I forget is the convention is to apply dropout at the end of a module or at the beginning of the next use (I remember this caused an issue during the workshop).

src/bert/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+        if self.embeddings_mode != "only":
+            # encoded_layers is a list of layer activations, each of which is
+            # <float32> [batch_size, seq_len, output_dim]
+            encoded_layers, _ = self.model(ids, token_type_ids=torch.zeros_like(ids),
```
W4ngatang: does this do double computation of embedding layers if you're in `cat` mode? If so, maybe reuse `h_lex` if it's there? (though I guess lookups are fast).

src/models.py
```diff
@@ -69,6 +69,21 @@ def build_model(args, vocab, pretrained_embs, tasks):
...
+        # should prevent these from being enabled anyway.
+        from .bert.utils import BertEmbedderModule
+        log.info(f"Using BERT model ({args.bert_model_name}); skipping other embedders.")
+        cove_layer = None
+        # Set PYTORCH_PRETRAINED_BERT_CACHE environment variable to an existing
+        # cache; see
+        # https://github.com/huggingface/pytorch-pretrained-BERT/blob/master/pytorch_pretrained_bert/file_utils.py
+        bert_cache_dir = (os.path.join(args.exp_dir, "bert_cache") if
+                          args.bert_fine_tune else None)
+        if bert_cache_dir:
```
W4ngatang: dumb question: why only cache if fine tuning? Is the cache dir for storing fine tuned weights?

src/preprocess.py
```diff
@@ -259,6 +262,13 @@ def build_tasks(args):
                              ("OpenAI transformer is not supported alongside"
                               " other indexers due to tokenization!")
         indexers["openai_bpe_pretokenized"] = SingleIdTokenIndexer("openai_bpe")
+    if args.bert_model_name:
+        assert not indexers, ("BERT is not supported alongside"
+                              " other indexers due to tokenization!")
+        assert args.tokenizer == args.bert_model_name, \
+                             ("BERT is not supported alongside"
+                              " other indexers due to tokenization!")
```
W4ngatang: should be "tokenizers"?

src/bert/utils.py
iftenney: I don't think so, < 512 should also work since BERT has an internal attention mask (https://github.com/huggingface/pytorch-pretrained-BERT/blob/master/pytorch_pretrained_bert/modeling.py#L620)

src/bert/utils.py
iftenney: Good catch. We pass the tokens through the BERT WordpieceTokenizer, which checks for OOV and replaces with it's own unk token https://github.com/huggingface/pytorch-pretrained-BERT/blob/master/pytorch_pretrained_bert/tokenization.py#L329 (which has some id > 1 in the main vocab), so this should never happen.

Added an assert to complain in case something goes wrong.

src/bert/utils.py
iftenney: FWIW, this is basically a shortcut to treating the BERT WPM like GloVe vectors. Erring on the side of no dropout here, since there are a bunch of places we can add it later if these are the only embeds?

I'd rather not add an additional "is_probing" param since that introduces a lot of cruft.

src/bert/utils.py
iftenney: Yes it's a little wasteful but I'd have to break open the BertModel implementation to re-use h_lex. As implemented here we don't have to modify that code at all.

src/models.py
iftenney: Good point; I think this is unnecessary. Safest thing is to download once then mark as read-only, I guess?

There was a line in the documentation advising to use separate directories for each worker in distributed training, but this might just be to avoid hammering a single file server.

src/preprocess.py
iftenney: Fixed.

src/models.py
iftenney: Removed this special case; should always use cache if the environment var is set.

src/bert/utils.py
W4ngatang: :+1:

src/bert/utils.py
W4ngatang: sounds good, nit: should add a more descriptive error message here

src/bert/utils.py
W4ngatang: :+1:

sleepinyourhat:
Ready to merge?

sleepinyourhat:
Nudge again. Other stuff is waiting on this.

PR #454, opened 2019-01-08
Task definitions cleanup

iftenney:
Delete over 200 lines of boilerplate classes that were used to parameterize parent classes; these are now handled with `@register_task` directly.

Also add `**kw` to task constructors and forward to parent class, so we can add features to the Task class without further modifying the subclasses.

### DIFF
--- a/src/tasks/tasks.py
+++ b/src/tasks/tasks.py
@@ -182,16 +182,10 @@ class ClassificationTask(Task):
     ''' General classification task '''
     pass
 
-    #  def __init__(self, name):
-    #      super().__init__(name)
-
 
 class RegressionTask(Task):
     ''' General regression task '''
     pass
-
-    #  def __init__(self, name):
-    #      super().__init__(name)
 
 
 class SingleClassificationTask(ClassificationTask):
@@ -315,9 +309,6 @@ class RankingTask(Task):
     ''' Generic sentence ranking task, given some input '''
     pass
 
-    #  def __init__(self, name):
-    #      super().__init__(name)
-
 
 class LanguageModelingTask(SequenceGenerationTask):
     """Generic language modeling task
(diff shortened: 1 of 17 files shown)


iftenney:
Yes, but I thought it would be easier to split this out since there are so many lines changed. This PR should be a no-op, while #456 affects loading functionality for some tasks (edge probing for now).

src/tasks/tasks.py
```diff
@@ -412,14 +406,11 @@ def get_sentences(self) -> Iterable[Sequence[str]]:
                 yield sent
 
 
+@register_task('bwb', rel_path='BWB/')
```
W4ngatang: TODO: this is weird but out of scope of this PR

src/tasks/tasks.py
```diff
@@ -740,24 +721,18 @@ def get_metrics(self, reset=False):
                 'precision': pcs, 'recall': rcl}
 
 
-@register_task('qqp-alt', rel_path='QQP/')
-class QQPAltTask(QQPTask):
```
W4ngatang: From the other PR thread: (Alex) I think the original motivation for structuring the -alt tasks is that you could have attention for the -alt, but not the original task. Does this setup still allow us to do that? I think so because task-specific module construction depends on task name and not task class, but seems worth testing to be sure.

src/tasks/tasks.py
```diff
@@ -1811,31 +1671,14 @@ def _make_instance(input1, input2, labels):
...
-class DisSentWikiBigFullTask(DisSentTask):
-    ''' Task class for DisSent with Wikitext 103 considering clauses from within a single sentence
-        or across two sentences.
-        Data sets should be prepared as described in Nie, Bennett, and Goodman (2017) '''
-
-    def __init__(self, path, max_seq_len, name="dissentwikifullbig"):
-        super().__init__(path, max_seq_len, "wikitext.dissent.big", name)
-
-
+# TODO: does this even work? What is n_classes for this?
```
W4ngatang: From other PR thread:
I think n_classes = 8 for all DisSent tasks.
I think we didn't use DisSentWikiSingleTask.
Maybe @tommccoy1 can comment on if that's right, and if it's okay to delete the within-sentence task?

src/tasks/tasks.py
```diff
@@ -1457,12 +1339,15 @@ def load_data(self, path, max_seq_len):
         log.info("\tFinished loading JOCI data.")
 
 
+@register_task('wmt_debug', rel_path='wmt_debug/', max_targ_v_size=5000)
```
W4ngatang: From other PR thread: TODO: add TODO to clean up dummy / debug tasks

src/tasks/tasks.py
W4ngatang: oh sorry, I just realized this comment is not in reference to the DisSent stuff; ignore.

src/tasks/tasks.py
W4ngatang: Confirmed that if you use settings like `sts-b-pair_attn = 0, sts-b-alt-pair_attn = 1`, you will create a model with two different components, one of which has attention and one of which will not. Looks good!

PR #455, opened 2019-01-08
Split out some tasks into separate .py files

iftenney:
Figured I'd do this alongside #454. Not sure if this is the ideal split, but it's a start and should make it a bit easier to find things.

See #420 

**Note:** I've tested `demo.conf`, but not all the individual tasks in these breakout files. I manually checked most symbols, but worst case is there might be some import errors when running these other tasks.


src/tasks/lm.py
```diff
@@ -0,0 +1,154 @@
...
+        for split in self.files_by_split:
+            # Don't use test set for vocab building.
+            if split.startswith("test"):
+                continue
+            path = self.files_by_split[split]
+            for sent in self.load_data(path):
+                yield sent
+
+
+@register_task('bwb', rel_path='BWB/')
```
W4ngatang: see comment in other PR about funkiness here

This PR seems like the place to address the issue. I or Yada can take a look.

src/tasks/reddit.py
```diff
@@ -0,0 +1,202 @@
...
+from .tasks import RankingTask, PairClassificationTask
+from .tasks import sentence_to_text_field
+from .registry import register_task
+
+@register_task('reddit', rel_path='Reddit_2008/')
+@register_task('reddit_dummy', rel_path='Reddit_2008_TestSample/')
+@register_task('reddit_3.4G', rel_path='Reddit_3.4G/')
+@register_task('reddit_13G', rel_path='Reddit_13G/')
+@register_task('reddit_softmax', rel_path='Reddit_2008/')
+class RedditTask(RankingTask):
```
W4ngatang: are these all really ranking tasks? These could use cleanup + more informative names, e.g. what is `reddit` vs `reddit_softmax`.

src/tasks/reddit.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def get_metrics(self, reset=False):
+        '''Get metrics specific to the task'''
+        acc = self.scorer1.get_metric(reset)
+        return {'accuracy': acc}
+
+
+@register_task('mt_pair_classif', rel_path='wmt14_en_de_local/')
+@register_task('mt_pair_classif_dummy', rel_path='wmt14_en_de_mini/')
+class MTDataPairClassificationTask(RedditPairClassificationTask):
```
W4ngatang: little weird to have this in a file named `reddit`

src/tasks/reddit.py
iftenney: @pappagari might know?

src/models.py
```diff
@@ -30,10 +30,12 @@
...
 from .tasks.tasks import CCGTaggingTask, ClassificationTask, CoLATask, GroundedSWTask, \
-    GroundedTask, LanguageModelingTask, MTTask, MultiNLIDiagnosticTask, PairClassificationTask, \
-    PairOrdinalRegressionTask, PairRegressionTask, RankingTask, RedditSeq2SeqTask, \
+    GroundedTask, MultiNLIDiagnosticTask, PairClassificationTask, \
+    PairOrdinalRegressionTask, PairRegressionTask, RankingTask, \
     RegressionTask, SequenceGenerationTask, SingleClassificationTask, SSTTask, STSBTask, \
-    TaggingTask, WeakGroundedTask, Wiki103Seq2SeqTask, JOCITask
+    TaggingTask, WeakGroundedTask, JOCITask
+from .tasks.lm import LanguageModelingTask
+from .tasks.mt import MTTask, RedditSeq2SeqTask, Wiki103Seq2SeqTask
```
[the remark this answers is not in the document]
iftenney: Not sure I follow - most of these are still in tasks.py, the ones I moved are imported from tasks.lm, tasks.mt, etc.

src/tasks/reddit.py
iftenney: Agreed, but oddly this depends on `RedditPairClassificationTask` - we should probably fix this by cleaning up the task hierarchy.

sleepinyourhat:
I'd vote to kill tasks.py, and split it into something like `base`, `glue`, and `misc`.

src/tasks/reddit.py
pappagari: Yeah all of them are ranking tasks. All of those tasks optimize binary cross entropy except reddit_softmax. In reddit_softmax, softmax is taken over all the samples in the batch (which includes one positive sample and batch_size-1 negative samples). And likelihood for the positive sample is maximized.

src/tasks/reddit.py
W4ngatang: :+1: let's deal with this in another PR

src/models.py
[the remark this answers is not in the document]
W4ngatang: oh nvm, I was confused. We only need to import the tasks for which there are task-specific module considerations.

src/tasks/reddit.py
W4ngatang: Gotcha. Might rename some of them in cleaning up tasks but seems okay to leave it for this PR

PR #456, opened 2019-01-08
Tokenizer name as config param

iftenney:
Passed to `Task` constructor. Currently used by edge probing to select retokenized data, but can also be used (not yet implemented) to run tokenizers when the data is loaded.


src/preprocess.py
```diff
@@ -228,35 +228,37 @@ def build_tasks(args):
...
 
     # 1) create / load tasks
-    tasks, train_task_names, eval_task_names = \
-        get_tasks(parse_task_list_arg(args.pretrain_tasks),
-                  parse_task_list_arg(args.target_tasks), args.max_seq_len,
-                  path=args.data_dir, scratch_path=args.exp_dir,
-                  load_pkl=bool(not args.reload_tasks),
-                  nli_prob_probe_path=args['nli-prob'].probe_path,
-                  max_targ_v_size=args.max_targ_word_v_size)
+    tasks, train_task_names, eval_task_names = get_tasks(args)
```
W4ngatang: looks like this already includes the `get_tasks` refactor?

src/preprocess.py
```diff
@@ -391,52 +393,62 @@ def parse_task_list_arg(task_list):
             task_names.append(task_name)
     return task_names
 
-
-def get_tasks(train_task_names, eval_task_names, max_seq_len, path=None,
-              scratch_path=None, load_pkl=1, nli_prob_probe_path=None,
-              max_targ_v_size=20000):
+def _get_task(name, args, data_path, scratch_path):
```
W4ngatang: I'm assuming this is the same as in the other PR

src/tasks/tasks.py
```diff
@@ -138,15 +139,13 @@ def count_examples(self, splits=['train', 'val', 'test']):
             count = self.get_num_examples(st)
             self.example_counts[split] = count
 
+    def tokenizer_is_supported(self, tokenizer_name):
+        ''' Check if the tokenizer is supported for this task. '''
+        return tokenizer_name == utils.TOKENIZER.__class__.__name__
```
W4ngatang: isn't this always `MosesTokenizer`? I'm not seeing where in utils this might change if you specify a different tokenizer.

src/tasks/tasks.py
```diff
@@ -412,14 +406,11 @@ def get_sentences(self) -> Iterable[Sequence[str]]:
                 yield sent
 
 
+@register_task('bwb', rel_path='BWB/')
 class WikiTextLMTask(LanguageModelingTask):
```
W4ngatang: there's something funky going on here...

We might as well merge rename this class (I think the only difference with the base LM class is that we assume a particular format of the raw data) and merge it with the Wiki103 task, maybe even with the base LM class?

W4ngatang: though maybe this is an issue for a different PR

src/tasks/tasks.py
```diff
@@ -707,11 +687,12 @@ def get_metrics(self, reset=False):
 
 
 @register_task('qqp', rel_path='QQP/')
+@register_task('qqp-alt', rel_path='QQP/')  # second copy for different params
```
W4ngatang: I think the original motivation for structuring the `-alt` tasks is that you could have attention for the `-alt`, but not the original task. Does this setup still allow us to do that?

W4ngatang: I think yes because modules are based on names and not the actual task class, but we should test to make sure...

src/tasks/tasks.py
```diff
@@ -1457,12 +1339,15 @@ def load_data(self, path, max_seq_len):
         log.info("\tFinished loading JOCI data.")
 
 
+@register_task('wmt_debug', rel_path='wmt_debug/', max_targ_v_size=5000)
```
W4ngatang: do we want to keep this debug example?

src/tasks/tasks.py
```diff
@@ -1811,31 +1671,14 @@ def _make_instance(input1, input2, labels):
...
-class DisSentWikiBigFullTask(DisSentTask):
-    ''' Task class for DisSent with Wikitext 103 considering clauses from within a single sentence
-        or across two sentences.
-        Data sets should be prepared as described in Nie, Bennett, and Goodman (2017) '''
-
-    def __init__(self, path, max_seq_len, name="dissentwikifullbig"):
-        super().__init__(path, max_seq_len, "wikitext.dissent.big", name)
-
-
+# TODO: does this even work? What is n_classes for this?
```
W4ngatang: I think `n_classes = 8` for all DisSent tasks.
I think we didn't use `DisSentWikiSingleTask`.
Maybe @tommccoy1 can comment on if that's right, and if it's okay to delete the within-sentence task?

src/tasks/tasks.py
iftenney: Yes; I'm just referencing that so it doesn't need to be defined in as many places.

src/tasks/tasks.py
iftenney: Yeah, I didn't want to do anything that I couldn't easily tell was a no-op.

src/tasks/tasks.py
iftenney: I didn't find any references for the alt class names, only for the actual task names. Can you verify this?

src/tasks/tasks.py
iftenney: Let's leave as a TODO for now so that this PR stays a no-op.

src/tasks/tasks.py
iftenney: Let's leave as a TODO for now so that this PR stays a no-op.

src/preprocess.py
iftenney: Whoops, will merge that in and update this.

iftenney:
@W4ngatang : the changes to `tasks.py` are part of #454 - can you comment there?

src/tasks/tasks.py
W4ngatang: Gotcha. Just to check my understanding, a task with different tokenization should overwrite this to something like `tokenizer_name in [tokenizer1, tokenizer2, ...]`?

And once BERT is supported, this should look something like `tokenizer_name in [MosesTokenizer, BERTTokenizer]`?

config/defaults.conf
```diff
@@ -160,7 +160,12 @@ max_targ_word_v_size = 20000  // Maximum target word vocab size for seq2seq task
 
 
 // Input Handling //
-
+tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task constructor for appropriate
+                              // handling during data loading. It's expected that all tasks use the same 
+                              // tokenizer; doing otherwise will cause an error.
+                              // 'MosesTokenizer' is the default. Use "" for edge probing tasks to use the
+                              // native tokenization, or one of 'OpenAI.BPE',
```
W4ngatang: can we clarify what "native tokenization" here is?

src/tasks/tasks.py
W4ngatang: cool, thanks!

config/defaults.conf
W4ngatang: gotcha

PR #458, opened 2019-01-09
Remap edge probing configs to use JIANT_PROJECT_PREFIX

iftenney:
This makes it easier to debug locally, since it'll use an experiment directory
in the user's home directory instead of on NFS. Kubernetes runs are unaffected
since JIANT_PROJECT_PREFIX and NFS_PROJECT_PREFIX are the same in that
environment.


iftenney:
Force-merging this because it only affects edge probing.

PR #461, opened 2019-01-20
BERT -> CCG 

pruksmhc:
Got OpenAI -> CCG working, which had a little trickiness with bpe tokenization and the target tag labels. 
Following BERT paper and how it handles NER, we only take into account in the loss and score the probability of the first sub-piece after BPE tokenization. For eaxmple, 
e.g., if the original dataset had:
([Mr., Smith, is, tall], [N/N, N, S/N, S/N/(S/N)]) (edited) 
(just making up those tags)
Then the BPE tokenizer might turn it into:
([Mr, ., Smith, is, tall], [N/N, #, N, S/N/(S/N)]), where # will not be scored.


sleepinyourhat:
I assume this is ready for @iftenney to double check? (Thanks!)

src/utils/utils.py
```diff
@@ -295,6 +301,64 @@ def tags_to_ixs(tags, tag_to_ix_dict, ix_to_tag_dic):
             'ix_to_knowledge_dic': ix_to_knowledge_dic
             }
 
+def moses_tokenizer(text):
+    from sacremoses import MosesTokenizer
+    mt = MosesTokenizer()
+    return mt.tokenize(text, return_str=True)
+
+def adjust_targs_BPE(targs, bpe_input, orig_input, orig_tokenizer_func):
```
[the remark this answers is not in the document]
pruksmhc: alright looking into it - thanks for the heads up!

pruksmhc:
Actually there seems to be a circular dependency from the latest push to master - right now, if we want to add OpenAI BPE to utils.py, we need to instantiate openai_utils. I think we should either a) do tokenizing elsewhere outside of utils.py (probably bad idea) b) not use any functions in utils.py in any modules that do tokenization (bad idea), or c) move _all_tokenizer instantiation to utils.py (probably the best out of these 3 ideas), so we instantiate openai_utils tokenize and encode function in utils.py. thoughts @iftenney ?

iftenney:
I think a) is the right way to go, we can make a separate file `src/utils/tokenizers.py` that can depend on `src/utils/utils.py`. But I think we want to keep the OpenAI and BERT tokenizers alongside their implementations (i.e. in src/bert/... and src/openai_transformer/...), so either `tokenizers.py` can import these or we can set up some registry (similar to the tasks registry) so they're loaded on import of the OpenAI and BERT packages.

pruksmhc:
Actually, I think tokenization + loading data should be in src/preprocess.py since that is part of preprocessing.

sleepinyourhat:
No rush, but where does this stand?

src/utils/utils.py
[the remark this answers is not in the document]
pruksmhc: The Levenshtein distance doesn't work as well unfortunately

sleepinyourhat:
(@pruksmhc)

pruksmhc:
This is modified to become BERT -> CCG. It's currently transferred over to Haokun + Jason), cleaning it up right now to make it ready for merge

config/none_ft/ccg.conf
```diff
@@ -0,0 +1,31 @@
+include "../defaults.conf"
```
sleepinyourhat: What does none_ft mean?

config/none_ft/ccg.conf
```diff
@@ -0,0 +1,31 @@
...
+
+pretrain_tasks = ccg
+target_tasks= ccg
```
sleepinyourhat: Is there a reason this needs to be set in both? Can we just set it as a target task?

(Also, please be consistent with spacing.)

config/none_ft/ccg.conf
```diff
@@ -0,0 +1,31 @@
...
+project_dir = bert_stilts
+exp_dir = bert_ccg
+run_dir = main
+max_seq_len = 261 // This is the maximum length of all the various datasets in CCG.
```
sleepinyourhat: Two spaces before //

config/none_ft/ccg.conf
```diff
@@ -0,0 +1,31 @@
...
+load_model = 0
```
sleepinyourhat: Why override this here?

config/none_ft/ccg.conf
```diff
@@ -0,0 +1,31 @@
...
+local_log_path=logs
```
sleepinyourhat: Why override this here?

sleepinyourhat: Ditto for random_seed, reload*.

config/none_ft/ccg.conf
sleepinyourhat: You also set it twice in this file.

src/utils/utils.py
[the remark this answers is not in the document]
iftenney: Err, weird. Can you open an issue documenting what wasn't working here?

scripts/preprocessing/CCG/align_tags_to_bert.py
```diff
@@ -0,0 +1,100 @@
...
+import os
+sys.path.insert(1, os.path.join(sys.path[0], '../../'))
+
+"""
+Usage:
+    python ccg_allignment.py --data_dir {path/to/data/file} -t {tokenizer_name}
+This preprocessing file will preprocess the CCG data using the tokenizer,
+saving it alongside the original files.
+
+This file introduces a new tag to sub-words (if the tokenizer splits a word
```
iftenney: Not sure what this comment means. Can you give an example of the output?

scripts/preprocessing/CCG/align_tags_to_bert.py
```diff
@@ -0,0 +1,100 @@
...
+import sys
+import argparse
+import pandas as pd
+import pickle
+import os
+sys.path.insert(1, os.path.join(sys.path[0], '../../'))
+
+"""
+Usage:
+    python ccg_allignment.py --data_dir {path/to/data/file} -t {tokenizer_name}
```
iftenney: Document what the input and output formats for this are?

scripts/preprocessing/CCG/align_tags_to_bert.py
```diff
@@ -0,0 +1,100 @@
+
+from src.utils import retokenize
+from src import utils
+import sys
+import argparse
+import pandas as pd
+import pickle
+import os
+sys.path.insert(1, os.path.join(sys.path[0], '../../'))
```
iftenney: This is really brittle - use relative imports instead, like:

```python
from ..foo_module import foo
```

scripts/preprocessing/CCG/align_tags_to_bert.py
```diff
@@ -0,0 +1,100 @@
...
+        split: str,
+        tokenizer_name: str,
+        data_dir: str,
+
+        Returns
+        --------------
+        None, saves tag alligned files to same directory as the original file.
+    """
+    tags_to_id = pickle.load(open("tags_to_id", "rb"))
+    ccg_text = pd.read_csv(
```
iftenney: nit: hard to read, put this on one or two lines?

iftenney:
What is `scripts/preprocessing/CCG/tags_to_id`?

src/models.py
```diff
@@ -965,10 +965,21 @@ def _seq_gen_forward(self, batch, task, predict):
...
+                predict: (boolean) predict mode (not supported)
+        Returns
+            out: (dict)
+                - 'logits': output layer, dimension: [batchSize * task.max_seq_len, task.num_tags]
+                - 'loss': size average CE loss
+        '''
         out = {}
-        b_size, seq_len, _ = batch['inputs']['elmo'].size()
+        # batch[inputs] only has one item
+        b_size, seq_len = next(iter(batch["inputs"].values())).size()
```
iftenney: Why the `next(iter(` construct?

src/models.py
```diff
@@ -981,7 +992,15 @@ def _tagger_forward(self, batch, task, predict):
             logits = logits.view(b_size * seq_len, -1)
             out['logits'] = logits
             targs = batch['targs']['words'][:, :seq_len].contiguous().view(-1)
-
+        if "mask" in batch:
+            # prevent backprop for tags generated for tokenization-introduced tokens
```
iftenney: Why not just mask out the loss for these positions, so the gradients are zero?

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,64 @@ def get_metrics(self, reset=False):
...
+                dataset_mask = []
+                for i in range(len(dataset[2])):
+                    mask = ma.getmask(ma.masked_where(np.array(dataset[2][i]) != self.introduced_token, np.array(dataset[2][i])))
+                    mask_indices = np.where(mask == True)[0].tolist()
+                    dataset_mask.append(mask_indices)
+                masks.append(dataset_mask)
+
+        # mock labels for test data (tagging)
+        te_targs = [['0'] * len(x) for x in te_data[0]]
+        te_mask = [range(len(x)) for x in te_data[0]]
```
iftenney: use `list(range(...))` ? (range gives an iterator in Python3)

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,64 @@ def get_metrics(self, reset=False):
...
+                           s1_idx=1, s2_idx=None, label_idx=2, skip_rows = 1, col_indices=[0, 1, 2], delimiter="\t", has_labels=False)
+        self.max_seq_len = max_seq_len
+
+        # Get the mask for each sentence, where the mask is whether or not 
+        # the token was split off by tokenization. We want to only count the first
+        # sub-piece in the BERT tokenization in the loss and score, following Devlins NER
+        # experiment [BERT: Pretraining of Deep Bidirectional Transformers for Language Understanding]
+        # (https://arxiv.org/abs/1810.04805)
+        if self.bert_model:
+            from ast import literal_eval
```
iftenney: Delete this?

src/models.py
pruksmhc: Yep that's also definitely another way to do this!

pruksmhc:
tags_to_id.json is a dictionary mapping between the tag names and the current IDs

config/ccg.conf
```diff
@@ -0,0 +1,17 @@
+include "defaults.conf"
```
sleepinyourhat: Maybe title this something like ccg-bert.conf, or else get rid of BERT here.

scripts/ccg/align_tags_to_bert.py
```diff
@@ -0,0 +1,99 @@
...
+tags.
+This preprocessing file will preprocess the CCG data using the tokenizer,
+saving it alongside the original files.
+
+This file introduces a new tag to sub-words (if the tokenizer splits a word.
+Currently, this supports BERT tokenization.)
+For example, 
+[Mr., Porter] -> [Mr, .,, Por, ter]. Thus, if Mr. was given a tag of 5 and Porter 6
+in the original CCG, then the alligned tags will be [5, 1362, 6, 1362], where 1362 indicates
+a subpiece word that has been split due to tokenization.
```
sleepinyourhat: ...of a word

scripts/ccg/align_tags_to_bert.py
```diff
@@ -0,0 +1,99 @@
...
+    res_tags = []
+    introduced_tokenizer_tag = len(tag_dict) - 1
+    for i in range(len(text)):
+        token = text[i]
+        _, new_toks = aligner_fn(token)
+        res_tags.append(tag_dict[current_tags[i]])
+        if len(new_toks) > 1:
+            for tok in new_toks[1:]:
+                res_tags.append(introduced_tokenizer_tag)
+                # based on BERT-paper for wordpiece,, we only predict tag
```
sleepinyourhat: typo

scripts/ccg/align_tags_to_bert.py
```diff
@@ -0,0 +1,99 @@
...
+    str_tags = [str(s) for s in res_tags]
+    return " ".join(str_tags)
+
+
+def align_tags_BERT(split, tokenizer_name, data_dir):
+    """
+        We align BERT tags such that any introduced tokens introudced
+        by BERT will be assigned a special tag, which later on in
+        preprocessing/task building will be converted into a tag mask (so that
+        we do not take into account the model predictin for tokens introduced by
```
sleepinyourhat: typo

scripts/ccg/align_tags_to_bert.py
```diff
@@ -0,0 +1,99 @@
...
+        the tokenizer).
+        Args
```
sleepinyourhat: Line break here

scripts/ccg/align_tags_to_bert.py
```diff
@@ -0,0 +1,99 @@
+import sys
```
sleepinyourhat: Haven't verified the logic in detail. If you haven't already, run a few sanity checks to make sure we're not getting off-by-ones in position or index anywhere.

src/models.py
```diff
@@ -965,10 +965,21 @@ def _seq_gen_forward(self, batch, task, predict):
...
+                predict: (boolean) predict mode (not supported)
+        Returns
+            out: (dict)
+                - 'logits': output layer, dimension: [batchSize * task.max_seq_len, task.num_tags]
+                - 'loss': size average CE loss
+        '''
         out = {}
-        b_size, seq_len, _ = batch['inputs']['elmo'].size()
+        # batch[inputs] only has one item
+        b_size, seq_len = list(batch["inputs"].values()[0]).size()
```
sleepinyourhat: Assuming this works both with and without ELMo?

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
...
 class CCGTaggingTask(TaggingTask):
     ''' CCG supertagging as a task.
         Using the supertags from CCGbank. '''
-
-    def __init__(self, path, max_seq_len, name, **kw):
+    def __init__(self, path, max_seq_len, name="ccg", **kw):
         ''' There are 1363 supertags in CCGBank. '''
-        super().__init__(name, num_tags=1363, **kw)
+        super().__init__(name, 1363, **kw)
+        self.introduced_token = '1362'
```
sleepinyourhat: Capitalize this, since it's a constant.

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
...
+        self.bert_model = 1 if kw['tokenizer_name'].startswith("bert") else 0
```
sleepinyourhat: You can just delete '1 if' and 'else 0', right?

sleepinyourhat: Also, something like bert_tokenization would be a more informative name.

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
...
         self.load_data(path, max_seq_len)
         self.sentences = self.train_data_text[0] + self.val_data_text[0]
+        self.max_seq_len = max_seq_len
+        if self._tokenizer_name.startswith("bert-"):
+            # the +1 is for the tokenizatin added token
```
sleepinyourhat: Typo

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
...
+                          s1_idx=1, s2_idx=None, label_idx=2, skip_rows = 1, col_indices=[0, 1, 2],  delimiter="\t", label_fn=lambda t: t.split(' '))
+        val_data = load_tsv(self._tokenizer_name, os.path.join(path, "ccg.dev."+self._tokenizer_name), max_seq_len,
+                            s1_idx=1, s2_idx=None, label_idx=2, skip_rows = 1, col_indices=[0, 1, 2], delimiter="\t", label_fn=lambda t: t.split(' '))
+        te_data = load_tsv(self._tokenizer_name, os.path.join(path, 'ccg.test.'+self._tokenizer_name), max_seq_len,
+                           s1_idx=1, s2_idx=None, label_idx=2, skip_rows = 1, col_indices=[0, 1, 2], delimiter="\t", has_labels=False)
+        self.max_seq_len = max_seq_len
+
+        # Get the mask for each sentence, where the mask is whether or not 
+        # the token was split off by tokenization. We want to only count the first
+        # sub-piece in the BERT tokenization in the loss and score, following Devlins NER
```
sleepinyourhat: apostrophe

src/models.py
pruksmhc: Yep!

iftenney: Parens are off? (`list.size()` doesn't make sense, does it?)

src/models.py
iftenney: Ok. If going with a gather/select approach, can you add a comment to that extent, something like: "select only the targets that correspond to original tags, ignoring the introduced placeholder tags"

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
...
     ''' CCG supertagging as a task.
         Using the supertags from CCGbank. '''
-
-    def __init__(self, path, max_seq_len, name, **kw):
+    def __init__(self, path, max_seq_len, name="ccg", **kw):
         ''' There are 1363 supertags in CCGBank. '''
-        super().__init__(name, num_tags=1363, **kw)
+        super().__init__(name, 1363, **kw)
+        self.INTRODUCED_TOKEN = '1362'
+        self.bert_tokenization = 1 if kw['tokenizer_name'].startswith("bert") else 0
```
iftenney: Use `self._tokenizer_name.startswith("bert-")` instead of looking in `kw`?

Also don't need the 1 if ... else 0, should be fine to just store the bool.

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
...
 
-
 @register_task('ccg', rel_path='CCG/')
 class CCGTaggingTask(TaggingTask):
     ''' CCG supertagging as a task.
         Using the supertags from CCGbank. '''
-
-    def __init__(self, path, max_seq_len, name, **kw):
+    def __init__(self, path, max_seq_len, name="ccg", **kw):
         ''' There are 1363 supertags in CCGBank. '''
```
iftenney: Are there 1363 supertags with or without INTRODUCED_TOKEN?

pruksmhc: without introduced token

scripts/ccg/align_tags_to_bert.py
pruksmhc: Test written

scripts/ccg/align_tags_to_bert.py
```diff
@@ -0,0 +1,102 @@
...
+import argparse
+import pandas as pd
+import os
+import json
+from ..src import utils
+from src.utils import retokenize
```
iftenney: This is a little confusing - either use relative imports or only rely on the symlink (probably safer)?
```
from src import utils
from src.utils import retokenize
```

iftenney: This is a little confusing - either use relative imports or only rely on the symlink (probably safer)?
```
from src import utils
from src.utils import retokenize
```

src/tasks/tasks.py
```diff
@@ -1558,43 +1558,63 @@ def get_metrics(self, reset=False):
         acc = self.scorer1.get_metric(reset)
         return {'accuracy': acc}
 
-    def process_split(self, split, indexers) -> Iterable[Type[Instance]]:
```
iftenney: Are there any other tasks that subclass TaggingTask and might rely on this?

pruksmhc: Nope!

PR #466, opened 2019-01-25
Feed variable-length sequences to BertModel

iftenney:
Original implementation used fixed-length padding by analogy with OpenAI model, but the huggingface BertModel implementation supports an attention mask so we can pass variable-length sequences directly. This speeds up runs by **4-6x**.

Verified that a standalone `src.bert.utils.BertEmbedderModule` gives the same behavior with or without padding, and on small edge probing runs there is no difference in F1 or MCC outside of normal random variation.


iftenney:
Verified with `demo.conf`, and with edge probing BERT experiments.

PR #467, opened 2019-01-25
Minor cleanup of edge probing runner script

iftenney:
Specify GPU type alone with the cluster so we don't need to run twice with different arguments.


iftenney:
Force-merging because this only affects edge probing job configs.

PR #469, opened 2019-01-26
[DNM] Bert friends exps

W4ngatang:
Set up experiments for BERT versions of ELMo's Friends experiments. The relevant changes are in `config/final-bert.conf` and `scripts/final_glue_runs.sh`. The rest of the PR is added by #468, so we should wait for that PR before merging (but the jobs can be launched from this branch).

Basically:
- only using ELMo character-level word encoder => only using BERT lexical embeddings
- using full ELMo model => using top level BERT embeddings (or should this be concatenation of lexical embeddings and top level outputs?)

Other details:
- I'm planning on using BERT base cased as things will train faster and fit in memory easier.
- I'm *not* planning on fine-tuning BERT during pretraining. We could fine-tune during target task training, which would be more in-line with BERT's original usage, but that would be less in-line with the rest of the paper.
- I haven't yet actually tried launching all of these, but I anticipate some GPU memory issues and needing to change some training settings to adjust.


sleepinyourhat:
What's the status of this PR? We should either merge it into master soon (before the hackathon?) or just leave it as a permanent branch.

sleepinyourhat:
If this doesn't get merged in, we should at least turn this branch into a release branch and add some documentation for when the paper comes out.

PR #476, opened 2019-01-30
Killing many zombies on utils.py

pruksmhc:
While I was refactoring utils.py, I came across legacy code that were copy-pasted from AllenNLP that was not being used elsewhere.


src/utils/utils.py
```diff
@@ -368,552 +368,6 @@ def split_data(data, ratio, shuffle=1):
...
-
-
-def subsequent_mask(size):
-    "Mask out subsequent positions."
-    attn_shape = (1, size, size)
-    subsequent_mask = np.triu(np.ones(attn_shape), k=1).astype('uint8')
-    return torch.from_numpy(subsequent_mask) == 0
-
-
 @Seq2SeqEncoder.register("masked_multi_head_self_attention")
```
iftenney: TODO: We should probably move this (and anything like it) to something under `src/modules`

pruksmhc:
Yep, verified with ./demo.sh

W4ngatang:
FWIW I think this code was used a long time ago when I was using their BiDAF model but for whatever reason didn't want to completely invest in their infrastructure, so I pulled out these helper functions with it. If they're not being used anywhere, they're safe to delete (though probably good to search the repo in GitHub or some smart editor to be sure they're not tucked away in some random module file).

src/utils/utils.py
```diff
@@ -20,7 +20,7 @@
 from torch.nn import Parameter
 from torch.nn import init
 
-from allennlp.nn.util import masked_log_softmax
+from allennlp.nn.util import masked_softmax
```
W4ngatang: why do we not need to take log softmax anymore? Taking log softmax is usually more numerically stable

src/utils/utils.py
```diff
@@ -368,552 +368,6 @@ def split_data(data, ratio, shuffle=1):
...
-    if tensor.dim() != mask.dim():
-        raise ConfigurationError(
-            "tensor.dim() (%d) != mask.dim() (%d)" %
-            (tensor.dim(), mask.dim()))
-    one_minus_mask = 1.0 - mask
-    values_to_add = replace_with * one_minus_mask
-    return tensor * mask + values_to_add
-
-
-def device_mapping(cuda_device: int):
```
iftenney: This is called still in `utils.py` and is causing my experiments to crash. Can we please run tests so we're not breaking things 3 weeks before a paper deadline?

https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/utils/utils.py#L69

PR #477, opened 2019-01-30
Rewrote load_tsv with pandas 

pruksmhc:
Wanted to get some input on this.  I saw that load_tsv is currently a very, very long function that also takes a long time, so re-implemented using widely used pandas. Also changed some parameters of the functions to get it to be a little bit more intuitive (for example, having return_indices as a function because there's some tasks where you return the indices for shuffling purposes). 
Would like input on how the functions should be used (also the role of load_diagnostic_tsv, since right now it's only used for one task). 
This has been tested on demo.conf with all the flags set to 1 (for loading, indexing a task etc), and also built a Dockerfile to make sure nose2 doesn't break any dependencies.


sleepinyourhat:
Ian and Alex will know this code better than I do. This sounds worthwhile, though!

pruksmhc:
Alright, will add unit tests to make sure the behavior isn't changed. With load_diagnostic_tsv, its quite a strange function because it indexes the non-label columns. I think it's fine to leave it as it is right now since only 1 task is using it, but if we need to do this for more tasks then we might have to do a bigger refactor.

sleepinyourhat:
Ready for review?

pruksmhc:
@sleepinyourhat  yep!

src/utils/utils.py
```diff
@@ -187,108 +190,75 @@ def load_lines(filename: str) -> Iterable[str]:
...
+            word_to_ix = {word: i+1 for i, word in enumerate(vocab) if word !=''}
+            ix_to_word = {i+1:word for i, word in enumerate(vocab) if word !=''}
+            rows[index] = rows[index].apply(lambda x: [word_to_ix[x]] if x != '' else [])
+            ix_to_word[0] = "missing"
+            word_to_ix["missing"] = 0
+            return word_to_ix, ix_to_word, rows[index]
+        sent1s = rows[s1_idx].apply(lambda x: process_sentence(x, max_seq_len))
+        sent2s = rows[s2_idx].apply(lambda x: process_sentence(x, max_seq_len))
+        labels = rows[label_idx].apply(lambda x: label_fn(x))
+        # build indices for field attributes
```
sleepinyourhat: Add a line above. (And capitalize.)

src/utils/utils.py
```diff
@@ -187,108 +190,75 @@ def load_lines(filename: str) -> Iterable[str]:
...
+                            header=None, \
+                            skiprows=skip_rows, \
+                            quoting=csv.QUOTE_NONE,\
+                            encoding='utf-8')
+        rows = rows.fillna('')
+        def targs_to_idx(index):
+            # this function builds the index to vocab (and its inverse) mapping
+            # and then indexes the column based on the index passed in
+            vocab = set(rows[index].values)
+            word_to_ix = {word: i+1 for i, word in enumerate(vocab) if word !=''}
```
sleepinyourhat: Double check pep8—should be `i + 1`.

src/utils/utils.py
```diff
@@ -187,108 +190,75 @@ def load_lines(filename: str) -> Iterable[str]:
...
+                            error_bad_lines=False, \
+                            header=None, \
+                            skiprows=skip_rows, \
+                            quoting=csv.QUOTE_NONE,\
+                            encoding='utf-8')
+        rows = rows.fillna('')
+        def targs_to_idx(index):
+            # this function builds the index to vocab (and its inverse) mapping
+            # and then indexes the column based on the index passed in
+            vocab = set(rows[index].values)
```
sleepinyourhat: Consider renaming. This implies (wrongly, usually) that these values will be ordinary words.

sleepinyourhat: Unless I'm missing something...

src/utils/utils.py
```diff
@@ -187,108 +190,75 @@ def load_lines(filename: str) -> Iterable[str]:
...
+            word_to_ix = {word: i+1 for i, word in enumerate(vocab) if word !=''}
+            ix_to_word = {i+1:word for i, word in enumerate(vocab) if word !=''}
+            rows[index] = rows[index].apply(lambda x: [word_to_ix[x]] if x != '' else [])
+            ix_to_word[0] = "missing"
```
sleepinyourhat: Say what this means in this context? Do you expect to see it?

src/utils/utils.py
```diff
@@ -299,63 +269,66 @@ def tags_to_ixs(tags, tag_to_ix_dict, ix_to_tag_dic):
...
+    Returns:
+        List of first and second sentences, labels, and if applicable indices
+    '''
+    sent1s, sent2s, labels = pd.Series(), pd.Series(), pd.Series()
+    rows = pd.read_csv(data_file, \
+                        sep=delimiter, \
+                        error_bad_lines=False, \
+                        header=None, \
+                        skiprows=skip_rows, \
+                        quoting=csv.QUOTE_NONE,\
```
sleepinyourhat: Add a comment about this choice, and do some quick testing to make sure this isn't a problem on any of the existing data files—it's the kind of thing that can easily mess us up.

src/utils/utils.py
sleepinyourhat: Same for comments elsewhere.

src/utils/utils.py
```diff
@@ -187,108 +190,75 @@ def load_lines(filename: str) -> Iterable[str]:
...
+    try:
+        rows = pd.read_csv(data_file, \
+                            sep=delimiter, \
+                            error_bad_lines=False, \
+                            header=None, \
+                            skiprows=skip_rows, \
+                            quoting=csv.QUOTE_NONE,\
+                            encoding='utf-8')
+        rows = rows.fillna('')
+        def targs_to_idx(index):
```
iftenney: It seems weird to re-implement this functionality - can we just use an AllenNLP `Vocabulary` object here as we do elsewhere in the code?

src/utils/utils.py
```diff
@@ -187,108 +190,75 @@ def load_lines(filename: str) -> Iterable[str]:
...
+        data_file: string
+        max_seq_len: int
+        s1_idx: int
+        s2_idx: int
+        label_fn: function
+        skip_rows: list of ints
+        delimiter: string
+
+    Returns:
+        A dictionary of the necessary indexed fields, the tokenized sent1 and sent2
```
iftenney: It seems like a leaky abstraction to do the indexing in this function - can we do the indexing layer as with other tasks, and use the standard indexing/vocab code?

iftenney: I guess this functionality was here before, so fine to leave a TODO to clean it up.

src/utils/utils.py
pruksmhc: not exactly sure what that does, but I was simply mimicking the functionality of the load_diagnostic_tsv  @W4ngatang  @Jan21 might know more

src/utils/utils.py
pruksmhc: Yep, I'll leave as TODO

src/utils/utils.py
```diff
@@ -190,105 +194,74 @@ def load_lines(filename: str) -> Iterable[str]:
...
+        sent2s = rows[s2_col].apply(lambda x: process_sentence(x, max_seq_len))
+        labels = rows[label_col].apply(lambda x: label_fn(x))
+        # Build indices for field attributes
+        lex_sem_to_ix_dic, ix_to_lex_sem_dic, lex_sem = targs_to_idx("Lexical Semantics")
+        pr_ar_str_to_ix_di, ix_to_pr_ar_str_dic, pr_ar_str = targs_to_idx("Predicate-Argument Structure")
+        logic_to_ix_dic, ix_to_logic_dic, logic = targs_to_idx("Logic")
+        knowledge_to_ix_dic, ix_to_knowledge_dic, knowledge = targs_to_idx("Knowledge")
+        idxs = rows.index
+
+    except Exception as e:
```
iftenney: See https://docs.python.org/3.1/howto/doanddont.html#except - this is pretty much guaranteed to mask subtle (or not so subtle) bugs. 

What kinds of issues is this expected to trigger on?

src/utils/utils.py
W4ngatang: I don't really know why this is a separate function...I think Jan Hula authored this?

src/utils/utils.py
W4ngatang: I don't think I authored this, but I think instead of a try/except, we can add an assertion that the row is the desired length; I think that's what the except was trying to catch originally? The data should be well-formed, in any case, so I'm not totally sure why this was needed or if it was being triggered.

W4ngatang: Agree that we shouldn't catch general exception though

src/utils/utils.py
```diff
@@ -299,63 +272,66 @@ def tags_to_ixs(tags, tag_to_ix_dict, ix_to_tag_dic):
...
+    '''
+    Load a tsv.
+    To load only rows that have a certain value for a certain column,
+    like genre in MNLI, set filter_idx and filter_value.
+    Args:
+        s1_idx; int
+        s2_idx: int
+        targ_idx: int
+        return_indices: bool that describes if you need to return indices (for purposes of matching)
+        label_fn is a function that expects a row and outputs the label
```
W4ngatang: can you add documentation for what `filter_idx` is? Not obvious to me

src/tasks/tasks.py
```diff
@@ -792,11 +802,11 @@ def __init__(self, path, max_seq_len, name, probe_path="probe_dummy.tsv",
     def load_data(self, path, max_seq_len, probe_path):
         targ_map = {'neutral': 0, 'entailment': 1, 'contradiction': 2}
         tr_data = load_tsv(os.path.join(path, 'train_dummy.tsv'), max_seq_len,
-                           s1_idx=1, s2_idx=2, targ_idx=None, targ_map=targ_map, skip_rows=0)
+                           s1_idx=1, s2_idx=2, has_labels=False, targ_map=targ_map, skip_rows=0)
```
W4ngatang: why does the training data not have labels?

pruksmhc:
Actually there's a slightly bigger problem - this pandas rewriting works for all datasets except SNLI. The SNLI/train.tsv file downloaded from download_glue_data.py is very noisy - more specifically, many of the lines have multiple labels, while most have only 2 labels. For example, 

11  -{more columns before}	entailment	neutral	entailment	neutral	neutral	neutral
vs 
12. - {more columns before}	neutral	neutral

The 11th row in train.tsv has 15 columns, with an extra 4 that are some sort of label, while most other rows only have 11 columns. 
The load_tsv script handled this by looping over each line and getting the last index (-1) as the gold label. However, in panas, we preload the line before getting the labels, and there's no easy way to dynamically fill a pandas dataframe. Right now, the load_tsv will define the number of columns as 11 and then simply skip over the noisy rows that have the extra labels. 
Thoughts? @W4ngatang  @iftenney 
We might either not do this pandas rewriting, clean the train.tsv, or simply for snli use the old way of doing load_tsv only and use the pandas way for all others.

src/tasks/tasks.py
pruksmhc: The previous implementation for this task didn't return any labels as well- let me check for NPS why there are no labels for train though.

sleepinyourhat:
Hrm. Could this be a problem with the GLUE data preparation script? In the
original SNLI training set, there's never any ambiguity over which column
the gold label appears in. Some other columns may be blank, but that
column's always at the same index...

iftenney:
> However, in panas, we preload the line before getting the labels, and there's no easy way to dynamically fill a pandas dataframe.

Pandas does have very good support for missing values, so it might be possible to load the _maximum_ number of columns and then propagating the last value for each row. See https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html - I haven't tried this, but `axis=1` and `method="ffill"` (forward fill) might do the right thing.

pruksmhc:
So seems like there is a slightly hacky way to do it: https://stackoverflow.com/questions/15242746/handling-variable-number-of-columns-with-pandas-python .
After talking with Sam, we'll just for now temporarily not support SNLI (since it's barely used, and open an issue to internally conver the jsonl to the format we can use). We'll just stick to this current version for now.

src/tasks/tasks.py
pruksmhc: Actually,I'm honestly not sure what NPS is - I can't find it in download_data_glue

src/utils/data_loaders.py
```diff
@@ -13,157 +17,135 @@ def load_tsv(
...
+    Returns:
+        A dictionary of the necessary indexed fields, the tokenized sent1 and sent2
+        and indices
+        Note: If a field in a particular row in the dataset is empty, we return []
+        for that field for that row, otherwise we return an array of ints (indices)
+        Else, we return an array of indices
+    '''
+    # TODO: Abstract indexing layer from this function so that MNLI-diagnostic calls load_tsv
+    assert len(s1_col) > 0 and len(label_col) > 0, "Make sure you passed in column names for sentence 1 and labels"
+    sent1s, sent2s, targs, idxs, lex_sem, pr_ar_str, logic, knowledge = pd.Series(), pd.Series(), pd.Series(), pd.Series(), pd.Series(), pd.Series(), pd.Series(), pd.Series()
```
iftenney: See note above - don't need placeholders for most of these

iftenney:
> So seems like there is a slightly hacky way to do it: https://stackoverflow.com/questions/15242746/handling-variable-number-of-columns-with-pandas-python .

This actually doesn't seem hacky at all. Does it work?

src/utils/data_loaders.py
```diff
@@ -13,157 +17,135 @@ def load_tsv(
...
+        sent2s = rows[s2_idx].apply(lambda x: process_sentence(tokenizer_name, x, max_seq_len))
+
+    if has_labels:
+        if label_fn is None:
+            labels = rows[label_idx]
+        else:
+            labels = rows[label_idx].apply(lambda x: label_fn(x))
+    else:
+        # If dataset doesn't have labels, for example for test set, then mock labels
+        labels = np.zeros(len(rows), dtype=int)
```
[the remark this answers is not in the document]
pruksmhc: That will set the label_idx column to 0 but we want to actually return labels as a list for consistency

pruksmhc:
So loading in variable number of columns actually isn't we want, the bad thing about that we still have to transform our labels (labels = rows[label_idx].apply(lambda x: label_fn(x))), and if we use the loading of variable number of columns, we'll have to somehow keep track of the last place that isn't "NaN" and it'll just be a lot more complicated for a dataset which is messy. 
Alternatively, I'm making a script right now that will parse SNLI and cleans it so that it will be able to be loaded

src/tasks/tasks.py
```diff
@@ -550,7 +552,7 @@ def get_metrics(self, reset=False):
 @register_task('snli', rel_path='SNLI/')
 class SNLITask(PairClassificationTask):
     ''' Task class for Stanford Natural Language Inference '''
-
+    # TODO: CLean up SNLI dataset.
```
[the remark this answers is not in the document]
pruksmhc: Made it compatible

src/utils/data_loaders.py
```diff
@@ -13,157 +17,135 @@ def load_tsv(
...
+        rows = rows[rows[filter_idx] == filter_value]
+    sent2s = pd.Series()
+    # Filter for sentence1s that are of length 0
+    # Filter if row[targ_idx] is nan
+    mask = (rows[s1_idx].str.len() > 0)
+    if has_labels:
+        mask = mask & rows[label_idx].notnull()
+    rows = rows.loc[mask]
+    sent1s = rows[s1_idx].apply(lambda x: process_sentence(tokenizer_name, x, max_seq_len))
+    if s2_idx is not None:
```
iftenney: nit: have both cases here rather than relying on the default init earlier:
```
if s2_idx is None:
    sent2s = pd.Series()  # or should this be None?
else:
    sent2s = rows[s2_idx].apply(...)
```

src/utils/data_loaders.py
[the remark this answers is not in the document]
iftenney: Sorry, I meant before this. But just a style thing - this is fine.

iftenney:
LGTM

sleepinyourhat:
Merge when ready! (Assuming you have merge access—LMK if not.)

PR #480, opened 2019-01-31
Add ELMo-style scalar mixing to BERT wrapper

iftenney:
Per suggestion of BERT authors, this makes everything dramatically better when using BERT-as-a-feature. Results coming soon.


src/bert/utils.py
```diff
@@ -24,6 +26,16 @@ def __init__(self, args, cache_dir=None):
         for param in self.model.parameters():
             param.requires_grad = bool(args.bert_fine_tune)
 
+        # Configure scalar mixing, ELMo-style
```
sleepinyourhat: Make this a warning?

config/defaults.conf
```diff
@@ -216,6 +216,8 @@ bert_embeddings_mode = "none"  // How to handle the embedding layer of the
                                // "cat" returns top-layer concatenated with
                                //   lexical layer
                                // "only" returns only lexical layer
+                               // "mix" uses ELMo-style scalar mixing (with
+                               // learned weights) across all layers
```
sleepinyourhat: Make it clear what one should do to run BERT-style fine-tuning?

src/bert/utils.py
```diff
@@ -24,6 +26,16 @@ def __init__(self, args, cache_dir=None):
...
+        # TODO: if doing multiple target tasks, allow for multiple sets of
+        # scalars. Seee the ELMo implementation here:
+        # https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L115
+        if self.embeddings_mode == "mix":
+            num_layers = self.model.config.num_hidden_layers
+            self.scalar_mix = scalar_mix.ScalarMix(num_layers + 1,
+                                                   do_layer_norm=False)
```
sleepinyourhat: Are gradients to these handled in the same way as ELMo scalar gradients? (i.e., trained even when the rest of BERT isn't?)

config/defaults.conf
iftenney: That's handled by the `bert_fine_tune` parameter below.

src/bert/utils.py
iftenney: Made an assert.

src/bert/utils.py
iftenney: Yes, that's the intent. Added comment to clarify.

config/defaults.conf
sleepinyourhat: Er, sorry: What should one set here to follow the standard BERT protocol (when bert_fine_tune = 1). None? Does it not matter?

src/bert/utils.py
sleepinyourhat: Cool. I'll assume you have a way to double check this.

config/defaults.conf
sleepinyourhat: Clarify and merge?

sleepinyourhat:
Still not my intended question. Rephrasing: If you want to use jiant to
exactly reproduce one of the experiments in the BERT paper, how should you
set bert_embeddings_mode?

iftenney:
@sleepinyourhat : Sorry, didn't understand what you were asking before. `bert_embeddings_mode=none` should return just the top layer, which is the standard behavior.

Updated comments and added "top" as a synonym.

sleepinyourhat:
Tx!

PR #471, opened 2019-01-28
Pytorch and AllenNLP upgrade

pruksmhc:
Upgrading to 1.0.0 and 0.8.1
Refer to #442  for notes. 
The last_dim_softmax in AllenNLP was merged into masked_softmax (See this PR: https://github.com/allenai/allennlp/pull/1687/files).

PR #481, opened 2019-01-31
Update Dockerfile for compatibility with new AllenNLP and PyTorch 1.0

iftenney:
Fix breakages in Dockerfile due to #471, mostly related to different package requirements in the new version of AllenNLP. Verified that `demo.conf` (no overrides) and `edgeprobe_bert.conf` with `target_tasks=edges-spr2,bert_embeddings_mode=cat,bert_model_name=bert-base-uncased` produce identical results (up to random noise) before and after this update. 

_(Note: before tests run using a snapshot before #471 for compatibility with older AllenNLP and PyTorch)_.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,19 +4,11 @@
 # install instructions (gcp/kubernetes/README.md) to install Docker and
 # nvidia-docker.
 #
-# Usage:
-#   docker build -t jiant-sandbox:v1 .
-#   export JIANT_PATH="/nfs/jsalt/path/to/jiant"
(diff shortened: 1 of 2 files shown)


iftenney:
Can you run it locally? That should be enough for testing.

sleepinyourhat:
Hypothetically yes, but not in the near future. If you think it's crucial
to have an outside test, Yada may be able to?

iftenney:
Ok. I'll merge for now, since it's definitely not more broken than at head :)

PR #483, opened 2019-02-05
Edge probing: notebook to find diff examples, and better pretty-printing

iftenney:
Now also prints predictions that are above the threshold, and decodes label indices if a vocab is given.

Included notebook can be used to find sentences with significant (or minimal) differences between two runs.


iftenney:
Force-merge: edge probing only

PR #485, opened 2019-02-09
Add scalar mixing to OpenAI Transformer LM

iftenney:
Follow-up for #480, add this to the GPT model so we can do an apples-to-apples comparison.

### DIFF
--- a/src/openai_transformer_lm/utils.py
+++ b/src/openai_transformer_lm/utils.py
@@ -143,7 +143,7 @@ class TransformerModel(nn.Module):
         if self.embeddings_mode in ["none", "top"]:
             h = encoded_layers[-1]
         #  elif self.embeddings_mode == "only":
-        #      h = h_lex
+        #      handled above by early return
         elif self.embeddings_mode == 'cat':
             # Concatenate embeddings layer.
             h = torch.cat([encoded_layers[-1], h_lex], dim=2)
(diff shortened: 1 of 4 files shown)


src/openai_transformer_lm/utils.py
```diff
@@ -179,6 +200,24 @@ def __init__(self, args, n_special=3, n_ctx=512):
...
+            # TODO: if doing multiple target tasks, allow for multiple sets of
+            # scalars. See the ELMo implementation here:
+            # https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L115
+            assert len(parse_task_list_arg(args.target_tasks)) <= 1, \
+                    ("openai_embeddings_mode='mix' only supports a single set of "
+                     "scalars (but if you need this feature, see the TODO in "
+                     "the code!)")
+            if not args.openai_transformer_fine_tune:
+                log.warning("NOTE: openai_embeddings_mode='mix', so scalar "
+                            "mixing weights will be fine-tuned even if BERT "
```
W4ngatang: BERT => openai

iftenney: Whoops, fixed :)

iftenney:
Tested and this works, merging now. As expected, mixing performs a bit better than concatenation.

PR #487, opened 2019-02-14
Inference interface

zphang:
Added a REPL/corpus-level inference interface.


src/interface/inference.py
```diff
@@ -0,0 +1,180 @@
+'''Run a REPL or corpus-level inference
```
sleepinyourhat: Add a bit more in the way of usage instructions.

src/interface/inference.py
```diff
@@ -0,0 +1,180 @@
...
+
+def handle_arguments(cl_arguments):
+    parser = argparse.ArgumentParser(description='')
+    # Configuration files
+    parser.add_argument('--config_file', '-c', type=str, nargs="+",
+                        help="Config file(s) (.conf) for model parameters.")
+    parser.add_argument('--overrides', '-o', type=str, default=None,
+                        help="Parameter overrides, as valid HOCON string.")
+    parser.add_argument('--inference_mode', type=str, default="repl",
+                        help="Running as REPL or process corpus."
```
sleepinyourhat: Odd phrasing.

src/interface/inference.py
sleepinyourhat: Make it clear this is CoLA only, and add some notes on what it'd take to make this flexible, or else just do the work to make it flexible.

src/interface/inference.py
```diff
@@ -0,0 +1,180 @@
...
+            torch.cuda.set_device(args.cuda)
+        except Exception:
+            log.warning(
+                "GPU access failed. You might be using a CPU-only"
+                " installation of PyTorch. Falling back to CPU."
+            )
+            args.cuda = -1
+
+    # Prepare data #
+    pretrain_tasks, target_tasks, vocab, word_embs = build_tasks(args)
```
sleepinyourhat: Skip as much as possible of the preprocessing for any non-CoLA tasks, and consider moving the CoLA assert earlier.

src/interface/inference.py
sleepinyourhat: Also spell out what a REPL is.

src/interface/inference.py
```diff
@@ -0,0 +1,180 @@
...
+from allennlp.nn.util import move_to_device
+
+log.basicConfig(format='%(asctime)s: %(message)s',
+                datefmt='%m/%d %I:%M:%S %p', level=log.INFO)
+
+
+def handle_arguments(cl_arguments):
+    parser = argparse.ArgumentParser(description='')
+    # Configuration files
+    parser.add_argument('--config_file', '-c', type=str, nargs="+",
```
iftenney: Is there an option to load an existing run, instead of relying on the configs to match?

src/interface/inference.py
iftenney: nit: add example usage in top-level docstring

src/interface/inference.py
```diff
@@ -0,0 +1,180 @@
...
+
+def main(cl_arguments):
+    ''' Run REPL for a CoLA model '''
+    cl_args = handle_arguments(cl_arguments)
+    args = config.params_from_file(cl_args.config_file, cl_args.overrides)
+
+    # Raise error if obsolete arg names are present
+    check_arg_name(args)
+
+    if args.cuda >= 0:
```
iftenney: A lot of the set-up code is shared with `main.py` - can we unify the code somehow to avoid duplication?

src/interface/inference.py
```diff
@@ -0,0 +1,180 @@
...
+
+def repl_loop(model, vocab, indexers, task, args):
+    print("Input CTRL-C or enter 'QUIT' to terminate.")
+    while True:
+        try:
+            input_string = input(" INPUT: ")
+            if input_string == "QUIT":
+                break
+
+            tokens = process_sentence(
```
iftenney: Not all tasks use this input processing. If we're sticking to that abstraction (processing done on the Task object), maybe refactor so this calls a `task.process_input(...)` method?

src/interface/inference.py
sleepinyourhat: What would that involve?

src/interface/inference.py
sleepinyourhat: Agreed, though I'd encourage you (@zphang) to err on the side of getting a working version checked in.

src/interface/inference.py
zphang: We can also use the `params.conf` generated for each run.

src/interface/inference.py
zphang: Should I leave a raise NotImplementedError on the remaining tasks? I feel like it might take some work before I can implement a more general abstraction applicable across tasks.

zphang:
Updated with changes:

- Docstring documentation
- Earlier check for CoLA
- Misc other cleaning up of naming

sleepinyourhat:
Cool. @iftenney: Back to you.

sleepinyourhat:
@iftenney - Nudge. Thanks!

src/interface/inference.py
iftenney: yeah, that's fine.

iftenney:
Sorry for the delay, hands full with an ACL submission.

Can you verify that `demo.conf` works, and this should be ok? Also, since only CoLA is supported for now, can we name the file as such? It also seems a little odd to have this in a subdirectory under `src/`, since that's usually for libraries - maybe have it be just `cola_repl.py` in the jiant root directory (next to `main.py`)?

zphang:
Verified that demo.conf leads to same results with/without changes.

Regarding putting the file in root / naming: I agree that moving the file to root makes sense, although I also don't want to end up putting too many scripts in root. That, combined with potentially extending the script to work with subsequent tasks in the future, makes me think that perhaps we should call it `inference.py`.

The alternative is to call is `cola_inference.py` for now, and change it if/when it gets extended later.

sleepinyourhat:
Let's go with Ian's suggestion and leave it in root. Calling it cola_inference for now and renaming it later sounds good.

If that's done and the demo sanity-check works out, let me know, and I can merge.

zphang:
Moved/renamed gto cola_inference.py.

`demo.conf` sanity check passed, so should be good to merge.

PR #488, opened 2019-02-15
Refactor trainer.py 

pruksmhc:
Current changes: 
-Modularizing trainer.py 
-Removed unused "for_train" parameter 
-Put build_trainer_params inside build_trainer

Tested on config/demo.conf on pretraining, target task training and full eval


src/trainer.py
```diff
@@ -28,8 +28,7 @@
 
 
 def build_trainer_params(args, task_names):
-    ''' In an act of not great code design, we wrote this helper function which
-    extracts trainer parameters from args. In particular, we want to search args
+    ''' helper function which extracts trainer parameters from args. In particular, we want to search args
```
sleepinyourhat: Be consistent about capitalization.

src/trainer.py
```diff
@@ -96,22 +102,20 @@ def build_trainer(params, model, run_dir, metric_should_decrease=True):
...
+                           'lr_decay': .99, 'min_lr': params['min_lr'],
+                           'keep_all_checkpoints': params['keep_all_checkpoints'],
+                           'val_data_limit': params['val_data_limit'],
+                           'max_epochs': params['max_epochs'],
+                           'dec_val_scale': params['dec_val_scale'],
+                           'training_data_fraction': params['training_data_fraction']})
+    if train_type == "SamplingMultiTaskTrainer":
+        trainer = SamplingMultiTaskTrainer.from_params(model, run_dir,
+                                                       copy.deepcopy(train_params))
     return trainer, train_params, opt_params, schd_params
```
sleepinyourhat: What value does trainer have if train_type isn't SamplingMultiTaskTrainer? If it's supposed to fail, make this an assert earlier on.

src/trainer.py
```diff
@@ -671,8 +692,130 @@ def _aggregate_results(self, tasks, task_infos, metric_infos):
...
+        all_val_metrics['macro_avg'] /= len(tasks)
+        return n_examples_overall, task_infos, all_val_metrics
+
     def _validate(self, epoch, tasks, batch_size, periodic_save=True):
-        ''' Validate on all tasks and return the results and whether to save this epoch or not '''
+        '''
+        Validate on all tasks and return the results and whether to save this epoch or not
+        Parameters
+        ----------
+        epoch: int,
```
sleepinyourhat: ?

src/trainer.py
```diff
@@ -671,8 +692,130 @@ def _aggregate_results(self, tasks, task_infos, metric_infos):
...
+            if task_name == 'macro':
+                new_best_macro = True
+        if out_of_patience:
+            metric_infos[metric]['stopped'] = True
+            log.info("Out of patience. Stopped tracking %s", task_name)
+        return metric_infos, this_epoch_metric, should_save, new_best_macro
+
+    def _get_validation_performance(self, task, task_infos, tasks, batch_size, all_val_metrics, n_examples_overall):
+        """
+        This function evaluates each task, builds validation generator, and gets the validation metrics.
```
sleepinyourhat: This doesn't quite make sense. Don't you have to build the generators in order to evaluate each task?

src/trainer.py
```diff
@@ -671,8 +692,130 @@ def _aggregate_results(self, tasks, task_infos, metric_infos):
...
+        Parameters
+        ----------
+        task: Current task to get validation performance of.
+        task_infos: Instance of information about the task.
```
sleepinyourhat: Say where to find the format for this.

src/trainer.py
```diff
@@ -671,8 +692,130 @@ def _aggregate_results(self, tasks, task_infos, metric_infos):
...
+        tasks: A list of task objects to train on.
+        batch_size: The batch size to use for the tasks.
+        all_val_metrics: A dcitionary storing the validation performance.
+        n_Examples_overall = int, current number of examples the model is validated on.
```
sleepinyourhat: Not sure what this means.

src/trainer.py
```diff
@@ -671,8 +692,130 @@ def _aggregate_results(self, tasks, task_infos, metric_infos):
...
+        Returns
+        -------
+        n_examples_overall - int, current number of examples.
```
sleepinyourhat: Be consistent about : vs - vs = here.

src/trainer.py
```diff
@@ -671,8 +692,130 @@ def _aggregate_results(self, tasks, task_infos, metric_infos):
...
+        '''
+        Validate on all tasks and return the results and whether to save this epoch or not
+        Parameters
+        ----------
+        epoch: int,
+        tasks: A list of task objects to train on.
+        batch_size: int, the batch size to use for the tasks.periodic_save
+        periodic_save: Boolean value of whether or not to save model and progress periodically
+
+        Returns
```
sleepinyourhat: Nix?

src/trainer.py
pruksmhc: Yes you do - this function does build generators and then evaluates on the task. Could you clarify what is confusing for you?

sleepinyourhat: This phrasing makes it sound like it evaluates, then builds generators, then builds metrics.

src/trainer.py
```diff
@@ -96,22 +102,21 @@ def build_trainer(params, model, run_dir, metric_should_decrease=True):
...
         log.info('\tUsing ReduceLROnPlateau scheduler!')
 
-    train_params = Params(
-        {
-            'cuda_device': params['cuda'],
-            'patience': params['patience'],
-            'grad_norm': params['max_grad_norm'],
-            'val_interval': params['val_interval'],
-            'max_vals': params['max_vals'],
-            'lr_decay': .99,
```
sleepinyourhat: This *really* shouldn't be set here. Worth looking into. That said, we can merge this before dealing with it.

PR #489, opened 2019-02-18
Refactoring tokenizers + adding OpenAI tokenizer

pruksmhc:
Tested on demo.conf (MosesTokenizer) and OpenAI tokenizer, made sure that the functions in retokenize_edge_data that uses Moses detokenization from util.py still works as expected.

The thinking behind the change was to localize tokenizations in one place - there are still the implementations of BERT and Openai transformer that also have the respective tokenizations but this diff will be a move towards localization + a standard way of adding tokenizers. 
TODO: May make an actual registry + add class decorators to tokenizers.


probing/retokenize_edge_data.py
```diff
@@ -75,7 +75,7 @@ def _get_bert_tokenizer(model_name, do_lower_case):
 # of a TokenAligner instance and a list of tokens.
 def align_moses(text: Text) -> Tuple[retokenize.TokenAligner, List[Text]]:
     moses_tokens = MosesTokenizer.tokenize(text)
-    cleaned_moses_tokens = utils.unescape_moses(moses_tokens)
+    cleaned_moses_tokens = MosesTokenizer.detokenize(moses_tokens)
```
iftenney: It's not clear if this behaves the same as the old version. Can we keep the old `utils.unescape_moses` here until after the ACL deadline at least?

src/preprocess.py
```diff
@@ -248,7 +249,7 @@ def build_tasks(args):
         indexers["words"] = SingleIdTokenIndexer()
     if args.elmo:
         indexers["elmo"] = ELMoTokenCharactersIndexer("elmo")
-        assert args.tokenizer in {"", "MosesTokenizer"}
+        assert args.tokenizer in tokenizers.AVAILABLE_TOKENIZERS
```
iftenney: We need to keep the emptystring case (`args.tokenizer == ""`) - per correspondence with ELMo authors, ELMo doesn't expect Moses tokenization so the default should be to just split on spaces. 

(I left `MosesTokenizer` as an option to not break other experiments, but we should actually prefer emptystring and have a warning if trying to use Moses with ELMo).

src/utils/tokenizers.py
```diff
@@ -0,0 +1,62 @@
...
+main Tokenizer class.
+
+"""
+import os
+
+from ..openai_transformer_lm.tf_original import utils as openai_utils
+from ..openai_transformer_lm.tf_original.text_utils import TextEncoder
+from nltk.tokenize.moses import MosesTokenizer as NLTKMosesTokenizer, MosesDetokenizer
+
+"""Global task registry.
```
iftenney: Old comment?

src/utils/tokenizers.py
```diff
@@ -0,0 +1,62 @@
+"""
+Tokenizer class
+To add a tokenizer, add to the below inherting from
+main Tokenizer class.
+
+"""
+import os
+
+from ..openai_transformer_lm.tf_original import utils as openai_utils
```
iftenney: Better to import these in `OpenAIBPETokenizer.__init__()` so they're not loaded if we're not using that model?

src/utils/utils.py
```diff
@@ -25,18 +23,17 @@
 from allennlp.modules.seq2seq_encoders.seq2seq_encoder import Seq2SeqEncoder
 from allennlp.common.params import Params
 
+from .tokenizers import OpenAIBPETokenizer, MosesTokenizer
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
+TOKENIZERS = {
```
iftenney: Define this in one place in `tokenizers.py` and import to this module?

src/utils/utils.py
```diff
@@ -25,18 +23,17 @@
 from allennlp.modules.seq2seq_encoders.seq2seq_encoder import Seq2SeqEncoder
 from allennlp.common.params import Params
 
+from .tokenizers import OpenAIBPETokenizer, MosesTokenizer
```
iftenney: Maybe we should move the data loaders out of `utils.py` and into another module, so that `utils.py` can stay a low-level thing that doesn't have any other dependencies witihin jiant?

probing/retokenize_edge_data.py
pruksmhc: Sure - the detokenize function is a renaming of unescape_moses,  the function is the same, but I'll keep the old utils.unescape_moses as well.

src/utils/data_loaders.py
```diff
@@ -0,0 +1,190 @@
...
+                    sent2s.append(sent2)
+
+                if idx_idx is not None:
+                    idx = int(row[idx_idx])
+                    idxs.append(idx)
+
+                sent1s.append(sent1)
+                targs.append(targ)
+
+            except Exception as e:
```
iftenney: Please remove this?

src/utils/data_loaders.py
```diff
@@ -0,0 +1,190 @@
...
+                logic_sample = tags_to_ixs(row[2], logic_to_ix_dic, ix_to_logic_dic)
+                knowledge_sample = tags_to_ixs(row[3], knowledge_to_ix_dic, ix_to_knowledge_dic)
+
+                idxs.append(row_idx)
+                lex_sem.append(lex_sem_sample)
+                pr_ar_str.append(pr_ar_str_sample)
+                logic.append(logic_sample)
+                knowledge.append(knowledge_sample)
+
+            except Exception as e:
```
iftenney: Please remove this?

src/utils/data_loaders.py
```diff
@@ -0,0 +1,190 @@
...
+            'ix_to_pr_ar_str_dic': ix_to_pr_ar_str_dic,
+            'ix_to_logic_dic': ix_to_logic_dic,
+            'ix_to_knowledge_dic': ix_to_knowledge_dic
+            }
+
+def process_sentence(tokenizer_name, sent, max_seq_len, sos_tok=SOS_TOK, eos_tok=EOS_TOK):
+    '''process a sentence '''
+    max_seq_len -= 2
+    TOKENIZER = AVAILABLE_TOKENIZERS[tokenizer_name]
+    assert max_seq_len > 0, "Max sequence length should be at least 2!"
```
iftenney: Should this assert `max_seq_len >= 2` then?

src/utils/tokenizers.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+import os
+from nltk.tokenize.moses import MosesTokenizer as NLTKMosesTokenizer, MosesDetokenizer
+
+class Tokenizer(object):
+
+    def tokenize(self, sentence):
+        raise NotImplementedError
+
+class OpenAIBPETokenizer(Tokenizer):
```
iftenney: Not necessary now, but it might be nice to have `detokenize()` implementations on the BPE tokenizers as well. Maybe add a TODO?

src/utils/data_loaders.py
```diff
@@ -0,0 +1,190 @@
+"""
+Functions having to do with loading data from output of
+files downloaded in scripts/download_data_glue.py
+
+"""
+from .tokenizers import OpenAIBPETokenizer, MosesTokenizer, AVAILABLE_TOKENIZERS
```
iftenney: Do we need to import the `OpenAIBPETokenizer` and `MosesTokenizer` symbols anymore?

src/utils/tokenizers.py
```diff
@@ -0,0 +1,59 @@
...
+    def tokenize(self, sentence):
+        return self._tokenizer.tokenize(sentence)
+
+    def detokenize(self, tokens):
+        '''Unescape Moses punctuation tokens.
+
+        Replaces escape sequences like &#91; with the original characters
+        (such as '['), so they better align to the original text.
+        '''
+        detokenizer = MosesDetokenizer()
```
iftenney: Instantiate this in `__init__()` so it's not re-loaded on every call?

src/utils/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
+import utils
+import unittest
+from io import StringIO
+
+class TestLoadTsvLablesOneSentence(unittest.TestCase):
+    def test(self):
+        mock_csv = StringIO("sentence	label",
```
iftenney: nit: use triple-quoted strings for multi-line literals?

src/utils/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+        "it 's a charming and often affecting journey . 	1",
+        "unflinchingly bleak and desperate 	0")
+        with open('temp_output.tsv', 'w', newline='') as f_output:
+            tsv_output = csv.writer(f_output, delimiter='\t')
+            tsv_output.writerow(mock_csv)
+        max_seq_len = 30
+        import pdb; pdb.set_trace()
```
iftenney: Remove PDB call before check in!

iftenney:
LGTM except for the `pdb` line :)

pruksmhc:
alright - checked!

PR #490, opened 2019-02-18
Edge probing: Add TACRED and SemEval relation-classification tasks

iftenney:
New tasks: `edges-rel-tacred` and `edges-rel-semeval`. Add to the standard data-prep and experiment scripts.

Also adds `edges-ner-tacred` for sanity-checking with the entity type tags from TACRED, although these are silver labels.


probing/deterministic_split.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python
+
+# Helper script to determinstically, but pseudo-randomly split a file by lines.
+# This is easy to to with UNIX tools (shuf and split), but the former may
+# behave differently on different systems. For repeatability, we implement this
+# using a seeded RNG in Python instead.
```
sleepinyourhat: Naïve question: How confident are we that the seed will behave the same way in future Python versions?

iftenney: Good question. It looks like they do guarantee reproducibility: "The generator’s random() method will continue to produce the same sequence when the compatible seeder is given the same seed.
" from https://docs.python.org/3/library/random.html#notes-on-reproducibility

iftenney: Added a comment to this effect.

PR #492, opened 2019-02-18
Clean up Kubernetes script

iftenney:
Remove some experiments that we're not using.


iftenney:
Force-merge: edge probing scripts only

PR #496, opened 2019-02-20
Add toleration to k8s pod config for new GPU nodes

iftenney:
GCP now adds a taint to GPU nodes to prevent non-GPU pods from scheduling on
them, so we need to add an explicit toleration to handle this.


iftenney:
Force-merging, only affects Kubernetes.

PR #497, opened 2019-02-22
Adding BERT Tokenizer

pruksmhc:
Adding wordpiece tokenizer. There will have to be further changes to add BERT to the current flow (which @W4ngatang  is working on), but this is solely for tokenization.


src/utils/data_loaders.py
```diff
@@ -3,9 +3,10 @@
 files downloaded in scripts/download_data_glue.py
 
 """
-from .tokenizers import AVAILABLE_TOKENIZERS
+from tokenizers import AVAILABLE_TOKENIZERS, BertTokenizer
```
sleepinyourhat: Why is this handled separately from the other two?

sleepinyourhat: (Add a comment to clarify.)

src/utils/data_loaders.py
```diff
@@ -173,7 +174,13 @@ def process_sentence(tokenizer_name, sent, max_seq_len, sos_tok=SOS_TOK, eos_tok
     '''process a sentence '''
     max_seq_len -= 2
     assert max_seq_len > 0, "Max sequence length should be at least 2!"
-    TOKENIZER = AVAILABLE_TOKENIZERS[tokenizer_name]
+    if "bert" in tokenizer_name:
+        import pdb; pdb.set_trace()
```
W4ngatang: delete

src/utils/tokenizers.py
```diff
@@ -53,8 +53,16 @@ def detokenize(self, tokens):
         '''
         return [self._detokenizer.unescape_xml(t) for t in tokens]
 
+class BertTokenizer(Tokenizer):
+    def __init__(self, bert_model_name):
+        from pytorch_pretrained_bert import BertTokenizer as HFBertTokenizer
```
W4ngatang: what does HF stand for?

src/tasks/tasks.py
```diff
@@ -141,8 +141,8 @@ def count_examples(self, splits=['train', 'val', 'test']):
 
     def tokenizer_is_supported(self, tokenizer_name):
         ''' Check if the tokenizer is supported for this task. '''
-        return tokenizer_name in AVAILABLE_TOKENIZERS.keys()
-
+        return tokenizer_name in AVAILABLE_TOKENIZERS.keys() or "bert" in tokenizer_name
```
W4ngatang: `AVAILABLE_TOKENIZERS.keys()` -> `AVAILABLE_TOKENIZERS`

i think it's cleaner to just have all the bert variants in `AVAILABLE_TOKENIZERS`

src/utils/data_loaders.py
```diff
@@ -173,7 +174,13 @@ def process_sentence(tokenizer_name, sent, max_seq_len, sos_tok=SOS_TOK, eos_tok
...
+        TOKENIZER = BertTokenizer(tokenizer_name)
```
W4ngatang: does this re-create the BERT tokenizer at every `process_sentence` call? Does that make preprocessing really slow?

src/utils/tokenizers.py
```diff
@@ -53,8 +53,16 @@ def detokenize(self, tokens):
...
+        do_lower_case = bert_model_name.endswith('uncased')
+        self._tokenizer = HFBertTokenizer.from_pretrained(bert_model_name, do_lower_case=do_lower_case)
+
+    def tokenize(self, sentence):
+        return self._tokenizer.tokenize(sentence)
 
 AVAILABLE_TOKENIZERS = {
     "OpenAI.BPE": OpenAIBPETokenizer(),
-    "MosesTokenizer": MosesTokenizer()
+    "MosesTokenizer": MosesTokenizer(),
```
W4ngatang: I think it's cleaner and clearer to have all the BERT tokenizers listed here, even if you are handling them specially later

W4ngatang: To avoid initializing BERT tokenizer on each process sentence call, we could just construct all four(?) BERT tokenizers here, but that might be annoying if you're not using BERT

src/tasks/tasks.py
iftenney: Another idea: instead of having `AVAILABLE_TOKENIZERS` as a dict, have it as a function and use `functools.lru_cache` to memoize. This way we can have whatever logic we want, and the cache will ensure that we only load the tokenizers we need.

Here's an example: https://github.com/jsalt18-sentence-repl/jiant/blob/master/probing/retokenize_edge_data.py#L69
https://github.com/jsalt18-sentence-repl/jiant/blob/master/probing/retokenize_edge_data.py#L104 (although we don't need the aligner part so this can be a bit simpler)

(Note that we definitely shouldn't pre-load all the BertTokenizers, because this will force downloading all the Bert models on every run.)

src/utils/data_loaders.py
iftenney: Yeah, this is going to be painful. See comment above about `lru_cache`.

src/utils/tokenizers.py
```diff
@@ -53,8 +53,16 @@ def detokenize(self, tokens):
         '''
         return [self._detokenizer.unescape_xml(t) for t in tokens]
 
+class BertTokenizer(Tokenizer):
+    def __init__(self, bert_model_name):
+        from pytorch_pretrained_bert import BertTokenizer as HFBertTokenizer
+        do_lower_case = bert_model_name.endswith('uncased')
+        self._tokenizer = HFBertTokenizer.from_pretrained(bert_model_name, do_lower_case=do_lower_case)
+
+    def tokenize(self, sentence):
```
iftenney: We don't necessarily need a wrapper class here, we can put the constructor logic in a `get_tokenizer` function (see note above).

If we want to implement a `detokenize()` function, it's not the worst idea to send a PR to the `pytorch_pretrained_bert` maintainers.

src/utils/tokenizers.py
iftenney: See note above on using a cache.

src/utils/tokenizers.py
iftenney: Huggingface?

If making a wrapper class, better to name the wrapper something like `BertTokenizerWrapper` and keep the name on the original one.

src/tasks/tasks.py
iftenney: Since the cache ensures we're not wasting computation, we can have this function be something like:
```python
def tokenizer_is_supported(self, tokenizer_name):
    return tokenizers.get_tokenizer(tokenizer_name) is not None
```

src/utils/tokenizers.py
```diff
@@ -53,8 +55,15 @@ def detokenize(self, tokens):
...
-}
+@functools.lru_cache(maxsize=8, typed=False)
+def get_tokenizer(tokenizer_name):
+    if tokenizer_name.startswith("bert"):
+        from pytorch_pretrained_bert import BertTokenizer
+        do_lower_case = tokenizer_name.endswith('uncased')
+        tokenizer = BertTokenizer.from_pretrained(tokenizer_name, do_lower_case=do_lower_case)
+    else:
+        # return the class instantiation given class name
+        assert tokenizer_name in globals(), "Tokenizer name not found in tokenizers available!"
```
iftenney: Using `globals()` is **strongly** discouraged (this will lead to all kinds of bugs you don't expect). 

Just put some if/else statements here? We're not adding new tokenizers that often so there's not a big need to avoid a couple lines of boilerplate. Plus, this way it's easy to do lazy package imports like we do for BERT.

pruksmhc:
Implemented the changes - also ran demo.conf and edgeprobing_openai.conf and got the same results for sanity check.

config/defaults.conf
```diff
@@ -164,7 +164,7 @@ tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task c
                               // handling during data loading. It's expected that all tasks use the same
                               // tokenizer; doing otherwise will cause an error.
                               // 'MosesTokenizer' is the default. Use "" for edge probing tasks to use the
-                              // original tokenization, or one of 'OpenAI.BPE',
+                              // original tokenization, or one of 'OpenAIBPETokenizer',
```
iftenney: Can we keep `OpenAI.BPE` as supported? (we shouldn't break existing configs right before a paper deadline.)

See note on `get_tokenizer()` below - we don't need these to match class names (they won't for BERT anyway) as long as the correspondence is reasonably clear.

probing/retokenize_edge_data.py
```diff
@@ -47,8 +47,8 @@
 
 # For now, this module expects MosesTokenizer as the default.
 # TODO: change this once we have better support in core utils.
-assert "MosesTokenizer" in tokenizers.AVAILABLE_TOKENIZERS
-MosesTokenizer = tokenizers.AVAILABLE_TOKENIZERS["MosesTokenizer"]
+assert tokenizers.get("MosesTokenizer") is not None
```
iftenney: nit: switch these lines, like:
```
MosesTokenizer = tokenizers.get_tokenizer("MosesTokenizer")
assert MosesTokenizer is not None
```

src/utils/tokenizers.py
pruksmhc: Gotcha - I was going to do simple if-else here but thought to reduce boilerplate.

config/defaults.conf
pruksmhc: Yep - I'm reverting the changes to OpenAI.BPE right now

probing/retokenize_edge_data.py
```diff
@@ -47,8 +47,8 @@
 
 # For now, this module expects MosesTokenizer as the default.
 # TODO: change this once we have better support in core utils.
-assert "MosesTokenizer" in tokenizers.AVAILABLE_TOKENIZERS
-MosesTokenizer = tokenizers.AVAILABLE_TOKENIZERS["MosesTokenizer"]
+MosesTokenizer = tokenizers.get_tokenizer("MosesTokenizer")
+assert tokenizers.get_tokenizer("MosesTokenizer") is not None
```
iftenney: nit: `assert MosesTokenizer is not None`

src/utils/data_loaders.py
```diff
@@ -173,9 +174,12 @@ def process_sentence(tokenizer_name, sent, max_seq_len, sos_tok=SOS_TOK, eos_tok
     '''process a sentence '''
     max_seq_len -= 2
     assert max_seq_len > 0, "Max sequence length should be at least 2!"
-    TOKENIZER = AVAILABLE_TOKENIZERS[tokenizer_name]
+    tokenizer = get_tokenizer(tokenizer_name)
+    if tokenizer_name.startswith("bert"):
```
iftenney: This seems a little odd, because it overrides the `sos_tok` and `eos_tok` that are passed in as parameters. Are these parameters actually used anywhere, or do we always take the defaults? 

If the latter, we should just remove them and have:
```
if tokenizer_name.startswith("bert-"):
    sos_tok, eos_tok = BERT_SEP_TOK, BERT_CLS_TOK
else:
    sos_tok, eos_tok = SOS_TOK, EOS_TOK

File: src/utils/tokenizers.py [hunk 13]
```diff
@@ -53,8 +54,16 @@ def detokenize(self, tokens):
...
-    "MosesTokenizer": MosesTokenizer()
-}
+@functools.lru_cache(maxsize=8, typed=False)
+def get_tokenizer(tokenizer_name):
+    log.info(f"Loading Tokenizer {tokenizer_name}")
+    if tokenizer_name.startswith("bert"):
+        from pytorch_pretrained_bert import BertTokenizer
+        do_lower_case = tokenizer_name.endswith('uncased')
+        tokenizer = BertTokenizer.from_pretrained(tokenizer_name, do_lower_case=do_lower_case)
+    else:
```
iftenney: nit: flatten, just do:
```
if tokenizer_name.startswith("bert-"):
   # do stuff
elif tokenizer_name == "OpenAI.BPE":
    return OpenAIBPETokenizer()
elif tokenizer_name == "MosesTokenizer":
    return MosesTokenizer()
elif tokenizer_name == "":
   return SpaceTokenizer()  # we should implement this or wrap something, this is the right way to use ELMo
else:
   return None
```

iftenney: Also should really have MosesTokenizer require `tokenizer_name == "MosesTokenizer"` rather than falling through? But OK to leave this and put a warning if this will break anything in-flight for ACL. (If not, we should just fix it now.)

Also see note about a space tokenizer. Can leave a TODO for this, but would be good to have this function be consistent with how edge probing behaves.

src/utils/data_loaders.py
pruksmhc: Yeah we're taking the defaults almost always

src/utils/tokenizers.py
pruksmhc: Do you mean should we exactly match MosesTokenizer as "MosesTokenizer?" I think it's better to be explicit with what type of tokenizer you want rather than defaulting on a certain tokenizer if that's what you mean.

iftenney: Yes, that's what I was thinking.

src/utils/tokenizers.py
```diff
@@ -53,8 +54,17 @@ def detokenize(self, tokens):
...
+    if tokenizer_name.startswith("bert"):
+        from pytorch_pretrained_bert import BertTokenizer
+        do_lower_case = tokenizer_name.endswith('uncased')
+        tokenizer = BertTokenizer.from_pretrained(tokenizer_name, do_lower_case=do_lower_case)
+    elif tokenizer_name == "OpenAI.BPE":
+        tokenizer = OpenAIBPETokenizer()
+    elif tokenizer_name == "MosesTokenizer":
+        tokenizer = MosesTokenizer()
+    else:
+        tokenizer = None
```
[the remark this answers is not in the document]
pruksmhc: The way we check for tokenizer support in tokenizer_is_supported function. We _could_ throw an error in this actual but we already check for if the tokenizer is valid for a task.

src/utils/tokenizers.py
```diff
@@ -53,8 +55,19 @@ def detokenize(self, tokens):
...
 
-
-AVAILABLE_TOKENIZERS = {
-    "OpenAI.BPE": OpenAIBPETokenizer(),
-    "MosesTokenizer": MosesTokenizer()
-}
+@functools.lru_cache(maxsize=8, typed=False)
+def get_tokenizer(tokenizer_name):
+    log.info(f"Loading Tokenizer {tokenizer_name}")
+    if tokenizer_name.startswith("bert"):
```
iftenney: nit: `.startswith("bert-")` for consistency.

iftenney:
Thanks for tidying this up!

PR #499, opened 2019-02-23
Cleanup edge probing tasks

iftenney:
Clean up the core list to just things in the paper


iftenney:
Force-merge: edge probing only

PR #500, opened 2019-02-26
Edge probing: move more helper code into analysis.py

iftenney:
Move a bunch of relatively self-contained helper functions from notebooks into `analysis.py`. In particular, this includes sort keys for stable ordering of tasks and experiment functions to save a lot of annoyance when re-generating tables and figures.


probing/analysis.py
```diff
@@ -18,6 +19,146 @@
 
 from typing import Iterable, Dict, List, Tuple
 
+##
+# Task list for stable ordering, and human-friendly display names.
+TASK_TO_DISPLAY_NAME = collections.OrderedDict([
+    ("pos-ontonotes", "Part-of-Speech"),
+    ("nonterminal-ontonotes", "Constituents"),
+    ("constituent-ontonotes", "BAD DO NOT USE"),
```
W4ngatang: ?
remove?

iftenney: Done

PR #502, opened 2019-02-27
Adding CoLa in and out of domain 

pruksmhc:
Adds CoLA in and out of domain for the datasets that Alex Warstadt provided. 
Developed with @HaokunLiu


src/tasks/__init__.py
```diff
@@ -16,7 +16,7 @@
 ##
 # Task lists for handling as a group; these names correspond to the keys in 
 # the task registry.
-ALL_GLUE_TASKS = ['sst', 'cola', 'mrpc', 'qqp', 'sts-b',
+ALL_GLUE_TASKS = ['sst', 'cola', 'cola-in-domain', 'cola-out-domain', 'mrpc', 'qqp', 'sts-b',
```
sleepinyourhat: Let's not do this, at least for now. The GLUE version of CoLA is the version with only a single evaluation set. If people want the full CoLA eval, they can explicitly add it to the task list.

src/tasks/tasks.py
```diff
@@ -367,6 +367,56 @@ def get_metrics(self, reset=False):
...
+    def __init__(self, path, max_seq_len, name="train_in_domain", **kw):
+        super(
+            CoLAInDomainTask,
+            self).__init__(
+            path,
+            max_seq_len,
+            name=name,
+		**kw)
+
+    def load_data(self, path, max_seq_len):
```
iftenney: Fine as-is, but it might be possible to save some boilerplate by adding parameters to `@register_task` that are passed to the constructor.

See https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/tasks/edge_probing.py#L245 and https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/tasks/edge_probing.py#L57 for an example

pruksmhc:
Haokun is working on a slight change to the main do_full_eval task to incorporate the 3 scripts. I'm currently focusing on helping Alex close out bert-tasks

pruksmhc:
changes to this code addressing the review comments were opened up in Haokun's PR so closing this

PR #506, opened 2019-03-04
Update inference with tokenizer name

zphang:


zphang:
(Updated - I missed another `process_sentence` call.)

`process_sentence` was recently refactored to add an additional first argument for tokenizer_name. This PR fixes the call to reflect that change.

I thought that, since the order of non-keyword argument is now modified, it would be better to be explicit about the arguments in the function call. I can switch to non-keyword arguments if necessary.

PR #510, opened 2019-03-10
Add additional columns to inference output

zphang:
See #507

Issue #508, opened 2019-03-06
Support labeled data in inference script


sleepinyourhat:
Add an option to run with labeled datasets (same format as would be used for a normal task dataset) and see accuracy/F1/MCC in addition to individual example labels.

@zphang @anhad13 – any interest?

zphang:
This means inference + evaluation? 

Should this point to the specific train/dev.tsv rather than a full task folder? My thoughts are that this will piggy-back off the data-loading/metrics specific in the tasks.

sleepinyourhat:
Not sure what you mean. The goal would be to make it easy to swap out the test/dev TSVs with new TSVs, and still have preprocessing etc. handled correctly. We can almost do that with the current inference executable, but it won't automatically do evaluation (right?).

zphang:
Here's my plan:

- Additional argument for `input_format`: plain, train, dev, test (plain is just one sentence per line, train/dev/test will read in the corresponding format based on the `tasks.py` definition for that task - just CoLA for now)
- Additional argument for `eval_output_path` to optionally output a JSON with the metrics computed based on `tasks.py`, only works if `input_format`is `train` or `dev` (since only those two formats have labels)

sleepinyourhat:
input_format: Sounds good.
eval_output_path: At least for now, it'd probably be reasonable to just write these to the end of the log or the end of the output file. Up to you, though.

pruksmhc:
What's the status of this?

PR #511, opened 2019-03-10
Inference labeled data

zphang:
See #508. Merge after #510.

### DIFF
--- a/cola_inference.py
+++ b/cola_inference.py
@@ -13,8 +13,18 @@ To process a corpus:
     python cola_inference.py \
         --config_file PATH_TO_CONFIG_FILE \
         --model_file_path PATH_TO_FILE_PATH \
-        --inference mode corpus \
-        --input_path PATH_TO_INPUT_CORPUS
+        --inference_mode corpus \
+        --input_path PATH_TO_INPUT_CORPUS \
+        --output_path PATH_TO_WRITE_OUTPUT
+
+To process+evaluate (e.g.) the CoLA dev set:
+
+    python cola_inference.py \
+        --config_file PATH_TO_CONFIG_FILE \
+        --model_file_path PATH_TO_FILE_PATH \
+        --inference_mode corpus \
+        --input_path PATH_TO_INPUT_CORPUS \
+        --input_format dev \
         --output_path PATH_TO_WRITE_OUTPUT
 
 (Ensure that the repository is in your PYTHONPATH when running this script.)
@@ -22,6 +32,7 @@ To process a corpus:
 '''
 # pylint: disable=no-member
 import argparse
+import json
 import numpy as np
 import os
 import pandas as pd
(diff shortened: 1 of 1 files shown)


cola_inference.py
```diff
@@ -212,5 +254,40 @@ def take_one(ls):
     return ls[0]
 
 
+def load_cola_data(input_path, task, input_format, max_seq_len):
```
sleepinyourhat: Add a comment on why you need this here, rather than using the existing jiant loader.

cola_inference.py
```diff
@@ -212,5 +254,40 @@ def take_one(ls):
...
+    elif input_format == "test":
+        data = load_tsv(task.tokenizer_name, input_path, max_seq_len,
+                        s1_idx=1, s2_idx=None, has_labels=False, return_indices=True, skip_rows=1)
+        tokens, labels = data[0], None
+    else:
+        raise KeyError(input_format)
+    return tokens, labels
+
+
+def get_cola_metrics(logits, preds, labels, task):
```
sleepinyourhat: Ditto. Why does this need to be in this file?

cola_inference.py
zphang: The current jiant loader leads data for all three phases at once, hence it is not possible to separate that functionality:

https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/tasks/tasks.py#L354-L359

Once possibility would be to expose e.g. `get_train`/`get_dev`/`get_test` methods specifically for the CoLA Task class. This can be done cleanly so as to not affect the regular usage of the CoLA Task/data loading or the broader Task api.

cola_inference.py
zphang: Similar to the above, the logic for computing metrics specific to CoLA is deeply embedded in the model code:

https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/models.py#L672-L675

Unlike the above, I don't think there is currently a clean way to expose the same functionality. I recommend keeping this as is given the simplicity of the logic here, as well as the potential for adding additional metrics specifically for CoLA evaluation in the future.

cola_inference.py
sleepinyourhat: Hrm. I guess we can merge what you have now, but create an issue to clean this up.

cola_inference.py
sleepinyourhat: SGTM.

sleepinyourhat:
Let me know if this is ready to merge.

zphang:
Should be good to merge. I will make an issue regarding the data loading.

sleepinyourhat:
Ergh—I merged the two in the wrong order, and GitHub didn't process the revert correctly. Mind creating a new PR with these changes? Sorry!

zphang:
Added #518.

PR #512, opened 2019-03-10
Subset evaluation

HaokunLiu:
based on Yada's new data_loader, add cola in/out/full domain and Alex's cola dev set annotation to task. add ling3340 baseline experiments config files to config.


config/cola-elmo.conf
```diff
@@ -0,0 +1,50 @@
+
```
sleepinyourhat: Add a comment explaining what this file is for.

sleepinyourhat: You should have a note for everything that isn't obvious. Why this learning rate, for example?

config/cola-openai.conf
```diff
@@ -0,0 +1,20 @@
+
```
sleepinyourhat: Same as above: Looks good, but add a few comments.

config/cola-openai.conf
```diff
@@ -0,0 +1,20 @@
...
+include "cola-elmo.conf"
+
+
+// Training setting
+
+do_target_task_training = 0
+allow_reuse_of_pretraining_parameters = 1
+batch_size = 2
```
sleepinyourhat: Why so small?

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,20 @@
+
+# random-elmo
+rm *.out
```
sleepinyourhat: ?

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,20 @@
...
+JIANT_OVERRIDES="exp_name = cola-elmo-baseline, run_name = random-elmo, allow_untrained_encoder_parameters = 1, allow_missing_task_map = 1, do_pretrain = 0" JIANT_CONF="config/cola-elmo.conf" sbatch ~/prince.sbatch
```
sleepinyourhat: I don't think you've shared the sbatch file.

sleepinyourhat: Add some explanation of what that file needs to be, at least. And mention how it differs from the existing ones.

scripts/download_glue_data.py
```diff
@@ -49,8 +49,26 @@ def download_and_extract(task, data_dir):
     with zipfile.ZipFile(data_file) as zip_ref:
         zip_ref.extractall(data_dir)
     os.remove(data_file)
+    if task == "SNLI":
```
sleepinyourhat: This should be its own PR—we're not dealing with CoLA here.

src/models.py
```diff
@@ -673,11 +673,12 @@ def _single_sentence_forward(self, batch, task, predict):
                 task.scorer2(logits, labels)
                 _, preds = logits.max(dim=1)
                 task.scorer1(labels, preds)
+            elif isinstance(task, CoLAAnalysisTask):
+                task.update_metrics(logits, labels, tagmask=batch['tagmask'])
             else:
                 task.scorer1(logits, labels)
                 if task.scorer2 is not None:
                     task.scorer2(logits, labels)
-
```
sleepinyourhat: Put this back in.

src/tasks/tasks.py
```diff
@@ -25,13 +25,13 @@
```
sleepinyourhat: It looks like there are some changes in here that aren't related to CoLA. Either get rid of those or explain why they have to be here.

src/utils/utils.py
```diff
@@ -10,7 +10,6 @@
 import logging
 import codecs
 import time
-
```
sleepinyourhat: Revert.

config/cola-openai.conf
HaokunLiu: I tried several different batch size, ranging from 2 to 6. 6 is as large as we can afford on k80 but will out-of-memory on smaller gpus. 2 can run on any gpu in prince, and it performs better. Some researches show smaller bs results in better generalization.

scripts/cola_baseline.sh
HaokunLiu: Sorry, this one was because I made too many .out files during debugging. This shouldn't be in the commited version.

scripts/download_glue_data.py
HaokunLiu: This wasn't me. I forked from load_tsv branch, and it was there. I presume this diff won't exist once load_tsv is merged?

src/models.py
HaokunLiu: Sure

src/tasks/tasks.py
HaokunLiu: create_subset_scorer, update_subset_scorers, collect_subset_scores are mine. I put them outside cola because they can be reused for other tasks that need find-grained metrics. 
Any other changes outside cola are from load_tsv branch.

src/utils/utils.py
HaokunLiu: This wasn't me. This is also from load_tsv.

scripts/cola_baseline.sh
HaokunLiu: > I don't think you've shared the sbatch file.

No, I didn't. That file was used to set up environment, some path settings etc. Should I put that into jiant?

config/cola-openai.conf
sleepinyourhat: I've mostly seen papers showing the opposite effect for batch size, but it's okay to leave it in if it works.

scripts/cola_baseline.sh
sleepinyourhat: Up to you, but this isn't appropriate as-is. The master copy of the repo needs to be clearly usable by anyone. So either share the sbatch file (and remove any paths that are specific to you) or remove all references to it.

scripts/download_glue_data.py
sleepinyourhat: That already was merged (https://github.com/jsalt18-sentence-repl/jiant/pull/477), so something is wrong here. Make sure you don't have any conflicts with master.

HaokunLiu:
This is weird. 
I can see, with my own naked eyes, my files don't conflict with master.
I don't know why this file changes tool isn't comparing to the latest, real mater branch.

src/tasks/tasks.py
```diff
@@ -101,6 +101,43 @@ def _make_instance(input1, input2, labels, idx):
     #  return list(instances)
     return instances  # lazy iterator
 
+def create_subset_scorers(count, scorer_type, **args_to_scorer):
+    '''
+    Inputs
```
pruksmhc: Nit: For code consistency, put this as Parameters and outputs as "Returns"

pruksmhc: same for all below

HaokunLiu:
good, no conflicts now.

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
+// Run elmo model on cola-analysis dataset
```
sleepinyourhat: Two extra suggestions:
- Move this to a cola directory, with a readme explaining that this is for this seminar. These configs are probably useful to others, but they're distinctive enough that we should probably keep them separate.
- Proofread for typos. Epoches => epochs. numberof => number of.

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
...
+allow_reuse_of_pretraining_parameters = 1
+lr_patience = 4  // number of epoches between last validation improvement and lr annealing 
+patience = 20  // numberof epoches between last validation improvement and early stopping
+max_vals = 10000
+
+// Eval setting
+
+write_preds = 'test'
+
+
```
sleepinyourhat: Be consistent about spacing

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
...
+word_embs_file = ""
+fastText_model_file = ""
+
+// Paths
+
+exp_name = debug  // avoid messing up other experiment records
+run_name = debug  // overwrite this when running models
+
+// Data setting
+
```
sleepinyourhat: Why?

config/cola-elmo.conf
sleepinyourhat: 'Run' suggests just evaluating. => Train and run.

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
...
+reload_vocab = 1
+reload_tasks = 1
+reload_indexing = 1
+reindex_tasks = cola-analysis
+
+// Training setting
+
+load_model = 0
+
+pretrain_tasks = cola-analysis
```
sleepinyourhat: This doesn't look right. You shouldn't ever pretrain on an analysis task.

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
...
+do_pretrain = 1  // train the model on top of embedding layer
+
+target_tasks = cola-analysis
+do_target_task_training = 1  // tune the parameters on target task
+do_full_eval = 1  // evaluation the model after training
+
+lr = 3e-4  // this performs reasonablely well in practice
+allow_untrained_encoder_parameters = 0  // set this to 1 when running random-elmo
+allow_reuse_of_pretraining_parameters = 1
+lr_patience = 4  // number of epoches between last validation improvement and lr annealing
```
sleepinyourhat: Why did you have to change these from the default? Add a note.

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
...
+lr = 3e-4  // this performs reasonablely well in practice
+allow_untrained_encoder_parameters = 0  // set this to 1 when running random-elmo
+allow_reuse_of_pretraining_parameters = 1
+lr_patience = 4  // number of epoches between last validation improvement and lr annealing 
+patience = 20  // numberof epoches between last validation improvement and early stopping
+max_vals = 10000
+
+// Eval setting
+
+write_preds = 'test'
```
sleepinyourhat: Why test? We can't really do anything with the test set, since we don't have the labels.

config/cola-elmo.conf
```diff
@@ -0,0 +1,51 @@
...
+
+
+// Model settings
+
+elmo = 1
+elmo_chars_only = 0
+sep_embs_for_skip = 1
```
sleepinyourhat: Fix final line end.

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,25 @@
+# baseline experiments on cola.
```
sleepinyourhat: See comments above.

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,25 @@
...
+# don't actually run this sh file as a whole,
```
sleepinyourhat: Why not?

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,25 @@
...
+# yes, this sounds stupid, I know.
+
+# set up environment
+module purge
+module load tensorflow/python3.6/1.5.0
+module swap python3/intel anaconda3/5.3.1
+source activate jiant
+export NFS_PROJECT_PREFIX=/scratch/hl3236/ling3340
+export JIANT_DATA_DIR=/home/hl3236/data
+export WORD_EMBS_FILE=/home/hl3236/misc/crawl-200d-2M.vec
```
sleepinyourhat: These paths should be in your personal path setup script, not here.

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,25 @@
+# baseline experiments on cola.
+# don't actually run this sh file as a whole,
+# copy the lines you need and paste it to the command line.
+# yes, this sounds stupid, I know.
+
+# set up environment
+module purge
+module load tensorflow/python3.6/1.5.0
+module swap python3/intel anaconda3/5.3.1
```
sleepinyourhat: These are specific to Prince, so they shouldn't be checked in here.

scripts/cola_baseline.sh
```diff
@@ -0,0 +1,25 @@
...
+source activate jiant
+export NFS_PROJECT_PREFIX=/scratch/hl3236/ling3340
+export JIANT_DATA_DIR=/home/hl3236/data
+export WORD_EMBS_FILE=/home/hl3236/misc/crawl-200d-2M.vec
+
+# random-elmo
+srun --gres=gpu:k80:1 python main.py --config_file "config/cola-elmo.conf" --overrides "exp_name = cola-elmo-baseline, run_name = random-elmo, allow_untrained_encoder_parameters = 1, allow_missing_task_map = 1, do_pretrain = 0"
```
sleepinyourhat: These are specific to Prince, so they shouldn't be checked in here. Consider just giving the plain Python commands.

src/tasks/tasks.py
```diff
@@ -101,6 +101,45 @@ def _make_instance(input1, input2, labels, idx):
     #  return list(instances)
     return instances  # lazy iterator
 
+def create_subset_scorers(count, scorer_type, **args_to_scorer):
+    '''
```
sleepinyourhat: Add a comment saying what this is for. (Same below.)

src/tasks/tasks.py
```diff
@@ -101,6 +101,45 @@ def _make_instance(input1, input2, labels, idx):
...
+        labels: a (bs, *) tensor, ground truth
+        tagmask: a (bs, N_tag) 0-1 tensor, indicating tags of each sample
+    '''
+    for tid, scorer in enumerate(scorer_list):
+        subset_idx = torch.nonzero(tagmask[:, tid]).squeeze(dim=1)
+        subset_estimations = estimations[subset_idx]
+        subset_labels = labels[subset_idx]
+        if len(subset_idx) > 0:
+            scorer(subset_estimations, subset_labels)
+    return
```
sleepinyourhat: Do you need this?

src/tasks/tasks.py
```diff
@@ -366,6 +405,76 @@ def get_metrics(self, reset=False):
...
+            max_seq_len=max_seq_len,
+            tag_eval=True)
+        
+        tr_data = tsv_loader.load_file(os.path.join(path, "train_analysis.tsv"),
+            s1_cid=3, s2_cid=None, label_cid=2, skip_rows=1, tag2cid_dict={'Domain': 1})
+        val_data = tsv_loader.load_file(os.path.join(path, "dev_analysis.tsv"),
+            s1_cid=3, s2_cid=None, label_cid=2, skip_rows=1, tag2cid_dict={
+                'Domain': 1, 'Simple': 4, 'Pred': 5, 'Adjunct': 6, 'Arg Types': 7,
+                'Arg Altern': 8, 'Imperative': 9, 'Binding': 10, 'Question': 11, 'Comp Clause': 12,
+                'Auxillary': 13, 'to-VP': 14, 'N, Adj': 15, 'S-Syntax': 16, 'Determiner': 17, 'Violations': 18})
```
sleepinyourhat: Aren't there more tags than this?

src/utils/data_loaders.py
```diff
@@ -153,6 +153,103 @@ def targs_to_idx(col_name):
             'ix_to_knowledge_dic': ix_to_knowledge_dic
             }
 
+class TSVLoader(object):
```
sleepinyourhat: How does this relate to what Yada pushed recently?

src/tasks/tasks.py
HaokunLiu: Yes, these are only the coarse tags to column index dictionary. I reorganized cola_dev_annotated following the mnli-diagnostic format. If a sample contains multiple fine tags under one coarse tag, the fine tags are listed under corresponding column, separated by ';' .

src/tasks/tasks.py
HaokunLiu: Yes, these functions can be used to easily implement fine grained evaluation by any task that needs it (like mnli-diagnostic), they are not restricted to any particular tag or score metric.

src/utils/data_loaders.py
HaokunLiu: Previously, tsv loader didn't use pandas. I modified her pandas version.

config/cola-elmo.conf
HaokunLiu: Alex did give me the test labels, to make sure the program runs correctly (see if final run test results match up with the published results). All hyper-parameters are adjusted according to dev performance before the final run, following experiment norms.

config/cola-elmo.conf
HaokunLiu: sorry, it should be 'override'

config/cola-elmo.conf
HaokunLiu: Cola-analysis have exactly the same amount of data as cola. The only difference is it has tags. For elmo experiment, I followed the structure setting from elmo's friend, elmo + 2 layers of bi-lstm + mlp. If we turn off pretrain, the two layers bi-lstm will remain random weight.
I don't understand why we should not pretrain on cola-analysis.

HaokunLiu: Does the cola-analysis I described fit your definition of an analysis task?

scripts/cola_baseline.sh
HaokunLiu: Because it is srun, it will run one after another.
Now we have removed the prince related command, we don't need this line anymore.

src/tasks/tasks.py
sleepinyourhat: Got it. Does this code read those fine tags? We should.

src/utils/data_loaders.py
sleepinyourhat: I'm not sure I understand. We really shouldn't have two TSV readers unless they're doing very different things. How is this different from what's currently in master?

HaokunLiu: Does this looks better now? I removed the TSVLoader object, put its functionality into load_tsv.

src/tasks/tasks.py
HaokunLiu: Yes, all the fine tags will be loaded as "coarse__fine". e.g. "Determiner__Partitive"

config/cola-elmo.conf
sleepinyourhat: I guess it works, it's just not the intended use. We should set up training on cola, but evaluation on cola-analysis. If that's not possible right now, though, just make it add comments here _and_ in the Python to explain that the analysis task contains the normal training set.

HaokunLiu: Right now, jiant use different classifiers for different task, so we cannot train on cola and directly evaluate it on cola-analysis.

HaokunLiu:
Ready!

PR #514, opened 2019-03-11
Trainer for a fixed number of epochs

W4ngatang:
Includes:

* train for a maximum number of epochs (passes over all of a task's data)

* refactoring of stop check in trainer

* removal of a possible infinite loop bug

### DIFF
--- a/config/defaults.conf
+++ b/config/defaults.conf
@@ -124,8 +124,8 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
 val_interval = 1000  // Number of steps between validation checks.
                      // Must be divisible by bpp_base (which is usually 1).
 max_epochs = -1 // If positive, maximum number of epochs (full pass over a task's training data) to perform
-                         // Stopping is based on hitting either the max_epochs (for each task) or any other stopping criterion
-                         // (max validation over all tasks, minimum learning rate hit)
+                // Stopping is based on hitting either the max_epochs (for each task) or any other stopping criterion
+                // (max validation over all tasks, minimum learning rate hit)
 max_vals = 1000  // Maximum number of validation checks. Will stop once this limit has been reached.
 bpp_base = 1  // In multitask learning, number of steps to train each task before sampling a fresh task.
 patience = 5  // Patience in early stopping. Training will stop if performance does not improve at all in patience + 1 validations.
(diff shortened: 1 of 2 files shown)


src/trainer.py
```diff
@@ -312,7 +317,11 @@ def train(self, tasks, stop_metric,
         task_infos, metric_infos = self._setup_training(tasks, batch_size, train_params,
                                                         optimizer_params, scheduler_params, phase)
 
-        if shared_optimizer:  # if shared_optimizer, ignore task_specific optimizers
+        if shared_optimizer:  # If shared_optimizer, ignore task_specific optimizers
+            if self._max_epochs_per_task:
+                # TODO(Alex or Yada): make this generalizable to not depend
```
W4ngatang: what does this mean?

W4ngatang:
Ready for review, though I need to run a test.

src/trainer.py
pruksmhc: Actually revert this, I think this was from when I was trying to figure out how _max_epochs_per_task worked

config/defaults.conf
```diff
@@ -123,6 +123,7 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
                        // end of main.py that is invoked by do_full_eval.
 val_interval = 1000  // Number of steps between validation checks.
                      // Must be divisible by bpp_base (which is usually 1).
+max_epochs_per_task = None // If positive, maximum number of epochs (full pass over all training data) to perform
```
sleepinyourhat: Explain how this relates to max_vals. We run until _either_ criterion is hit, right? Might also be worth adding an option to set max_vals to None, with the same effect.

sleepinyourhat: Also, why 'per task'?

src/trainer.py
```diff
@@ -312,7 +317,11 @@ def train(self, tasks, stop_metric,
...
+                # on max_epochs_per_tasks.
+                n_epoch_steps = sum([info["n_tr_batches"] for info in task_infos.values()])
```
sleepinyourhat: This is never used, delete.

src/trainer.py
```diff
@@ -699,6 +708,15 @@ def _check_stop(self, epoch, stop_metric, tasks):
         ''' Check to see if should stop '''
         task_infos, metric_infos = self._task_infos, self._metric_infos
         g_optimizer = self._g_optimizer
+
+        if self._max_epochs_per_task > 0:
+            for task in tasks:
+                task_info = task_infos[task.name]
+                n_epochs_trained = task_info['total_batches_trained'] / task_info['n_tr_batches']
+                if n_epochs_trained >= self._max_epochs_per_task:
+                    log.info("Maximum batches trained on %s.", task.name)
```
sleepinyourhat: => 'epochs'

PR #516, opened 2019-03-11
Add alt versions of all tasks and qnliv2

W4ngatang:
Add alt versions of all GLUE tasks (so we don't have to the two runs for pretraining) and QNLIv2 (though maybe we should delete this b/c in the GLUE data distribution it's still QNLI)

### DIFF
--- a/src/tasks/tasks.py
+++ b/src/tasks/tasks.py
@@ -311,6 +311,7 @@ class RankingTask(Task):
 
 
 @register_task('sst', rel_path='SST-2/')
+@register_task('sst-alt', rel_path='SST-2/')
 class SSTTask(SingleClassificationTask):
     ''' Task class for Stanford Sentiment Treebank.  '''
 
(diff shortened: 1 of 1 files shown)


src/tasks/tasks.py
```diff
@@ -831,6 +835,8 @@ def load_data(self, path, max_seq_len):
 
 @register_task('qnli', rel_path='QNLI/')
 @register_task('qnli-alt', rel_path='QNLI/')  # second copy for different params
+@register_task('qnliv2', rel_path='QNLI/')
+@register_task('qnliv2-alt', rel_path='QNLI/')  # second copy for different params
```
sleepinyourhat: Document or cut before merging.

W4ngatang:
This avoids needing to do GLUE pretraining + fine-tuning as two runs, which was annoying.

sleepinyourhat:
Confused—did you mean to merge?

PR #517, opened 2019-03-11
Add first state pooling

W4ngatang:
Add first state "pooling" (i.e. take first hidden state of sequence). Relevant for BERT integration.


W4ngatang:
We can make a user-visible config option. I was planning on doing a PR with some pooling refactors anyways, so I'll do it there.

PR #520, opened 2019-03-13
Remove pooler from_params attribute + refactor pair sentence attention

W4ngatang:
Remove pooler `from_params` attribute because it really wasn't that hard to just use the normal constructor and it didn't expose the `pool_type` arg (relevant for BERT).

Refactor pair sentence attention constructor because it was weird.


W4ngatang:
Tested all the cases w/ and w/o pair attn

PR #521, opened 2019-03-13
Add tokenizer names to various tasks

W4ngatang:
Includes

* Make all tasks use `task._tokenizer_name` when processing data

* Remove removed usage of SOS/EOS tokens in calling `process_sentence`

* Create WikiText 2 skipthought task (mainly for debugging)

* Update comments

* Update some paths

### DIFF
--- a/src/tasks/mt.py
+++ b/src/tasks/mt.py
@@ -66,11 +66,9 @@ class MTTask(SequenceGenerationTask):
                 row = row.strip().split('\t')
                 if len(row) < 2 or not row[0] or not row[1]:
                     continue
-                src_sent = process_sentence(row[0], self.max_seq_len,
-                                            tokenizer_name=self._tokenizer_name)
+                src_sent = process_sentence(self._tokenizer_name, row[0], self.max_seq_len)
                 # Currently: force Moses tokenization on targets
-                tgt_sent = process_sentence(row[1], self.max_seq_len,
-                                            tokenizer_name="MosesTokenizer")
+                tgt_sent = process_sentence("MosesTokenizer", row[1], self.max_seq_len)
                 yield (src_sent, tgt_sent)
 
     def get_sentences(self) -> Iterable[Sequence[str]]:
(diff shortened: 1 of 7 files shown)


src/tasks/nli_probing.py
```diff
@@ -50,12 +51,17 @@ def __init__(self, path, max_seq_len, name, probe_path="probe_dummy.tsv",
...
-                           s1_idx=1, s2_idx=2, targ_idx=None, skip_rows=0)
-        val_data = load_tsv(os.path.join(path, 'lexnegs.tsv'), max_seq_len,
-                            s1_idx=8, s2_idx=9, targ_idx=10, targ_map=targ_map, skip_rows=1)
-        te_data = load_tsv(os.path.join(path, 'test_dummy.tsv'), max_seq_len,
-                           s1_idx=1, s2_idx=2, targ_idx=None, skip_rows=0)
+        label_fn = targ_map.__getitem__
+        tr_data = load_tsv(data_file=os.path.join(path, 'train_dummy.tsv'), max_seq_len=max_seq_len,
+                           s1_idx=1, s2_idx=2, label_idx=None, skip_rows=0,
+                           tokenizer_name=self._tokenizer_name)
+        # TODO(?): should this use probe_path?
```
W4ngatang: @najoungkim should these use `probe_path`? It's passed into the function but not used.

src/tasks/mt.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
         self.max_targ_v_size = max_targ_v_size
         self.target_indexer = {"words": SingleIdTokenIndexer(namespace=self._label_namespace)}
         self.files_by_split = {split: os.path.join(path, "%s.txt" % split) for
-                               split in ["train", "valid", "test"]}
+                               split in ["train", "val", "test"]}
```
sleepinyourhat: Why the change?

src/tasks/mt.py
```diff
@@ -66,13 +66,11 @@ def load_data(self, path):
                 row = row.strip().split('\t')
                 if len(row) < 2 or not row[0] or not row[1]:
                     continue
-                src_sent = process_sentence(row[0], self.max_seq_len)
-                # target sentence sos_tok, eos_tok need to match Seq2SeqDecoder class
-                tgt_sent = process_sentence(
-                    row[1], self.max_seq_len,
-                    sos_tok=allennlp_util.START_SYMBOL,
-                    eos_tok=allennlp_util.END_SYMBOL,
```
sleepinyourhat: I assume these are set as defaults already?

src/tasks/mt.py
```diff
@@ -128,7 +125,7 @@ def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
                          max_targ_v_size=max_targ_v_size, name=name,
                          **kw)
         self._label_namespace = None
-        self.target_indexer = {"words": SingleIdTokenIndexer("tokens")}
```
sleepinyourhat: Again, is this a no-op change?

src/tasks/mt.py
W4ngatang: "val" is consistent with the other tasks; not sure why we were using "valid" before :/

src/tasks/mt.py
W4ngatang: yea, and we removed these as args in the `process_sentence` function

src/tasks/mt.py
W4ngatang: yes (by default that arg is "tokens"), but it's worth being explicit

PR #524, opened 2019-03-14
Add BERT Adam support

W4ngatang:
Add an option to use the BERT optimizer, with and without a max number of epochs to train for.


src/trainer.py
```diff
@@ -313,7 +319,11 @@ def train(self, tasks, stop_metric,
                                                         optimizer_params, scheduler_params, phase)
 
         if shared_optimizer:  # If shared_optimizer, ignore task_specific optimizers
-            g_optimizer = Optimizer.from_params(train_params, copy.deepcopy(optimizer_params))
+            optimizer_params = copy.deepcopy(optimizer_params)
```
W4ngatang: FYI: we deep copy because AllenNLP pops from these dicts when the values are used, so if you want to reuse them, you have to have copies.

PR #525, opened 2019-03-15
Fix dataloading issues with MNLI and load_tsv

W4ngatang:
* in `load_tsv`, also check if sent2 has length 0

* `targ_idx` -> `label_idx`

* `idx_idx` -> `return_indices=True`


src/tasks/tasks.py
```diff
@@ -720,11 +720,11 @@ def load_data(self, path, max_seq_len):
         val_data = tuple(val_data)
 
         te_matched_data = load_tsv(self._tokenizer_name, os.path.join(path, 'test_matched.tsv'), max_seq_len,
-                                   s1_idx=8, s2_idx=9, targ_idx=None, idx_idx=0, skip_rows=1)
+                                   s1_idx=8, s2_idx=9, has_labels=False, return_indices=True, skip_rows=1)
```
sleepinyourhat: If it's easy, add a comment to load_tsv explaining what has_labels does. Missing now.

PR #526, opened 2019-03-17
BERT -> Gap Coreference 

pruksmhc:
Changes:
-Added GAP coreference to Jiant
-Added preprocessing script which aligns the span indices from GAP to each tokenization 
-Shifted get_tags function to utils.py rather than data_loaders.py (refactoring change)
-GAP coreference reports subset scores for female, male, and bias as well as overall

The experiment:
BERT -> (top-layer-activation) -> Span extraction -> Classifier  -> Output (binary classification)

The model can access only embeddings within given spans, such as a predicate-argument pair, and must predict properties, such as semantic roles, which typically require whole-sentence context.
We finetune BERT on the GAP dataset and use task specific embeddings rather than the frozen embeddings. 

Note: Some of the files are under probing/ - this is for convenience. There will be another refactor such that probing/ will become integrated into the rest of codebase


config/benchmarkssuperglue/gap_bert.conf
```diff
@@ -0,0 +1,39 @@
...
+bert_fine_tune = 1
+batch_size = 8
+max_seq_len = 511
+cove = 0
+word_embs = "none"
+elmo = 0
+
+sent_enc = "null"
+skip_embs = 1  // forward embeddings from lower level. 
+sep_embs_for_skip = 1  // use task embeddings since we skip the generic ones.
```
pruksmhc: In the task-specific classifier, sep_embs_for_skip is used to use the task-specific embedding

config/benchmarkssuperglue/gap_bert.conf
```diff
@@ -0,0 +1,39 @@
+// This imports the defaults, which can be overridden below.
```
sleepinyourhat: Nit: Use _ to separate words in dir names.

config/benchmarkssuperglue/gap_bert.conf
```diff
@@ -0,0 +1,39 @@
...
+include "../defaults.conf"  // relative path to this file
+
+project_dir = ${JIANT_PROJECT_PREFIX}
+exp_name = "bert_gap_running"  // configure this
```
sleepinyourhat: What does 'running' mean?

config/benchmarkssuperglue/gap_bert.conf
```diff
@@ -0,0 +1,39 @@
...
+run_name = "run"  // default
+
+pretrain_tasks = "gap-coreference"  // empty: don't run main training phase
+target_tasks = "gap-coreference"   // train classifier only
+
+// Eval will use task-specific params.
```
sleepinyourhat: Most of the below should live in a shared config file that inherits from defaults, as final.conf does. (And many of these are already set in defaults, and should be skipped.)

config/benchmarkssuperglue/gap_bert.conf
sleepinyourhat: Not sure I understand. Isn't everything task-specific with BERT?

probing/gap_preprocess.py
```diff
@@ -0,0 +1,137 @@
+import retokenize_edge_data as retokenize
```
sleepinyourhat: This shouldn't live in probing/, since it's a real target task.

probing/gap_preprocess.py
```diff
@@ -0,0 +1,137 @@
...
+import sys
+import argparse
+import pandas as pd
+import pickle
+from enum import Enum
+
```
sleepinyourhat: Add a brief introduction on what GAP is and what extra preprocessing you need for it.

sleepinyourhat: (And generally add a few general comments below to explain the structure of what's happening here.)

probing/gap_preprocess.py
```diff
@@ -0,0 +1,137 @@
...
+    current_tokenization.extend(span_tokens)
+
+    text = " ".join(current_tokenization)
+    if orig_span1_end > orig_span2_start:
+        return new_span2start, new_span2end, new_span1start, new_span1end,text
+    return  new_span1start, new_span1end, new_span2start, new_span2end, text
+
+def align_spans(split, tokenizer_name, data_dir):
+    """
+        This processes the dataset into the form that edge probing can read in
```
sleepinyourhat: This isn't a probing task. If we reuse code that's currently framed as being specific to edge probing, we should move or rename that existing code to make it clear that it's just about _edges_, not _probing_.

probing/gap_preprocess.py
```diff
@@ -0,0 +1,137 @@
...
+         "I like Bob Sutter yeah " becomes soemthing like
+        ["I", "like", "Bob", "Sut", "ter", "yeah"]
+        The gold labels given in GAP ha
+        s noun index as [7:16], however, with tokenization, we
+        want tokenization of [2:4]
+
+        Output: 
+        A TSV file readable
+    """
+    gap_text = pd.read_csv(data_dir+"gap-" + split + ".tsv",  header = 0, delimiter="\t")
```
sleepinyourhat: Run autopep8. Spacing nit.

src/models.py
```diff
@@ -666,6 +666,7 @@ def _single_sentence_forward(self, batch, task, predict):
             else:
                 labels = batch['labels'].squeeze(-1)
             out['loss'] = F.cross_entropy(logits, labels)
+            # update scores for each task
```
sleepinyourhat: Be consistent about newlines before comments, and capitalization in comments.

src/modules/edge_probing.py
```diff
@@ -165,6 +165,7 @@ def forward(self, batch: Dict,
             out['loss'] = self.compute_loss(logits[span_mask],
                                             batch['labels'][span_mask],
                                             task)
+            task.update_subset_metrics(logits[span_mask], batch["labels"][span_mask], tagmask=batch['tagmask'])
```
sleepinyourhat: Why the change here?

src/tasks/edge_probing.py
```diff
@@ -11,18 +11,26 @@
     BooleanAccuracy, F1Measure
```
sleepinyourhat: See general comments about probing code above.

probing/gap_preprocess.py
sleepinyourhat: Context: Probing tasks are tasks on which don't reflect any intended application, and for which we generally train only a very small subset of the full parameters (so as to highlight what knowledge is already in the model). Neither of those conditions applies here.

pruksmhc: I'm doing another refactor that will move inputs for tasks that expect spans as input from probing to tasks, but that's worthy of a separate PR

probing/gap_preprocess.py
pruksmhc: Again this will be in a separate PR coming up in the next few days - I just wanted to get GAP benchlines started asap with this PR

probing/gap_preprocess.py
sleepinyourhat: Sounds good.

src/tasks/tasks.py
```diff
@@ -153,6 +153,8 @@ def collect_subset_scores(scorer_list, metric_name, tag_list, reset=False):
         subset_scores: a dictionary from subset tags to scores
     '''
     subset_scores = {'%s_%s' % (metric_name, tag_str): scorer.get_metric(reset) for tag_str, scorer in zip(tag_list, scorer_list)}
+    if metric_name == "f1":
+        subset_scores = { key: value[2] for key, value in list(subset_scores.items())}
```
iftenney: Add a comment on what the structure of `value` is?

src/utils/utils.py
```diff
@@ -46,6 +46,21 @@ def wrap_singleton_string(item: Union[Sequence, str]):
...
+    Args:t
+        tag_vocab: the vocab that contains all tags 
+    Returns:
+        tag_list: a list of "coarse__fine" tag strings
+    '''
+    # get dictionary from allennlp vocab, neglecting @@unknown@@ and @@padding@@
+    tid2tag_dict = {key-2: tag \
+        for key, tag in tag_vocab.get_index_to_token_vocabulary().items() \
+            if key - 2 >= 0}
+    tag_list = [tid2tag_dict[tid].replace(':', '_').replace(', ', '_').replace(' ', '_').replace('+', '_') for tid in range(len(tid2tag_dict))]
```
iftenney: Why is this needed? IIRC if the vocab ends in `_labels`, then AllenNLP won't add the unk and padding tokens. See https://github.com/jsalt18-sentence-repl/jiant/blob/master/src/preprocess.py#L529

Also, use a regex instead of a chain of `.replace` calls?

iftenney:
Sorry for not chiming in earlier! I had another implementation of GAP coref as an edge probing task, but didn't have a chance to follow up on it. However, it's quite a bit less code - if you use the edge probing format & preprocessing scripts, you don't need any custom task loaders. It's on this branch if you want to cherry-pick: https://github.com/jsalt18-sentence-repl/jiant/blob/ep_rel_data/src/tasks/edge_probing.py#L329

That being said, I agree with Sam that if this is a core / non-probing task, it's probably best to have a codepath separate from `EdgeProbingTask`. I'll take a more thorough look, maybe we can just clone that and share things like the retokenization code that can be more easily abstracted to libraries.

pruksmhc:
Update: I currently have quite a lot of changes in this PR, which I'm working on extracting out into smaller PRs so that this only has changes relevant to GAP. 
Notably: 
-Refactoring around tag_list.
-Moving some of probing related to span modules into core.
-Adding some more modularity around some custom scorers (which would be helpful if we ever do micro, macro-F1). 
(I'll also clean up code in each of these smaller PRs).

src/utils/utils.py
pruksmhc: This is for backwards-compatibility with the original implementation of subset scorers. Shouldn't be too bad to just change the tasks currently using this though.

config/benchmarkssuperglue/gap_bert.conf
pruksmhc: Sorry, that was bad naming. Changed the names to be more informative.

config/benchmarkssuperglue/gap_bert.conf
pruksmhc: This shouldn't make a difference with BERT or OpenAI.GPT, just elmo.

sleepinyourhat:
Is this on track for the next week or so?

sleepinyourhat:
Is this on track for the next week or so?

pruksmhc:
Working on cleaning it up!

sleepinyourhat:
This is the PR I was mentioning the other day, BTW. Saw some activity.

pruksmhc:
Closing this since it's out of date and not a priority to merge as of now. Reopen once it is more ready for PR.

PR #528, opened 2019-03-18
Pep8 CleanUp of src

pruksmhc:
Self-explanatory


src/evaluate.py
```diff
@@ -47,7 +47,8 @@ def evaluate(model, tasks: Sequence[tasks_module.Task], batch_size: int,
     all_preds = {}
     n_examples_overall = 0
 
-    assert len(tasks) > 0, "Configured to evaluate, but specified no task to evaluate."
+    assert len(
+        tasks) > 0, "Configured to evaluate, but specified no task to evaluate."
```
sleepinyourhat: These break points are kinda ugly, but it's too late now. We can tidy them up as we see them later on.

PR #530, opened 2019-03-20
Refactoring scorer + update_metrics

pruksmhc:
Refactoring scorers to make it more flexible. 
Changes: 
-> Added update_metrics to Task class to allow for flexibility in scorers (moving that logic out of src/models.py to individual tasks) 
-> Added self.subset_scorers to tasks to fit API 

Currently only affects single single sentence tasks and pair sentence classification tasks (non-diagnostic)
Current status:
-Tested on WNLI, STSB, COLA
-Testing on the rest of the affected tasks

### DIFF
--- a/src/tasks/tasks.py
+++ b/src/tasks/tasks.py
@@ -458,7 +458,7 @@ class CoLAAnalysisTask(SingleClassificationTask):
         self.val_metric_decreases = False
         self.scorer1 = Correlation("matthews")
         self.scorer2 = CategoricalAccuracy()
-        self.scorers =[self.scorer1, self.scorer2]
+        self.scorers = [self.scorer1, self.scorer2]
 
     def load_data(self, path, max_seq_len):
         '''Load the data'''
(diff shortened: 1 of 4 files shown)


src/tasks/tasks.py
```diff
@@ -426,6 +435,12 @@ def get_metrics(self, reset=False):
         return {'mcc': self.scorer1.get_metric(reset),
                 'accuracy': self.scorer2.get_metric(reset)}
 
+    def update_metrics(self, logits, labels, tagmask=None):
+        logits, labels = logits.detach(), labels.detach()
```
sleepinyourhat: Why the detach? (Don't think it's wrong, just not sure.)

src/tasks/tasks.py
```diff
@@ -437,6 +452,7 @@ def __init__(self, path, max_seq_len, name, **kw):
         self.val_metric_decreases = False
         self.scorer1 = Correlation("matthews")
         self.scorer2 = CategoricalAccuracy()
+        self.scorers =[self.scorer1, self.scorer2]
```
sleepinyourhat: autopep8

iftenney:
Is this intended to only work for single-sentence tasks? It seems fairly safe and general enough to do everything.

src/tasks/tasks.py
HaokunLiu: Pytorch sometimes reports errors when converting a gpu tensor to cpu, if it is needed to compute some other tensor that requires grad. Here, logits and labels are used to compute loss, while they will be moved to cpu within the scorer updating function.

sleepinyourhat:
Thanks!

PR #531, opened 2019-03-21
Extracting out tokenization alignment functions

pruksmhc:
Changes:
Extracting out tokenization alignment functions from probing/ to src/utils/retokenize.py. This will allow non-probing tasks to also use these tokenization alignment methods.

### DIFF
--- a/src/utils/retokenize.py
+++ b/src/utils/retokenize.py
@@ -240,7 +240,7 @@ def align_moses(text: Text) -> Tuple[TokenAligner, List[Text]]:
 
 def align_openai(text: Text) -> Tuple[TokenAligner, List[Text]]:
     eow_tokens = space_tokenize_with_eow(text)
-    openai_utils = get_tokenizer("OpenAI.bpe_tokens")
+    openai_utils = get_tokenizer("OpenAI.BPE")
     bpe_tokens = openai_utils.tokenize(text)
     ta = TokenAligner(eow_tokens, bpe_tokens)
     return ta, bpe_tokens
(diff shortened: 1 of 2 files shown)


src/utils/retokenize.py
```diff
@@ -207,3 +210,61 @@ def project_span(self, start, end) -> Tuple[int, int]:
...
+    bpe_tokens = openai_utils.tokenize(text)
+    ta = TokenAligner(eow_tokens, bpe_tokens)
+    return ta, bpe_tokens
+
+def align_bert(text: Text, model_name: str) -> Tuple[TokenAligner, List[Text]]:
+    # If using lowercase, do this for the source tokens for better matching.
+    do_lower_case = model_name.endswith('uncased')
+    bow_tokens = space_tokenize_with_bow(text.lower() if do_lower_case else
+                                         text)
+    bert_tokenizer = get_tokenizer(bow_tokens)
```
iftenney: shouldn't this be `get_tokenizer(model_name)` ?

iftenney:
Not sure if this is ready to review yet, but can you also refactor `retokenize_edge_data.py` to import these functions so they aren't defined twice?

Also, can you test with some sample data for each tokenizer (SPR2 should be small and run quick) and compare to the old version? Thanks!

src/utils/retokenize.py
```diff
@@ -207,3 +210,61 @@ def project_span(self, start, end) -> Tuple[int, int]:
...
+def align_moses(text: Text) -> Tuple[TokenAligner, List[Text]]:
+    MosesTokenizer = get_tokenizer("MosesTokenizer")
+    moses_tokens = MosesTokenizer.tokenize(text)
+    cleaned_moses_tokens = utils.unescape_moses(moses_tokens)
+    ta = TokenAligner(text, cleaned_moses_tokens)
+    return ta, moses_tokens
+
+def align_openai(text: Text) -> Tuple[TokenAligner, List[Text]]:
+    eow_tokens = space_tokenize_with_eow(text)
+    openai_utils = get_tokenizer("OpenAI.bpe_tokens")
```
iftenney: `get_tokenizer("OpenAI.BPE")` ?

pruksmhc:
Tested on spr2!

PR #535, opened 2019-04-02
Cleaning up sep_embs_for_skip embs

pruksmhc:
Right now, sep_embs_for_skip have a slightly weird code flow where for GPT and BERT, if we don't have a sentence encoder, we use the task-specific skip embedding as the input to the classifier, which doesn't completely fit in with the idea of "skip_emb" in models that use a shared encoder (where the skip_emb is the + x in f(x) + x where f is the shared encoder. 
This has led to some confusion over what sep_embs_for_skip exactly does for current users. 
Suggested Changes:
-For GPT, BERT, we should block the use of skip_embs for these two since it doesn’t make sense (the transformer layers already have a lot of skip embeddings inside them. 
-We should only support task_specific skip embeddings for ELMO explicitly. 
The two changes will make the code flow simpler. 

What we should support:
Users should be able to:
-Use word embeddings without a shared encoder and feed directly into classifier.
For ELMO specifically: 
-Use a shared RNN encoder without a skip embedding.
-Use a shared encoder with a task-specific skip embedding. 
-Use a shared encoder with a non-task-specific skip embedding

Suggestions welcome!


W4ngatang:
For the first bullet, did you mean "For *GPT*, BERT..."? ELMo is not a Transformer.

I think there are still cases where you would want to explore using some kind of sequence model on top of GPT/BERT and include skip embeddings to let task specific classifiers see the original contextual representations. Maybe we can instead (1) enforce that if you use a "null" sentence encoder, `sep_embs_for_skip = 1` (2) rename `sep_embs_for_skip` , it's not very intuitive from the name.

pruksmhc:
Sorry I meant GPT not ELMO.

README.md
```diff
@@ -149,6 +149,10 @@ The base model class is a MultiTaskModel. To add another model, first add the cl
 Task-specific components include logistic regression and multi-layer perceptron for classification and regression tasks, and an RNN decoder with attention for sequence transduction tasks.
 To see the full set of available params, see [config/defaults.conf](config/defaults.conf). For a list of options affecting the execution pipeline (which configuration file to use, whether to enable remote logging or tensorboard, etc.), see the arguments section in [main.py](main.py).
 
+### Supported Functionality
+
+We allow for training on a shared encoder from the word contextual embeddings, as well as forwarding from the embeddings to the task specific module (by setting sent_enc = null).  Additionally, if you would like to add a skip connection from the contextual embedding directly to the module, set skip_embs = 1.
```
sleepinyourhat: High-level comment: It's fairly confusing that two of our basic use-cases are so different. To change from BERT+classifier to randomly-initialized-BiLSTM+classifier, you need to change quite a lot, since they don't play the same role in our pipeline. Do whatever you can to make the setup clear, including renaming variables where that can help, and adding explicit reminders in several places that ELMo/BERT/GPT are all treated as part of the input handler and aren't shared sentence encoders in our sense.

sleepinyourhat: Up to you how much of this belongs in this PR.

config/defaults.conf
```diff
@@ -234,18 +234,17 @@ bert_fine_tune = 0  // If false, BERT transformer weights will
...
 
 // Sentence Encoder //
-sent_enc = rnn  // Type of sentence encoder: 'bow', 'rnn' (for LSTM), or 'transformer'
-                // set to 'null' to do nothing but forward the skip connection.
+sent_enc = rnn  // Type of sentence encoder: 'bow', 'rnn' (for LSTM),, or 'transformer'
+                // set to 'null' to feed the output of the pretrained encoder into the task specific module.
                 // Note: Transformer has no known issues, but hasn't performed well, so there may be bugs.
-                // Note: 'bow' just skips the encoder step and passes the word representations to the task model, s
+                // Note: 'bow' just skips the encoder step and passes the word representations to the task model,
                 //   so it is possible to combine attention or max pooling with the 'bow' encoder.
```
sleepinyourhat: Make it clear here that a typical BERT/GPT use case should involve 'none', not 'transformer'.

config/defaults.conf
```diff
@@ -234,18 +234,17 @@ bert_fine_tune = 0  // If false, BERT transformer weights will
...
 bidirectional = 1  // If true, the RNN encoder should be bidirectional.
 d_hid = 1024  // Hidden dimension size (usually num_heads * d_proj for transformer)
 n_layers_enc = 2  // Number of encoder layers. Usually 8–12 for Transformer.
 skip_embs = 1  // If true, concatenate the encoder's input (ELMo or embeddings) with the encoder's output.
-sep_embs_for_skip = 0  // Whether the skip embedding uses the same embedder object as the original embedding (before skip).
-                       // Only makes a difference if we are using ELMo weights, where it allows the four tuned ELMo scalars
-                       // to vary separately for each target task.
+// Not supported for BERT and OpenAI GPT.
+sep_elmo_embs_for_skip = 0  // Whether the skip embedding uses the same embedder object as the original embedding (before skip), used only for ELMO
```
sleepinyourhat: Leave in the comment about scalars—I think that's helpful context.

src/models.py
```diff
@@ -116,19 +118,19 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
...
                                        skip_embs=args.skip_embs,
                                        cove_layer=cove_layer,
-                                       sep_embs_for_skip=args.sep_embs_for_skip)
+                                       sep_elmo_embs_for_skip=args.sep_elmo_embs_for_skip,
+                                       sent_enc_type=args.sent_enc)
         log.info("Using Transformer architecture for shared encoder!")
     elif args.sent_enc == 'null':
         # Expose word representation layer (GloVe, ELMo, etc.) directly.
-        assert_for_log(args.skip_embs, f"skip_embs must be set for "
-                       "'{args.sent_enc}' encoder")
```
sleepinyourhat: This seems likely to cause confusion. The most intuitive null sentence encoder would just be a passthrough `f(x)=x`, for which a skip connection is useless. It seems like we're currently doing something much less helpful like `f(x)=null` instead.

(Again, we don't have to fix this here.)

src/models.py
```diff
@@ -116,19 +118,19 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
...
         sent_encoder = SentenceEncoder(vocab, embedder, args.n_layers_highway,
                                        phrase_layer, skip_embs=args.skip_embs,
                                        dropout=args.dropout,
-                                       sep_embs_for_skip=args.sep_embs_for_skip,
-                                       cove_layer=cove_layer)
-        d_sent = 0  # skip connection added below
+                                       sep_elmo_embs_for_skip=args.sep_elmo_embs_for_skip,
+                                       cove_layer=cove_layer, 
+                                       sent_enc_type=args.sent_enc)
+        d_sent = args.d_hid # passing output of pretrained encoder into task classifier
```
sleepinyourhat: Why the change here?

src/models.py
```diff
@@ -226,6 +228,40 @@ def get_task_whitelist(args):
              str(train_task_names), str(eval_clf_names))
     return train_task_names, eval_clf_names
 
+def prepare_elmo_task_specific_skip_embeddings(args, tasks):
```
sleepinyourhat: I assume this is mostly repurposed old code?

src/models.py
```diff
@@ -421,6 +427,12 @@ def build_task_specific_modules(
     ''' Build task-specific components for a task and add them to model.
         These include decoders, linear layers for linear models.
      '''
+    assert not (args.skip_embs == 1 and args.sep_elmo_embs_for_skip == 0 and args.sent_enc == "null"), \
+        "Currently, skip_embs = 1, sep_elmo_embs_for_skip=1, and sent_enc = null, which means " \
+        "you're feeding in the contextual embeddings twice into your model. Please change settings"
```
sleepinyourhat: This seems to send a different message from the assert I commented on above. What does sent_enc = null do?

src/models.py
pruksmhc: yep!

src/models.py
pruksmhc: Adding sent_enc_type is just for code style/cleaner "if" condition purposes, d_sent must be changed from 0 because we're now passing the output of BERT/elmo/GPT through

src/models.py
pruksmhc: sent_enc = null means f(x) = x. What this assert basically says is that it doesn't make sense to pass in skip_embs and also have sent_enc = null, since that means we're passing into the classifier the output of the contextual encoder twice (once from the sent_enc = null, and once from the skip embedding).

pruksmhc:
Bump @W4ngatang  :)

config/defaults.conf
W4ngatang: I think this setting name is kind of unclear. Could we switch to something like `elmo_skip_connection_per_task` or something that conveys that this will have a different set of weights for each task

src/modules/modules.py
```diff
@@ -102,6 +103,40 @@ def __init__(self, vocab, text_field_embedder, num_highway_layers, phrase_layer,
 
         initializer(self)
 
+    def get_cove_layer_embeddings(sent, sent_embs, task_sent_embs):
```
W4ngatang: nice, but minor nit: add `_` in front of the method name, i.e. `_get_cove_layer_embeddings`

src/modules/modules.py
```diff
@@ -126,87 +161,45 @@ def forward(self, sent, task, reset=True):
...
-        if task_sent_embs is not None:
-            task_sent_embs = self._dropout(task_sent_embs)
+            task_sent_embs, sent_embs = get_cove_layer_embeddings(sent, sent_embs, task_sent_embs)
 
         # The rest of the model
         sent_mask = util.get_text_field_mask(sent).float()
         sent_lstm_mask = sent_mask if self._mask_lstms else None
-        if sent_embs is not None:
-            sent_enc = self._phrase_layer(sent_embs, sent_lstm_mask)
+        if self.sent_enc_type != 'null':
```
W4ngatang: this seems like a lot more work (of passing around `sent_enc_type` for a relatively small benefit. Why is this change necessary?

W4ngatang: e.g. we could still use the check ` isinstance(self._phrase_layer, NullPhraseLayer)`, I think?

src/models.py
W4ngatang: I think deleting this and the changes to `SentEncoder` address what Sam is saying (as in, these changes remove the confusion).

src/modules/modules.py
pruksmhc: isinstance(self._phrase_layer, NullPhraseLayer is not clear as to how the NullPhraaseLayer is triggered. At essence, it's triggered when sent_enc = null, so I passed it in for code clarity

pruksmhc:
High level comment:
As per our discussion (First few comments of this PR), this PR clears up the role of sep_embs_for_skip, specifically allowing:
-Use word embeddings without a shared encoder and feed directly into classifier.
For ELMO specifically:
-Use a shared encoder with a task-specific skip embedding.
-Use a shared encoder with a non-task-specific skip embedding

For others (specifically for BERT and GPT):
-To use a shared encoder with a skip embedding 
-To feed word-in-context embedding directly into classifier

The changes I made were around making sure that sent_embs is never None (if sent_enc = none), then sent_embs is just the output of the BertEmbedder, thus isolating task_sent_embs to truly only be used during skip connections .

config/defaults.conf
W4ngatang: Bump.

config/defaults.conf
```diff
@@ -233,18 +233,17 @@ onlstm_tying = 0  // Language Modeling tying of weights.
...
+                // Note: 'bow' just skips the encoder step and passes the word representations to the task model,
                 //   so it is possible to combine attention or max pooling with the 'bow' encoder.
 bidirectional = 1  // If true, the RNN encoder should be bidirectional.
 d_hid = 1024  // Hidden dimension size (usually num_heads * d_proj for transformer)
 n_layers_enc = 2  // Number of encoder layers. Usually 8–12 for Transformer.
 skip_embs = 1  // If true, concatenate the encoder's input (ELMo or embeddings) with the encoder's output.
-sep_embs_for_skip = 0  // Whether the skip embedding uses the same embedder object as the original embedding (before skip).
-                       // Only makes a difference if we are using ELMo weights, where it allows the four tuned ELMo scalars
-                       // to vary separately for each target task.
+// Not supported for BERT and OpenAI GPT.
```
W4ngatang: Nit: add indentation

config/defaults.conf
```diff
@@ -233,18 +233,17 @@ onlstm_tying = 0  // Language Modeling tying of weights.
...
+elmo_skip_connection_per_task = 0  // Whether the skip embedding uses the same embedder object as the original embedding (before skip), used only for ELMO
```
W4ngatang: A little more descriptive: "Whether to use the pretraining layer-mixing weights in the skip connection for target tasks or target-task specific layer mixing weights.

W4ngatang:
This looks good, but I still feel iffy about adding this new `sent_enc_type` parameter. It still seems like a lot of work for just re-writing one check (AFAICT). I don't think checking that the class of the `phrase_layer` is a `NullEncoder` is that hard for users to understand, so I'd prefer reverting. This isn't a hill I want to die on, so if you feel really strongly (though I'd still want to hear more reasoning), seems okay to me.

W4ngatang:
Also, testing-wise, please be sure to test everything that this change might touch, including (i) CoVe (ii) ELMo with and without task-specific skip embeddings (iii) BERT with and without skip embs. I think it should be pretty quick; just changing the configs?

pruksmhc:
Closing this since it's out of date and not a priority to merge as of now. Reopen once it is more ready for PR.

PR #536, opened 2019-04-02
Allow rel_path override

zphang:
Expose an override for rel_path for task data.

### DIFF
--- a/config/spring19_seminar/cola_modified.conf
+++ b/config/spring19_seminar/cola_modified.conf
@@ -1,0 +1,3 @@
+include "../defaults.conf"
+
+cola_rel_path = CoLA2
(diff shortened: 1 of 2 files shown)


zphang:
This is a stopgap measure for allowing users to run experiments on different datasets that can be considered to belong to the same task (e.g. a modified CoLA data set) without modifying the path to the overall JIANT data. 

Importantly, because only the data path is changed and not the task name, this override will not allow for multiple modified versions of the same task to exist in the same experiment, because the cached tokenized data is tied to the task name.

Ultimately, supporting such "subtasks" would require further consideration of how the task name is used in different parts of the pipeline, e.g. the task name is used to identify both the cache tokenized data (that we would not want to share), as well as the output heads of the model (that we probably do want to share).

config/spring19_seminar/cola_modified.conf
```diff
@@ -0,0 +1,3 @@
+include "../defaults.conf"
+
+cola_rel_path = CoLA2
```
sleepinyourhat: Add an example of this to defaults.conf, and add a comment there explaining what this does and why/how you should be careful with it.

Issue #505, opened 2019-03-02
Clean Up Finetuning Flags 


pruksmhc:
The new changes in BERT will introduce finetune paradigm vs frozen for BERT style training vs Elmo, which is redundant with bert_fine_tune and openai_fine_tune. 
We should clean this up (use flags etc).

PR #540, opened 2019-04-06
Finetune flag

HaokunLiu:
Removed redundant bert_fine_tune and openai_transformer_fine_tune flags. transfer_paradigm now controls whether those parameters in bert and gpt are trainable.
#505


pruksmhc:
Yep this looks good, thanks for tackling this Haokun! Do some runs to test that this works.

HaokunLiu:
> This looks reasonable to me, but I'd prefer that @pruksmhc or @W4ngatang be the main reviewers on this. It might help to add a comment to defaults clarifying what _counts_ as a pretrained model—this is for GPT/BERT/ELMo, not for smaller models that we train here.

Does it look better now?

HaokunLiu:
> Yep this looks good, thanks for tackling this Haokun! Do some runs to test that this works.

Tested, looks reasonable.

config/defaults.conf
```diff
@@ -70,8 +70,11 @@ do_full_eval = 1     // Evaluate the model on the tasks on target_tasks.
 
 // Related configuration
 load_model = 1  // If true, restore from checkpoint when starting do_pretrain. No impact on do_target_task_training.
-transfer_paradigm = "frozen" // How to do learning on top of the pretrained model
+transfer_paradigm = "frozen" // How to do learning on top of the task-independently pretrained model
+                             // Note: this option only works for models like ELMO/GPT/BERT (models that was pretrained on large amount of data)
```
W4ngatang: why wouldn't these apply if we're using LSTMs or some non pre-trained models?

README.md
```diff
@@ -181,7 +181,6 @@ Note: if you want to train and evaluate on a task, that task must be in both ``p
 We support two modes of adapting pretrained models to target tasks. 
 Setting `transfer_paradigm = finetune` will fine-tune the entire model while training for a target task.
 The mode will create a copy of the model _per target task_.
-If using a pretrained model such as BERT or GPT, be sure to also set the corresponding fine-tune flag, e.g. `bert_fine_tune = 1`.
```
W4ngatang: If we're deleting these comments, can we add a new comment somewhere that explains how to do the transfer?

config/defaults.conf
HaokunLiu: you are right, I don't what I was thinking.

README.md
HaokunLiu: I think the comments of transfer_paradigm roughly explains the thing?  Is there anything more we need to add there?

sleepinyourhat:
Now there are merge conflicts—dang. Mind fixing those, then we can merge?

PR #542, opened 2019-04-09
Subtasks [WIP]

zphang:
Introducing subtasks, to allow quick creation/use of task data that follows an existing task format concurrently with existing tasks, without the need to modify code. 

**Open to feedback on how to best integrate this into the current jiant infrastructure.**

Notes:

- To make them clearly separate from regular tasks, all subtasks are defined in a "subtasks" top-level config variable. 
- Each subtask is associated with a "core_task"
- Subtasks inherit all properties from their core task, unless explicitly overridden. Of note, either rel_path, or an "explicit_path" needs to be provided.
- **IMPORTANT**: Subtasks use different output heads by default. Subtasks can be set to use the core_task's (or some other subtask's) output head using the "use_classifier" argument. An example is in the subtasks_run.conf


sleepinyourhat:
Pull the '[WIP]' tag when this is ready for review. If you can get it done before Friday's refactor-fest, you'll avoid a potentially difficult merge.

sleepinyourhat:
@zphang Where does this stand?

zphang:
I think this requires more rigorous testing - if this gets used but has issues, the proble could be relatively subtle and hard to detect. Does this still feel like something people would use?

sleepinyourhat:
Hrm. I guess now that use_classifier seems somewhat stable, it's not crucial. Up to you.

sleepinyourhat:
Safe to close this?

PR #544, opened 2019-04-12
Minimal pair task, maskLM acceptability, relocate NPI task

HaokunLiu:
Introducing minimal pair task, its input is a pair of sentences, but it does not inherit from pair classification task, otherwise bert model will use "[CLS] sent1 [SEP] sent2". Two task names are registered, to differentiate between frozen and tuned setting. 

"cola-pair-tuned" compute the acceptability logits on two sentences independently with the classifier 
 trained on cola and choose the more likely one. "cola-pair-frozen" predict the word at the location where two sentences differs with pretrained MaskLM head.

Borrowing classifier from another task was initially introduced only for mnli-diagnostic, therefore the control logic in main.py are written specifically for mnli-diagnostic. This update generalize that mechanism and allow other tasks to borrow classifier from a intrinsically similar task.
To use this feature, set TASK1 {use_classifier = "TASK0"} in default.conf, so that TASK1 will reuse the classifier of TASK0, instead of create its own.


config/spring19_seminar/colapair_bertF.conf
```diff
@@ -0,0 +1,60 @@
+// Direct probe model on cola-pair dataset, the parameters are frozen, we only use MaskLM to do acceptability judgement 
+// F stands for frozen
```
sleepinyourhat: Might as well spell this out in the filename.

config/spring19_seminar/colapair_bertF.conf
```diff
@@ -0,0 +1,60 @@
...
+
+pretrain_tasks = ""  // we can choose pretrain the model on cola task 
+do_pretrain = 0  // but not in frozen setting
+
+target_tasks = cola-pair-frozen
+do_target_task_training = 0  // do not tune the parameters on target task
+do_full_eval = 1  // evaluation the model after training
+
+transfer_paradigm = "frozen"
+batch_size = 16
```
sleepinyourhat: FWIW, try to coordinate with others on this. We use 24 elsewhere, and this could cause some issues with comparison.

config/spring19_seminar/colapair_bertF.conf
```diff
@@ -0,0 +1,60 @@
...
+
+allow_untrained_encoder_parameters = 1  // set this to 1 when running random-elmo
+allow_reuse_of_pretraining_parameters = 0  // set this to 1 because we pretrain and finetune model on the same task
+
+// Model settings
+
+elmo = 0
+sep_embs_for_skip = 1
+tokenizer = "bert-base-uncased"
+bert_model_name = "bert-base-uncased"
```
sleepinyourhat: We use cased in some other experiments. Coordinate with others.

config/spring19_seminar/colapair_bertF.conf
```diff
@@ -0,0 +1,60 @@
...
+bert_embeddings_mode = "top"   // how to use the outputs of the BERT module
+                               // set as "top", we use only the top-layer activation
+                               // other options: "only" uses the lexical layer (first layer)
+                               //                "cat" uses lexical layer + top layer
+bert_fine_tune = 0
+pair_attn = 0 // shouldn''t be needed but JIC
+dropout = 0.1 // following BERT paper
```
sleepinyourhat: If you inherit from one of the BERT config files, will you get this for free? Best to try to add only the changes we really need.

main.py
```diff
@@ -313,11 +313,12 @@ def main(cl_arguments):
...
 
         for task in target_tasks:
-            # Skip mnli-diagnostic
+            # Skip tuning on a target task, if it does not really have a training set (like mnli-diagnostic)
             # This has to be handled differently than probing tasks because probing tasks require the "is_probing_task"
             # to be set to True. For mnli-diagnostic this flag will be False because it is part of GLUE and
             # "is_probing_task is global flag specific to a run, not to a task.
-            if task.name == 'mnli-diagnostic':
+            if task.get_split_text('train') == task.get_split_text('val') or \
+                task.get_split_text('train') == task.get_split_text('test'):
```
sleepinyourhat: If this is at all slow, we should just set an attribute of tasks that we can check for.

main.py
```diff
@@ -368,20 +369,21 @@ def main(cl_arguments):
...
+                finetune_path = os.path.join(args.run_dir, "model_state_%s_best.th" % name_to_load)
                 if os.path.exists(finetune_path):
                     ckpt_path = finetune_path
                 else:
                     ckpt_path = get_best_checkpoint_path(args.run_dir)
                 load_model_state(model, ckpt_path, args.cuda, skip_task_models=[], strict=strict)
 
-                tasks = [task]
-                if task.name == 'mnli':
-                    tasks += [t for t in target_tasks if t.name == 'mnli-diagnostic']
```
sleepinyourhat: Make sure to double check that mnli-diagnostic still works the same way.

scripts/cola_baseline.sh
```diff
@@ -12,3 +12,9 @@ python main.py --config_file "config/spring19_seminar/cola_elmo.conf" --override
```
sleepinyourhat: Mind moving this to a scripts/spring19-seminar directory?

config/spring19_seminar/colapair_bertF.conf
HaokunLiu: Yes. Some part of data(not NPI) is not cased, so I used uncased initially. Now that we have pivoted to focus on NPI, we can safely change to cased.

config/spring19_seminar/colapair_bertF.conf
HaokunLiu: I got batch_size = 16 from final_bert.conf , I couldn't find any bert config that have batch_size = 24.

config/spring19_seminar/colapair_bertF.conf
HaokunLiu: Yes, it should be better to import bert-final or make a stem config for spring19_seminar to import from.

scripts/cola_baseline.sh
HaokunLiu: Will do.

config/spring19_seminar/colapair_bertF.conf
sleepinyourhat: It looks like that might have just been @PterosDiacos. Fine to stick with 16.

config/spring19_seminar/colapair_bertF.conf
sleepinyourhat: Sounds good!

main.py
HaokunLiu: done

config/spring19_seminar/colapair_bertF.conf
HaokunLiu: ok

main.py
HaokunLiu: I can do it on Thursday or Friday.
Some part of the code in mnli-diagnostic is unnecessarily unique and long, since we need to test it either way, I wonder can we take this chance to refactor mnli-diagnostic?

config/spring19_seminar/cola_gpt.conf
```diff
@@ -8,6 +8,7 @@ include "cola_elmo.conf"
 do_target_task_training = 0  // only do main training phase, since our pretrain and target task are the same
 allow_reuse_of_pretraining_parameters = 1  // set this to 1 because we pretrain and finetune model on the same task
 batch_size = 2  // keep it small to avoid out-of-memory
+lr = 1e-5  // we need to keep lr small for fine-tuning gpt, especially so because our BS is smaller than that in gpt paper
```
sleepinyourhat: Fine to check this in, but we probably won't be publishing anything with GPT, right?

main.py
sleepinyourhat: Sure, but you'd have to be careful to test it thoroughly.

src/bert/utils.py
```diff
@@ -152,3 +152,13 @@ def get_output_dim(self):
             return 2 * self.model.config.hidden_size
         else:
             return self.model.config.hidden_size
+
+    def hatch_LM_head(self, args):
```
W4ngatang: What does `hatch` mean? Could you make the docstring a little clearer as to what this is doing?

src/models.py
```diff
@@ -748,6 +753,47 @@ def _single_sentence_forward(self, batch, task, predict):
...
+
+        if task.name == 'cola-pair-frozen':
+            bs = get_batch_size(batch)
+            # embed the sentence
+            sent_embs, _ = self.sent_encoder(batch['input0'], task)
+            # pass to a task specific classifier
+            classifier = self._get_classifier(task)
+            index = batch['index'].squeeze(dim=1)
+            logits_full = classifier(sent_embs)[range(bs), index]
+            sent1_key = batch['input1']['bert_wpm_pretokenized'][range(bs), index] - 2
```
W4ngatang: this will break whenever you aren't using BERT tokenization?

src/tasks/tasks.py
```diff
@@ -223,6 +223,11 @@ def n_train_examples(self):
     def n_val_examples(self):
         return self.example_counts['val']
 
+    @property
+    def no_train_data(self):
+        return self.get_split_text('train') == self.get_split_text('val') or \
```
W4ngatang: isn't this kind of expensive for what could otherwise be a Boolean we pass? I guess it does get kind of tricky if you want to have versions of the same task class that do and do not have training data, but I'm not sure how often that's happening

src/tasks/tasks.py
```diff
@@ -543,6 +546,101 @@ def get_metrics(self, reset=False):
...
+        collected_metrics = {
+            'mcc': self.scorer1.get_metric(reset),
+            'accuracy': self.scorer2.get_metric(reset)}
+        collected_metrics.update(
+            collect_subset_scores(
+                self.tag_scorers1,
+                'mcc',
+                self.tag_list,
+                reset))
+        # collected_metrics.update(
```
W4ngatang: might as well delete

src/tasks/tasks.py
```diff
@@ -543,6 +546,101 @@ def get_metrics(self, reset=False):
...
+            d["tagmask"] = MultiLabelField(tagids, label_namespace="tagids",
+                                           skip_indexing=True, num_labels=len(self.tag_list))
+                                           
+            return Instance(d)
+
+        instances = map(_make_instance, *split)
+        return instances  # lazy iterator
+    
+    def update_metrics(self, logits, labels, tagmask=None):
+        logits, labels = logits.detach(), labels.detach()
```
W4ngatang: do you need to detach explicitly? I thought within scorers we're calling `.item()` or something similar anyways.

src/tasks/tasks.py
```diff
@@ -543,6 +546,101 @@ def get_metrics(self, reset=False):
                 reset))
         return collected_metrics
 
+@register_task('cola-pair-frozen', rel_path='CoLA')
```
W4ngatang: Sorry to be annoying about this, but could you create a new file in `src/tasks/`? I'd like to avoid throwing more tasks, esp. tasks that are somewhat project specific, into this file.

src/models.py
```diff
@@ -422,7 +422,7 @@ def build_task_specific_modules(
         These include decoders, linear layers for linear models.
      '''
     task_params = model._get_task_params(task.name)
-    if isinstance(task, SingleClassificationTask):
+    if isinstance(task, SingleClassificationTask) or (isinstance(task, CoLAMinimalPairTask) and task.name == 'cola-pair-tuned'):
```
W4ngatang: Could you restructure to do the second check within a branch for `isinstance(task, CoLAMinimalPairTask)`? Would like to avoid the check for this branch including very task-specific things (although I know we're guilty of this elsewhere).

src/models.py
HaokunLiu: The program would not enter this part of the code, if it isn't using bert. (there won't be suitable task specific module for it, so it will get ValueError("Module not found for %s" % task.name)). This forward function is only for probing BERT with minimal pair, the name of this function did not reflect that, I should probably change it.

config/spring19_seminar/cola_gpt.conf
HaokunLiu: No, we won't

src/tasks/tasks.py
HaokunLiu: I think that is not a big issue. Time-wise, this command is run only once, won't make any difference. Avoiding setting up another Boolean make it less likely for people to make mistakes when adding new tasks. As for the case of same task have and don't have training data, if the training split is set as convention (like mnli-diagnostic), this function will give us the correct Boolean.

src/tasks/tasks.py
HaokunLiu: It's okay, I agree with you.

src/models.py
HaokunLiu: Sure. (that is so true ; )

src/bert/utils.py
HaokunLiu: Well, this name is not very intuitive. I have changed it and added notes

src/tasks/tasks.py
```diff
@@ -800,6 +803,90 @@ def load_data(self, path, max_seq_len):
         log.info("\tFinished loading MNLI data.")
 
 
+@regiscter_task('mnli-diag', rel_path='MNLI/')
```
W4ngatang: register

src/tasks/tasks.py
W4ngatang: Hmm okay, fair enough.

src/models.py
W4ngatang: Ah gotcha. Are you all planning on supporting non-BERT versions of this task? Otherwise renaming would be helpful.

src/models.py
```diff
@@ -433,6 +434,16 @@ def build_task_specific_modules(
         d_sent = args.d_hid + (args.skip_embs * d_emb)
         hid2voc = build_lm(task, d_sent, args)
         setattr(model, '%s_hid2voc' % task.name, hid2voc)
+    elif isinstance(task, CoLAMinimalPairTask):
+        if task.name == 'cola-pair-frozen' and args.bert_model_name:
+            mask_lm_head = model.sent_encoder._text_field_embedder.transplant_LM_head(args)
+            setattr(model, '%s_mdl' % task.name, mask_lm_head)
+        elif task.name == 'cola-pair-tuned':
+            module = build_single_sentence_module(task=task, d_inp=d_sent,
+                                            use_bert=model.use_bert, params=task_params)
```
W4ngatang: minor: indentation

main.py
```diff
@@ -313,11 +313,11 @@ def main(cl_arguments):
                            "they should not be updated! Check sep_embs_for_skip flag or make an issue.")
 
         for task in target_tasks:
-            # Skip mnli-diagnostic
+            # Skip tuning on a target task, if it does not really have a training set (like mnli-diagnostic)
             # This has to be handled differently than probing tasks because probing tasks require the "is_probing_task"
             # to be set to True. For mnli-diagnostic this flag will be False because it is part of GLUE and
             # "is_probing_task is global flag specific to a run, not to a task.
-            if task.name == 'mnli-diagnostic':
+            if task.no_train_data:
```
W4ngatang: thanks!

HaokunLiu:
No hurry review the PR, I haven’t done testing the new mnli-diagostic yet.

src/models.py
HaokunLiu: No. Minimal pairs is specially used as a tool to probe the properties of BERT model. I can't see a scenario where this task is necessary to understand something in other models.

src/models.py
HaokunLiu: thanks.

sleepinyourhat:
Quick comment (not jumping into the review process): See if you can finish this before Friday. There'll be a bunch of refactors then, so you're likely to hit merge conflicts if this is still unfinished after Friday morning.

HaokunLiu:
Finally tested. 
Please let me know if there is anything else we need to change.
@sleepinyourhat @W4ngatang

src/tasks/tasks.py
HaokunLiu: done

config/spring19_seminar/colapair_bert_frozen.conf
```diff
@@ -0,0 +1,30 @@
+// Direct probe model on cola-pair dataset, the parameters are frozen, we only use MaskLM to do acceptability judgement 
+
+include "../final-bert.conf"
+
+
+// Do not change
+
+word_embs_file = ""  // we do not need this in elmo, gpt or bert
+fastText_model_file = ""  // we do not need this in elmo, gpt or bert
```
sleepinyourhat: This is deprecated. Delete?

config/spring19_seminar/colapair_bert_frozen.conf
```diff
@@ -0,0 +1,30 @@
...
+target_tasks = cola-pair-frozen
+do_target_task_training = 0  // do not tune the parameters on target task
+
+transfer_paradigm = "frozen"
+
+allow_untrained_encoder_parameters = 1  // because we test the model directly
+
+// Model settings
+
+bert_fine_tune = 0
```
sleepinyourhat: Doesn't matter, since we never do training at all.

config/spring19_seminar/colapair_bert_frozen.conf
```diff
@@ -0,0 +1,30 @@
+// Direct probe model on cola-pair dataset, the parameters are frozen, we only use MaskLM to do acceptability judgement 
+
+include "../final-bert.conf"
+
+
+// Do not change
+
+word_embs_file = ""  // we do not need this in elmo, gpt or bert
```
sleepinyourhat: Delete. Even if this is specified, we won't load embeddings here.

src/evaluate.py
```diff
@@ -111,7 +111,7 @@ def evaluate(model, tasks: Sequence[tasks_module.Task], batch_size: int,
         all_preds[task.name] = task_preds
         log.info("Finished evaluating on: %s", task.name)
 
-    all_metrics["micro_avg"] /= n_examples_overall
+    all_metrics["micro_avg"] /= max(n_examples_overall, 1)
```
W4ngatang: is this to avoid issues with the diagnostics? If so, leave a comment saying so

src/tasks/tasks.py
```diff
@@ -457,6 +462,10 @@ def update_metrics(self, logits, labels, tagmask=None):
 
 @register_task('cola-analysis', rel_path='CoLA/')
 class CoLAAnalysisTask(SingleClassificationTask):
```
W4ngatang: actually, could you put this into the `minimal_pairs` task file and rename it to something like `tasks/acceptability.py`? I'm trying to avoid growing `tasks.py` any more.

HaokunLiu: Ok

src/evaluate.py
HaokunLiu: Yes

config/spring19_seminar/colapair_bert_frozen.conf
HaokunLiu: Yes

HaokunLiu:
We don't need a special config file for minimal pair experiments now, we just need a sh script. So those config are deleted.

HaokunLiu:
The script is not complete, I will adjust as needed when running it. 
For now, is it possible to merge this and allow me to upload a final version of the script afterwards?

HaokunLiu:
Some updates:
1. Since mnli diagnostic task have some changes in SuperGLUE, it is safer to keep it as is. So I removed my version of the mnli diagnostic task in the code.
2. I moved NPI task to acceptablity.py

sleepinyourhat:
Thanks! Make sure that tests all pass and let me know when it’s ready for review.

src/utils/data_loaders.py
```diff
@@ -26,12 +26,12 @@ def load_span_data(tokenizer_name, file_name, label_fn=None, has_labels=True):
     [Mr, ., Por, ter, is, nice ]. If the original span indices was [0,2], under the new tokenization, 
     it becomes [0, 3].
     The task file should of be of the following form: 
-        text: str, 
+        text: str
```
HaokunLiu: Strictly speaking, this is not something I should touch, since it's someone else's  function. But the format checker complained, and the following comma is inconsistent with everywhere else, it is not too terrible to make it cleaner, right?

src/bert/utils.py
```diff
@@ -149,3 +149,16 @@ def get_output_dim(self):
             return 2 * self.model.config.hidden_size
         else:
             return self.model.config.hidden_size
+
+    def transplant_LM_head(self, args):
+        """
+        download the masked LM classifier from huggingface bert
```
sleepinyourhat: Clean this up to match other fns here: Use a full sentence with capitalization. Otherwise looks good!

src/tasks/acceptability.py
```diff
@@ -0,0 +1,433 @@
...
+from ..utils.data_loaders import load_tsv, get_tag_list, BERT_MASK_TOK
+from .tasks import Task, sentence_to_text_field, SingleClassificationTask
+from .tasks import create_subset_scorers, update_subset_scorers, collect_subset_scores
+
+from typing import Iterable, Sequence, List, Dict, Any, Type
+
+
+@register_task("npi_pair_frozen", rel_path="NPI")
+class NPIMinimalPairFrozenTask(Task):
+    """ Task class for frozen minimal pair (cloze) acceptability judgement """
```
sleepinyourhat: How about 'NPIMinimalPairMLMTask' - MLM vs. classifier seems more important than frozen vs. tuned.

src/tasks/acceptability.py
```diff
@@ -0,0 +1,433 @@
+"""Define the tasks and code for loading their data.
```
sleepinyourhat: This comment seems wrong. Did you just copy it from somewhere else? It need to explain what this file is for, and when someone should use it.
 
Also, if CoLA isn't here, 'acceptability' isn't a good name for the file. 'acceptability_probing.py'?

src/tasks/acceptability.py
```diff
@@ -0,0 +1,433 @@
...
+        # Create score for each tag from tag-index dict
+        self.tag_list = get_tag_list(tag_vocab)
+        self.tag_scorers1 = create_subset_scorers(
+            count=len(self.tag_list), scorer_type=Correlation, corr_type="matthews"
+        )
+        self.tag_scorers2 = create_subset_scorers(
+            count=len(self.tag_list), scorer_type=CategoricalAccuracy
+        )
+
+        log.info("\tFinished loading CoLA sperate domain.")
```
sleepinyourhat: I don't know what this means. Clarify.

src/tasks/acceptability.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def process_split(self, split, indexers):
+        def _make_instance(input1, input2, labels, tagids):
+            """ from multiple types in one column create multiple fields """
```
sleepinyourhat: Use full sentences with punctuation.

src/tasks/acceptability.py
```diff
@@ -0,0 +1,433 @@
...
+@register_task("hd_cola_npi_negdet", rel_path="NPI/combs/minus_determiner_negation_biclausal")
+@register_task("hd_cola_npi_negsent", rel_path="NPI/combs/minus_sentential_negation_biclausal")
+@register_task("hd_cola_npi_only", rel_path="NPI/combs/minus_only")
+@register_task("hd_cola_npi_ques", rel_path="NPI/combs/minus_questions")
+@register_task("hd_cola_npi_quessmp", rel_path="NPI/combs/minus_simplequestions")
+@register_task("hd_cola_npi_qnt", rel_path="NPI/combs/minus_quantifiers")
+@register_task("hd_cola_npi_sup", rel_path="NPI/combs/minus_superlative")
+class CoLANPITask(SingleClassificationTask):
+    """Class for NPI-related task; same with Warstdadt acceptability task but outputs labels for
+       test-set
```
sleepinyourhat: Typo in Alex's name. Use full sentence.

src/tasks/acceptability.py
```diff
@@ -0,0 +1,433 @@
...
+        self.path = path
+        self.max_seq_len = max_seq_len
+
+        self.train_data_text = None
+        self.val_data_text = None
+        self.test_data_text = None
+
+        self.val_metric = "%s_mcc" % self.name
+        self.val_metric_decreases = False
+        # self.scorer1 = Average()
```
sleepinyourhat: Cut.

src/tasks/tasks.py
```diff
@@ -232,6 +232,12 @@ def n_train_examples(self):
     def n_val_examples(self):
         return self.example_counts["val"]
 
+    @property
+    def no_train_data(self):
+        return self.get_split_text("train") == self.get_split_text("val") or self.get_split_text(
+            "train"
+        ) == self.get_split_text("test")
```
sleepinyourhat: This is too dangerous: It's a common and easy sanity check to do an experiment where they load the same file for training and validation and see if the model can get to 100% on that data. This would break that setup. 

Find another way to do this—maybe set it to False in the general task class, and override it explicitly when you need it.

src/utils/data_loaders.py
sleepinyourhat: Fine by me!

src/tasks/acceptability.py
HaokunLiu: This was initially from process_split in tasks.py, since it's no longer there, I guess it's fine to remove this comment here as well?

src/tasks/acceptability.py
HaokunLiu: Ah, that's embarrassing, I should probably have noticed this when moving this code from tasks.py.

src/tasks/acceptability.py
HaokunLiu: Yes, that's good, I'll change this. This naming is pretty bad, especially so when considering that "frozen" is also used as a keyword for transfer paradigm.

src/tasks/acceptability.py
HaokunLiu: (1) Exactly.
(2) Sure.

src/tasks/tasks.py
HaokunLiu: Yes, I agree.

HaokunLiu:
This PR is now in an awkward situation. I changed the code after last review. But somehow the PR is still in "change-required", so it didn't received any further review for a long time.  This PR is a conglomeration of multiple things, that makes the commit-review-merge cycle on it quite slow. Now, jiant has gone through some major update. Some parts of this PR is still relevant, the other are no longer needed. It is not very convenient to merge things together.

How about, alternatively, I make several smaller independent PRs when working on Alex Warstadt's summer project, absorbing still necessary parts in this PR into those PRs. I think, this is the easiest way to make the features in this PR catch up with main jiant, and avoid any more prolonged PR processes.

sleepinyourhat:
Yes, that sounds good. Thanks!

- Make sure the future PRs take into account the comments from here.
- Only submit a PR in the future if you're willing to finish it within a couple of weeks.

PR #545, opened 2019-04-12
Conf files and new tasks for the NPI-probing experiments

sheng-fu:
npi_probing_finetune.conf: fine-tune ("do_pretraining") only

npi_probing_target_task.conf: load a model with fine-tuned BERT, do target-task training without fine-tuning BERT, and then do eval
(issue#1: does loading "model_state_main_epoch_1.best_macro.th" seem correct?)

Currently only two tasks are added to tasks.py; more to be created later
(issue: write_preds() not supported by the new tasks)


config/npi_probing_finetune.conf
```diff
@@ -0,0 +1,95 @@
+// Config modifications to apply during all final *BERT* runs.
+
+// This imports the defaults, which can be overridden below.
+include "defaults.conf"
```
sleepinyourhat: This'll be much easier to debug if you only put new content here. If this is based on one of the other conf files, include that instead of defaults, and only include lines you change.

config/npi_probing_target_task.conf
```diff
@@ -0,0 +1,110 @@
...
+
+load_model = 1
+reload_tasks = 0
+reload_indexing = 0
+reload_vocab = 0
+
+exp_dir = ${project_dir}""${exp_name}"/"  // required
+run_dir = ${project_dir}""${exp_name}"/"${run_name}  // required
+local_log_path = ${run_dir}"/eval_log.log"  // log file, goes in run directory
+// log name for remote logging; make as unique as possible
```
sleepinyourhat: Seems odd to include this.

src/tasks/tasks.py
```diff
@@ -455,6 +455,87 @@ def update_metrics(self, logits, labels, tagmask=None):
         self.scorer2(logits, labels)
         return
 
+@register_task('cola_probing', rel_path='CoLA_probing_npi/')
+class CoLAPTask(SingleClassificationTask):
```
sleepinyourhat: Use 'Probing', spelled out' in the class name, and add a comment explaining the difference from CoLATask.

src/tasks/tasks.py
```diff
@@ -455,6 +455,87 @@ def update_metrics(self, logits, labels, tagmask=None):
         self.scorer2(logits, labels)
         return
 
+@register_task('npi', rel_path='CoLA_probing_npi/')
```
HaokunLiu: If the code for these two tasks are identical to CoLA, it may not be necessary to make duplicate classes. Just add  "@register_task('npi', rel_path='CoLA_probing_npi/')" with different task name and rel_path should be enough. See mnli-*** as an example.

HaokunLiu: It may also be a good idea to do it with Jason's new subtask

config/spring19_seminar/npi_probing_finetune.conf
```diff
@@ -0,0 +1,95 @@
...
+}
+classifier = log_reg // following BERT paper
+pretrain_tasks = "cola"  // empty: don't run main training phase
+target_tasks = "probing_metadata"
+
+do_pretrain = 1
+do_target_task_training = 0
+do_full_eval = 0
+
+dropout = 0.1 // following BERT paper
```
sleepinyourhat: Make sure these match up with the other PRs.

config/spring19_seminar/npi_probing_finetune.conf
```diff
@@ -0,0 +1,95 @@
...
+optimizer = bert_adam
+batch_size = 16
+max_epochs = 3
+lr = .00001
+min_lr = .0000001
+lr_patience = 4
+patience = 20
+max_vals = 10000
+
+// For all generation tasks.
```
sleepinyourhat: We shouldn't need most of the below, right? Only set settings that you don't want to inherit from another conf file.

src/tasks/tasks.py
```diff
@@ -455,6 +467,12 @@ def update_metrics(self, logits, labels, tagmask=None):
         self.scorer2(logits, labels)
         return
 
+
```
sleepinyourhat: Cut all this space.

config/spring19_seminar/cola_var/bert.conf
```diff
@@ -0,0 +1,40 @@
+// Import the defaults using the relative path
+include "../../final.conf"
```
sleepinyourhat: Heads up for @HaokunLiu @PterosDiacos: Coordinate on whether to include final or defaults. The differences are pretty light, so either one should be okay.

HaokunLiu: This config seems pretty good. Since the experiments I run are all on BERT, can I simply import from this one?

sleepinyourhat: Sounds good to me!

sleepinyourhat:
Let me know if/when this is ready for a final review. It's best if we can get it in before Friday morning's hackathon thing.

sheng-fu:
I think this is ready for final review; The config file and the scripts are based on bert.conf in #561, but the tasks are registered with the following differences:

All tasks, including the NPI main tasks and the probing tasks, are registered with a NPITask, separated from CoLATask. NPITask uses test_full.tsv when writing predictions for the test set, which makes test-set evaluation slightly easier.

I roughly follow the same task-naming scheme in #561. I remove "cola_" from the tasks' names. I also change the path names a little bit so that they match what's in the data_generation repo.

config/spring19_seminar/cola_var/bert_target_tasks.conf
```diff
@@ -0,0 +1,25 @@
...
+
+// Import the defaults using the relative path
+include "bert.conf"
+
+// Load model checkpoints and save log file for target tasks
+exp_name = "" // supply with overrides
+run_name = "" // supply with overrides
+exp_dir = ${project_dir}""${exp_name}"/" 
+run_dir = ${project_dir}""${exp_name}"/"${run_name} 
+local_log_path = ${run_dir}"/eval_log.log"  // use a separate log file for probing tasks
```
sleepinyourhat: Do we really need to include this last block of code?

config/spring19_seminar/cola_var/bert_target_tasks.conf
```diff
@@ -0,0 +1,25 @@
+//  This config loads a model pretrained with bert.conf (currently in the cola_var branch) and does target-task training
```
sleepinyourhat: Use probing somewhere in the name or the comment, since this config seems to be for probing-style experiments with frozen BERT.

scripts/spring_19_seminar/metadata_probing_exps.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+#
+#SBATCH --nodes=1
+#SBATCH --ntasks-per-node=1
+#SBATCH --time=20:00:00
+#SBATCH --job-name=npi_probing
+#SBATCH --mail-type=END
+##SBATCH --mail-user=sfw268@nyu.edu
```
sleepinyourhat: Leave this out, or add a comment suggesting that people replace it.

scripts/spring_19_seminar/metadata_probing_exps.sh
```diff
@@ -0,0 +1,119 @@
...
+#SBARCH --mem=50000
+#SBATCH --output=slurm_%j.out
+#SBATCH --gres=gpu:1
+#SBATCH --partition=gpu
+
+# This script loads 46 pretrained checkpoints and run corresponding probing tasks
+# 4 models: bert, bertccg, bertmnli, bow)
```
sleepinyourhat: Typo. (NBD, just pointing it out.)

src/preprocess.py
```diff
@@ -616,6 +619,17 @@ def add_openai_bpe_vocab(vocab, namespace='openai_bpe'):
     vocab.add_token_to_namespace(utils.EOS_TOK, namespace)
 
 
+def add_wsj_vocab(vocab, data_dir, namespace='tokens'):
```
sleepinyourhat: This shouldn't be here...

sleepinyourhat: I think you messed something up when merging changes to this file.

src/tasks/__init__.py
```diff
@@ -19,6 +19,18 @@
 ALL_GLUE_TASKS = ['sst', 'cola', 'mrpc', 'qqp', 'sts-b',
                   'mnli', 'qnli', 'rte', 'wnli', 'mnli-diagnostic']
 
+# Tasks for the spring19_seminar; similar to cola but write predictions differently
```
sleepinyourhat: Big general question: Will these same tasks work for the non-probing experiments? For those, we'll need to be able to train some models on CoLA then evaluate them on the test sets for these tasks, right?

If we definitely need two different sets of NPI tasks, make it clearer in comments which ones are for which experiments. You may need to check in with @PterosDiacos.

src/preprocess.py
sheng-fu: I think this was because this branch was behind the master.... should be fixed now.

src/tasks/__init__.py
sheng-fu: I think it depends on what our BERT+cola model (and BERT_ccg+cola, BERT_mnli+cola...etc) is doing. Is cola a pretraining step or a target-task training step?

sleepinyourhat:
In some experiments, I think we're using it as part of a target task setup. See the whiteboard stuff or check with @PterosDiacos.

sheng-fu:
I left some comments at one of the scripts in #561, assuming that we are treating cola and the npi tasks as target tasks. 

What's relevant to this PS is that I think all the NPI main tasks and probing tasks can just be defined in the same way. We just run them differently: 

For probing tasks, we load model_state_eval_best.th and do target-task training on the probing data, as what the script is doing here.

For evaluation in the main tasks, e.g., target-task trained cola model on npi-x, we also load the model_state_eval_best.th from a target-trained model and use "do_target_task_training = 0, target_tasks = \"cola,npi_x\", use_classifier = cola" for eval-only runs.

sheng-fu:
I think there are two ways of doing it, one may require a reorganization of tasks/folder

1. cola and all the npi main tasks are treated as pretraining tasks (What we have now) in the 'trained on cola/npi' conditions. All the eval tasks would probably be treated as target tasks and undergo target-task training separately. 

2. cola and all the npi main tasks are treated as target tasks, so 'trained on cola/npi' conditions would mean doing target-task training. In this setting, during evaluation, we would be directly using a trained classifier from cola or the npi tasks. The issue is that for the 'all npi' and 'npi minus one' settings, we may need to lump them together as one task/folder (so that there would be one target-task classifier for it) and that requires additional organization.

src/tasks/tasks.py
```diff
@@ -414,6 +414,91 @@ def load_data(self, path, max_seq_len):
         self.test_data_text = te_data
         log.info("\tFinished loading SST data.")
```
sleepinyourhat: Tiny nit: Add a line break here.

src/tasks/tasks.py
```diff
@@ -414,6 +414,91 @@ def load_data(self, path, max_seq_len):
...
+@register_task('hd_cola_npi_cond', rel_path='NPI/combs/minus_conditionals')
+@register_task('hd_cola_npi_negdet', rel_path='NPI/combs/minus_determiner_negation_biclausal')
+@register_task('hd_cola_npi_negsent', rel_path='NPI/combs/minus_sentential_negation_biclausal')
+@register_task('hd_cola_npi_only', rel_path='NPI/combs/minus_only')
+@register_task('hd_cola_npi_ques', rel_path='NPI/combs/minus_questions')
+@register_task('hd_cola_npi_quessmp', rel_path='NPI/combs/minus_simplequestions')
+@register_task('hd_cola_npi_qnt', rel_path='NPI/combs/minus_quantifiers')
+@register_task('hd_cola_npi_sup', rel_path='NPI/combs/minus_npi_superlatives')
+class CoLANPITask(SingleClassificationTask):
+    '''Class for NPI-related task; same with Warstdadt acceptability task but outputs labels for test-set'''
```
sleepinyourhat: Add a note: Used for an NYU seminar, data not yet public.

sleepinyourhat:
Thanks! There may be further adjustments or cleanup in future PRs, but this seems useful and non-dangerous, so merging now.

PR #546, opened 2019-04-12
Cola var

PterosDiacos:
Config files for running cola-like experiments, in different models and with different pre-training tasks. Except for two files whose name contain 'ccg', all the other config files have been tested.


sleepinyourhat:
Closing—supplanted by #547

PR #547, opened 2019-04-12
Cola var

PterosDiacos:
Previously sent to "cola" branch:
Config files for running cola-like experiment in BERT or BiLSTM with different pretraining tasks


config/spring19_seminar/cola_var/cola_bert_ccg.conf
```diff
@@ -0,0 +1,77 @@
...
+
+// Output path
+project_dir = ${JIANT_PROJECT_PREFIX}
+exp_name = exp_pretrain
+run_name = run_bert_ccg
+
+
+// Logistics
+cuda = 0
+random_seed = 42
```
sleepinyourhat: Thanks! Only override things you really need to, so it's easier to sep differences between setups—I think these can just stay in defaults.conf, for example.

config/spring19_seminar/cola_var/cola_bert_ccg.conf
```diff
@@ -0,0 +1,77 @@
...
+reload_indexing = 0  // If true, force the rebuilding of the index files in preproc
+reload_vocab = 0     // If true, force the rebuilding of the vocabulary files in the experiment directory
+reindex_tasks = ""
+
+
+// Optimization
+lr = 3e-4  // this performs reasonablely well in practice
+lr_patience = 4  // number of epochs between last validation improvement and lr annealing, following final.conf
+patience = 20  // number of epochs between last validation improvement and early stopping, following final.conf
+max_vals = 10000 // following final.conf
```
sleepinyourhat: If you're following final.conf in a lot of places, just inherit from there instead of from defaults.conf.

config/spring19_seminar/cola_var/cola_bert_mnli.conf
```diff
@@ -0,0 +1,77 @@
...
+write_preds = "val,test"  // 0 for none, or comma-separated splits in {"train", "val", "test"} 
+                          // for which predictions are written to disk during do_full_eval
+
+
+// Pretraining tasks
+load_model = 0  // If true, restore from checkpoint when starting do_pretrain
+do_pretrain = 1
+allow_untrained_encoder_parameters = 0  // Set for experiments involving random un(-pre-)trained encoders only.
+pretrain_tasks = "mnli" // Comma-separated list of pretraining tasks or "glue" or "none";
+                                               // If there are multiple entries, the list should contain no spaces, and quotation marks are needed;
```
sleepinyourhat: If these two files only differ in cola vs. mnli, it's cleanest to use a single file for them both and use command line overrides to change pretrain_tasks.

config/spring19_seminar/cola_var/cola_bilstm_ccg.conf
```diff
@@ -0,0 +1,70 @@
...
+
+
+
+// Model
+
+// Model, biLSTM
+tokenizer = "MosesTokenizer" // The default tokenizer
+sent_enc = "rnn"  // "bow", "rnn" for LSTM, "null"
+bidirectional = 1
+word_embs = "scratch"
```
sleepinyourhat: This might hurt performance quite a bit, since you'll have a lot of OOV words. Consider using GloVe or fastText embeddings, or the ELMo word-level-only setup. Whatever's easy.

config/spring19_seminar/cola_var/cola_bert_ccg.conf
sleepinyourhat: Also, probably best to use batch_size=16 everywhere. It's the default in jiant, and it's what @sheng-fu and @HaokunLiu are using.

config/spring19_seminar/cola_var/cola_bert_ccg.conf
PterosDiacos: @sleepinyourhat  in the updated config files the default settings have been removed

config/spring19_seminar/cola_var/cola_bert_ccg.conf
PterosDiacos: @sleepinyourhat  now it's a batch_size of 16. Also used including final.conf

config/spring19_seminar/cola_var/cola_bilstm_ccg.conf
PterosDiacos: @sleepinyourhat  now it's 'glove':)

config/spring19_seminar/cola_var/cola_bert_mnli.conf
PterosDiacos: I've restructured the config files following this vein; now there's one config for each model. I also uploaded the updated script for generating both the configs and the command lines for running main.py with relevant overrides, and the command lines to running cola-inference

config/spring19_seminar/cola_var/cola_bert.conf
```diff
@@ -0,0 +1,38 @@
+// Import the defaults using the relative path
+include "../final.conf"
+
+
+// Output path
+project_dir = ${JIANT_PROJECT_PREFIX}
```
sleepinyourhat: Is there a reason you need to override this?

config/spring19_seminar/cola_var/cola_bert.conf
```diff
@@ -0,0 +1,38 @@
...
+                          // for which predictions are written to disk during do_full_eval
+
+
+// Pretraining tasks
+load_model = 0  // If true, restore from checkpoint when starting do_pretrain
+
+
+// Model
+// Models, BERT
+batch_size = 16
```
sleepinyourhat: If you inherit from final_bert, you won't have to set several of these.

config/spring19_seminar/cola_var/cola_bert.conf
```diff
@@ -0,0 +1,38 @@
...
+dropout = 0.1 // following BERT paper
+tokenizer = "bert-large-uncased"
```
sleepinyourhat: Let's use cased.

config/spring19_seminar/cola_var/cola_bilstm.conf
```diff
@@ -0,0 +1,31 @@
...
+write_preds = "val,test"  // 0 for none, or comma-separated splits in {"train", "val", "test"} 
+                          // for which predictions are written to disk during do_full_eval
+
+
+// Pretraining tasks
+load_model = 0  // If true, restore from checkpoint when starting do_pretrain
+
+
+// Model
+// Model, biLSTM
```
sleepinyourhat: Probably simplest to use the same batch size for both the BiLSTM and BERT.

config/spring19_seminar/cola_var/conf_gen.py
```diff
@@ -0,0 +1,78 @@
+import scratch
```
sleepinyourhat: What is this doing that we can't do with overrides and one or two config files? I worry that this'll just make debugging and comparison harder.

config/spring19_seminar/cola_var/addToRunMain.txt
```diff
@@ -0,0 +1,24 @@
+python main.py --config_file cola_confs\bert_ccg.conf \
```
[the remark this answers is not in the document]
PterosDiacos: @sleepinyourhat  yes, these lines are to be added into a bash file. Txt file here is just for illustration.
conf_gen.py produces backslashes on windows but forward ones on linux.

config/spring19_seminar/cola_var/cola_bert.conf
PterosDiacos: In default.conf and final.conf project_dir was set to  ${NFS_PROJECT_PREFIX}

config/spring19_seminar/cola_var/conf_gen.py
PterosDiacos: It generates the two config files in question and command lines with desired overrides

config/spring19_seminar/cola_var/cola_bilstm.conf
PterosDiacos: fixed

config/spring19_seminar/cola_var/cola_bert.conf
PterosDiacos: fixed

config/spring19_seminar/cola_var/addToRunMain.txt
```diff
@@ -0,0 +1,24 @@
...
+    --overrides "exp_name = exp_pretrain, run_name = run_bert_ccg, target_tasks = "cola", pretrain_tasks = "ccg", do_pretrain = 1"
```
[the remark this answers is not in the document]
PterosDiacos: fixed the bug

config/spring19_seminar/cola_var/addToRunMain.txt
```diff
@@ -0,0 +1,24 @@
...
+    --overrides "exp_name = exp_pretrain, run_name = run_bert_ccg, target_tasks = "cola", pretrain_tasks = "ccg", do_pretrain = 1" 
+
+python main.py --config_file cola_confs\bert_mnli.conf \
+    --overrides "exp_name = exp_pretrain, run_name = run_bert_mnli, target_tasks = "cola", pretrain_tasks = "mnli", do_pretrain = 1" 
+
+python main.py --config_file cola_confs\bert_none.conf \
+    --overrides "exp_name = exp_pretrain, run_name = run_bert_none, target_tasks = "cola", pretrain_tasks = "none", do_pretrain = 0, allow_untrained_encoder_parameters = 1" 
+
+python main.py --config_file cola_confs\bert_sst.conf \
+    --overrides "exp_name = exp_pretrain, run_name = run_bert_sst, target_tasks = "cola", pretrain_tasks = "sst", do_pretrain = 1"
```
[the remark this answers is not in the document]
PterosDiacos: that was just for testing 'conf_gen.py':)

config/spring19_seminar/cola_var/cola_bert.conf
sleepinyourhat: Why not just set NFS_PROJECT_PREFIX in your path config file?

config/spring19_seminar/cola_var/addToRunMain.txt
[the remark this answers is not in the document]
sleepinyourhat: Why not just write the bash file without the script? Easier to read/debug if there are only a dozen or so runs.

config/spring19_seminar/cola_var/conf_gen.py
sleepinyourhat: I'd suggest keeping this private and just checking in the config files and bash files. They'll be much easier to read and much more standard for jiant.

PterosDiacos: definitely:) this was for the original research questions where I thought there might be much more pretrain/model configurations to compare

config/spring19_seminar/cola_var/cola_bert.conf
PterosDiacos: It's in the 'Start Guide' and configs from Jason; I thought everyone was sharing that

PterosDiacos: I'll fix this

sleepinyourhat: NBD—either way works. Just make sure you're consistent with other PRs.

config/spring19_seminar/cola_var/conf_gen.py
sleepinyourhat: Yeah. Mind removing for now? This is pretty close to being ready to merge in.

PterosDiacos: Done!

sleepinyourhat:
This looks pretty much done.
– @PterosDiacos: Make sure that you've resolved any outstanding questions you have about things like hyperparameters.
– Everyone else (including @sheng-fu @HaokunLiu @W4ngatang @phu-pmh @alexwarstadt @anhad13 ...), take a careful look at this soon to make sure it fits with what you were expecting, and add comments with either an approval or questions/suggestions. This is our last chance to spot issues here.

config/spring19_seminar/cola_var/cola_bert.conf
```diff
@@ -0,0 +1,37 @@
+// Import the defaults using the relative path
```
sleepinyourhat: The filename is a bit misleading—I don't see anything about cola here.

config/spring19_seminar/cola_var/sample.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+#SBATCH --time=12:00:00
+#SBATCH --mem=20000
+#SBATCH --gres=gpu:1
+#SBATCH --job-name=myrun
+#SBATCH --output=slurm_%j.out
+
+python main.py --config_file cola_confs/cola_bert.conf \
```
sleepinyourhat: cola_confs or cola_var?

config/spring19_seminar/cola_var/cola_bert.conf
sheng-fu: This actually pretty much does what my finetuning config file wants to do. The difference is only in the pretraining tasks, which can be easily supplied with overrides. So I think I don't need to check my version.

As for my target-task config file, there's only some additional things about loading a local check point, and of course, the target tasks. Maybe we also just do that with overrides? In that case, I don't need to check in that either.

If so, my pull request would simply include (1) some sbatch scripts that mainly use this config for probing experiments (2) task registration, either through adding stuff on tasks.py or in the subtask-definition config if we are using Jason's subtasks setup. (3) adding the tasks to write_preds() in src/evaluate.py

sheng-fu:
I think I can just use cola_bert.conf (and I agree with Sam, the filename is a bit strange; for me it's more like a general setup config) in this PR with overrides to do what my fine-tuning script does. So I won't check in that config file.

As for my target-task config file, there's only some additional things about loading a local check point, and of course, the target tasks. Maybe I should just do that with overrides? In that case, I don't need to check in that either.

If so, my pull request would simply include (1) some sbatch scripts that mainly use this config for probing experiments (2) task registration, either through adding stuff on tasks.py or in the subtask-definition config if we are using Jason's subtasks setup. (3) adding the tasks to write_preds() in src/evaluate.py

config/spring19_seminar/cola_var/cola_bert.conf
```diff
@@ -0,0 +1,38 @@
...
+include "../final.conf"
```
sheng-fu: This has "max_epochs = -1"; are we doing this?

sleepinyourhat: Good catch. We should probably stick to a 3-epoch limit during both intermediate tuning and target task fine-tuning.

config/spring19_seminar/cola_var/sample.sh
PterosDiacos: cola_confs is the folder I store those config files on my laptop

config/spring19_seminar/cola_var/cola_bert.conf
PterosDiacos: fixed now

config/spring19_seminar/cola_var/cola_bert.conf
PterosDiacos: fixed

config/spring19_seminar/cola_var/bert.conf
```diff
@@ -0,0 +1,38 @@
+// Import the defaults using the relative path
+include "../final.conf"
```
sheng-fu: Another comment about epochs: the defaults.conf has val_interval = 1000 and that percolates here. When training on CoLA, 2000 batches are already 3.738 epochs. Should we set a smaller number?

sleepinyourhat:
Yep! You should make sure that val_interval is much less than one epoch.
Maybe 100? (We should enforce this automatically—mind adding a GitHub
issue?)

config/spring19_seminar/cola_var/bert.conf
PterosDiacos: Now val_interval and eval_val_interval are both set to 100

PterosDiacos:
I also had the question whether we should use max_vals =10000 as in final.conf or smaller, max_vals = 1000  as in default.conf

config/spring19_seminar/cola_var/sample.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
sleepinyourhat: Move to scripts/spring19_seminar

config/spring19_seminar/cola_var/sample.sh
```diff
@@ -0,0 +1,31 @@
...
+#SBATCH --mem=20000
+#SBATCH --gres=gpu:1
+#SBATCH --job-name=myrun
+#SBATCH --output=slurm_%j.out
+
+python main.py --config_file config/spring19_seminar/bert.conf \
+    --overrides "exp_name = npi_bert, run_name = run_bert_mnli_cola, target_tasks = "cola", pretrain_tasks = "mnli", do_pretrain = 1" 
+
+python main.py --config_file config/spring19_seminar/bert.conf \
+    --overrides "exp_name = npi_bert, run_name = run_bert_mnli_colax, target_tasks = "colax", pretrain_tasks = "mnli", do_pretrain = 1"
```
sleepinyourhat: Where's the code for the colax task? We'd have to check that in before checking this in.

sleepinyourhat:
It shouldn't matter—max_vals and max_epochs do the same thing, just measured in different units, and 1000 and 10000 should both be more than 3 epochs.

config/spring19_seminar/cola_var/sample.sh
PterosDiacos: Done!

PterosDiacos: Also relocated the configs under config/spring19_seminar (the cola_var folder is removed)

config/spring19_seminar/cola_var/sample.sh
PterosDiacos: Now they're there. Simply run each thing from scratch.

PterosDiacos: I added these lines into my local version. Not sure if this is helpful --- if we have a long list of cola-like tasks, all we need to do then is extend ALL_COLAX_TASKS

in "src/tasks/__init__.py"
ALL_COLAX_TASKS = ['colax']
ALL_GLUE_TASKS = ['sst', 'cola', 'mrpc', 'qqp', 'sts-b',
                  'mnli', 'qnli', 'rte', 'wnli', 'mnli-diagnostic'] + ALL_COLAX_TASKS

in "tasks.py"
from .__init__ import ALL_COLAX_TASKS
for colax in ALL_COLAX_TASKS:
    CoLATask = register_task(colax, rel_path='CoLA/%s/' % colax)(CoLATask)

sleepinyourhat: It doesn't really make sense to share/check-in the configs without this code. Go ahead and add it! Though I'd suggest something more informative than 'colax'. Maybe 'cola-npi-diagnostic'?

PterosDiacos: Just added! Called them cola_npi_{some environment name}

scripts/spring19_seminar/colax_sample.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
sleepinyourhat: Rename from colax.

src/tasks/__init__.py
```diff
@@ -14,10 +14,15 @@
...
 ##
+# All cola-like tasks for which we need write_preds() and append to ALL_GLUE_TASKS below
+ALL_COLA_NPI_TASKS = ['cola_npi_x']
+
+
 # Task lists for handling as a group; these names correspond to the keys in
 # the task registry.
 ALL_GLUE_TASKS = ['sst', 'cola', 'mrpc', 'qqp', 'sts-b',
-                  'mnli', 'qnli', 'rte', 'wnli', 'mnli-diagnostic']
+                  'mnli', 'qnli', 'rte', 'wnli', 'mnli-diagnostic'] + ALL_COLA_NPI_TASKS
```
sleepinyourhat: Remove this—it'll impact any experiment anyone outside does with GLUE.

src/tasks/__init__.py
```diff
@@ -14,10 +14,15 @@
 from .tasks import Task
 
 ##
+# All cola-like tasks for which we need write_preds() and append to ALL_GLUE_TASKS below
+ALL_COLA_NPI_TASKS = ['cola_npi_x']
```
sleepinyourhat: If there's only one item, you don't need a list—just refer to this directly later.

src/tasks/tasks.py
```diff
@@ -455,6 +455,13 @@ def update_metrics(self, logits, labels, tagmask=None):
         self.scorer2(logits, labels)
         return
 
+# register cola-like tasks with a non-decorator syntax
+# put data for colax in a subfoler named colax under CoLA
```
sleepinyourhat: typo

src/tasks/tasks.py
```diff
@@ -455,6 +455,13 @@ def update_metrics(self, logits, labels, tagmask=None):
         self.scorer2(logits, labels)
         return
 
+# register cola-like tasks with a non-decorator syntax
+# put data for colax in a subfoler named colax under CoLA 
+from .__init__ import ALL_COLA_NPI_TASKS
+for cola_npi_x in ALL_COLA_NPI_TASKS:
+    CoLATask = register_task(cola_npi_x, rel_path='CoLA/%s/' % cola_npi_x)(CoLATask)
```
sleepinyourhat: Right now this only creates one task. In any case, though, why not just name all of them? We can commit changes if we change the data—that's how it works for every other task here.

src/tasks/__init__.py
PterosDiacos: There should be nine. Still waiting for the data generation team's decision

src/tasks/tasks.py
PterosDiacos: there will be more --- the idea is to put all hand-coding into `__init__.py`, which we need to do anyway if we want these added tasks to be taken care by `write_preds()` in `evaluate.py`

scripts/spring19_seminar/colax_sample.sh
PterosDiacos: Done!

src/tasks/__init__.py
PterosDiacos: Done. Relocated that to `evaluate.py`

src/tasks/tasks.py
PterosDiacos: Fixed

src/tasks/tasks.py
sleepinyourhat: Let's just treat this the same way as MNLI, even if that means naming things in multiple places: Register everything where we define the task. Handling different tasks different ways will make this much harder to maintain later.

Also, are you sure the set isn't final? I think we need to set the code up for what the NPI team has now either way.

PterosDiacos: Updated!

scripts/spring19_seminar/cola_npi_x_sample.sh
```diff
@@ -0,0 +1,367 @@
+#!/bin/bash
+#SBATCH --time=12:00:00
+#SBATCH --mem=20000
+#SBATCH --gres=gpu:1
+#SBATCH --job-name=myrun
+#SBATCH --output=slurm_%j.out
+
```
sleepinyourhat: Add a comment here summarizing the naming scheme, and summarizing what combinations you're running.

src/tasks/tasks.py
PterosDiacos: I mean the train-dev-test split has not been out for each environment yet

scripts/spring19_seminar/cola_npi_x_sample.sh
sleepinyourhat: For the MNLI runs, it'll probably _much_ faster to train sequentially, since MNLI is pretty big:

- Run jiant to train a model on MNLI as a pretraining task.
- Then, for each CoLA/NPI config you're interested in, load the MNLI model (using a specified path), then pretrain on only the CoLA/NPI data. If it's not clear how to set this up, @sheng-fu or @zphang should be able to help.

src/tasks/tasks.py
sleepinyourhat: Hrm. Bug the data folks? It was supposed to be done by this point... (@alexwarstadt — do you know where this stands?)

alexwarstadt: Split data is now on the data_generation repo: https://github.com/alexwarstadt/data_generation/tree/master/outputs/npi/environments/splits, with thanks to @sheng-fu

config/spring19_seminar/bert.conf
```diff
@@ -0,0 +1,40 @@
...
+
+
+// Pretraining tasks
+load_model = 0  // If true, restore from checkpoint when starting do_pretrain
+
+
+// Models
+// Model, BERT
+tokenizer = "bert-large-cased"
+sent_enc = "null" // "bow", "rnn" for LSTM, "null"
```
sheng-fu: Maybe change to sent_enc = "none" and also clean up the comments?

PterosDiacos: seems that `sentence_encoder` function in `models.py` makes references to `null`

sleepinyourhat: The change from 'null' to 'none' was part of a recent bug fix—are you sure you're using the latest version of master? I don't see 'null' (in quotes) anywhere in models.py.

src/evaluate.py
```diff
@@ -128,7 +128,18 @@ def write_preds(tasks: Iterable[tasks_module.Task], all_preds, pred_dir, split_n
         preds_df = all_preds[task.name]
         # Tasks that use _write_glue_preds:
         glue_style_tasks = (tasks_module.ALL_NLI_PROBING_TASKS
-                            + tasks_module.ALL_GLUE_TASKS + ['wmt'])
+                            + tasks_module.ALL_GLUE_TASKS + ['wmt'] 
+                            + [ 'cola_npi_sup',
```
W4ngatang: sorry to be annoying, but could you define a custom list of tasks in `src/tasks/__init__.py` rather than listing them all here? Hopefully that should save you some work in other places if you need to reference this list again.

PterosDiacos: I set a `ALL_COLA_NPI_TASKS` in `src/tasks/__init__.py` before, which went with automated task registration lines in `src/tasks/tasks.py`. But since now the latter has been removed, I fall back to hand-coding here in `src/tasks/__init__.py` too to keep consistency

config/spring19_seminar/bert.conf
PterosDiacos: will fix this

PterosDiacos: done!

sleepinyourhat:
Discussion continues in #561.

Issue #541, opened 2019-04-08
'null' not identified as valid sentence encoder specification


xiaoyizy:
Hi,

Setting `sent_enc = null` in conf script caused error "No valid sentence encoder specified" from line 135 of `models.py`. Further tests showed that `args.sent_enc == "null"` returned `False` under `build_sent_encoder`. 

After changing `"null"` to `"none"` in `build_sent_encoder` else/if conditions, the code was functional.

pruksmhc:
We currently use HOCON for the config language. You probably want to try sent_enc = "null"  (with quotation marks).

xiaoyizy:
Thanks for your advice! However quotation mark had no effect on this issue. Running `print(args.sent_enc)` and `args.sent_enc == 'null'` returned `'null'` and `False`.

HaokunLiu:
Ha! I have done this before. It should be double quotation mark. "null", not 'null'.

sleepinyourhat:
Should be fixed! Changed the expected keyword from 'null' to 'none' to avoid some tricky HOCON escaping issues.

PR #551, opened 2019-04-14
Replace 'null' with 'none' in config files to avoid HOCON issues

sleepinyourhat:
See #541. Unlike most other string-valued config options, the word 'null' needs to be quoted in config files, which is easy to mess up. This fixes that.

(A few autopep8 changes are mixed in too.)


sleepinyourhat:
@iftenney Mind taking a look at this after all?

PR #557, opened 2019-04-18
Clarifying main and eval -> pretrain and target train stages

pruksmhc:
Since we're still sticking to 2-phase structure for now, I did a bit of cleanup around naming consistency with pretrain and finetuning stages. 
This PR includes: 
Changing main -> pretrain 
eval -> finetune

Tests already run: 
-Demo.conf 
-Only do_pretrain = 1, and do_target_task_training = 0
-The reverse of above
-load_eval_checkpoint = 1 and do_full_eval = 1 (do_pretrain = 0, do_target_task_training = 0)


config/defaults.conf
```diff
@@ -95,8 +95,8 @@ reload_vocab = 0     // If true, force the rebuilding of the vocabulary files in
 is_probing_task = 0  // Set if the tasks in target_tasks are NLI-style probing tasks. (Ellie: Is this right?)
 
 // Learning curves
-training_data_fraction = 1  // Use only the specified fraction of the training data in the do_pretrain phase.
-                            // Should not impact eval-phase training, even for the same task.
+pretraining_data_fraction = 1  // Use only the specified fraction of the training data in the do_pretrain phase.
```
sleepinyourhat: Be consistent with part of speech: pretrain/finetune or pretraining/finetuning

sleepinyourhat: Actually, a suggestion: pretraining/target_training. The term _fine-tuning_ isn't ideal, since it _also_ refers to a separate decision about which parameters to update.

pruksmhc: Let's stick with target task then.

main.py
```diff
@@ -88,25 +88,25 @@ def _kill_tb_child():
...
     Hierarchy is
-        1) best checkpoint from eval (target_task_training)
+        1) best checkpoint from target_task_training
         2) best checkpoint from pretraining
         3) checkpoint created from before any target task training
         4) nothing found (empty string) """
-    eval_best = glob.glob(os.path.join(run_dir, "model_state_eval_best.th"))
-    if len(eval_best) > 0:
-        assert_for_log(len(eval_best) == 1,
+    target_task_best = glob.glob(os.path.join(run_dir, "model_state_target_train_best.th"))
```
sleepinyourhat: Send out some @channel messages about this once it's checked in, since this breaks old checkpoints.

sleepinyourhat: Or try to set this up so it'll recognize old checkpoints.

config/defaults.conf
```diff
@@ -79,13 +79,13 @@ allow_reuse_of_pretraining_parameters = 0  // Set to 1 to allow task models that
                                            // reused in do_target_task_training. This may behave incorrectly if a run is stopped and
                                            // restarted in do_target_task_training (issues #285, #290).
 allow_missing_task_map = 0  // Dangerous: If true, ignore missing classifier_task_map.json
-                            // This is needed for bare-ELMo probing, since the main training phase is skipped for these models.
+                            // This is needed for bare-ELMo probing, since the pretraining training phase is skipped for these models.
```
W4ngatang: "pretraining training" -> "pretraining"

config/defaults.conf
```diff
@@ -79,13 +79,13 @@ allow_reuse_of_pretraining_parameters = 0  // Set to 1 to allow task models that
...
 reload_tasks = 0     // If true, force the rebuilding of the task files in the experiment directory, even if they exist.
 reload_indexing = 0  // If true, force the rebuilding of the index files in preproc/ for tasks in reindex_tasks, even if they
                      // exist.
-eval_data_fraction = 1  // Use only the specified fraction of the training data in the do_target_task_training phase.
-                        // should not impact main training, even for the same task.
-                        // training_data_fraction should not be used together with eval_data_fraction
+target_train_data_fraction = 1  // Use only the specified fraction of the training data in the do_target_task_training phase.
+                        // should not impact pretrain training, even for the same task.
+                        // training_data_fraction should not be used together with                 // target_train_data_fractio
```
W4ngatang: "fraction"

Also, random request: can you add that it's a random fraction of the data (I think)?

src/preprocess.py
```diff
@@ -351,50 +351,50 @@ def build_tasks(args):
     log.info("\tFinished indexing tasks")
 
     # 5) Initialize tasks with data iterators.
-    assert not (args.training_data_fraction < 1 and args.eval_data_fraction < 1), \
-        "training_data_fraction and eval_data_fraction could not be used at a same time (could not be < 1 together)"
+    assert not (args.pretrain_data_fraction < 1 and args.target_train_data_fraction < 1), \
```
W4ngatang: why is this assert a thing? It seems like these should be independent. Maybe a question for @sleepinyourhat

src/trainer.py
```diff
@@ -27,10 +27,10 @@
 from .utils import config
 
 
-def build_trainer_params(args, task_names):
+def build_trainer_params(args, task_names, phase='pretrain'):
     ''' Helper function which extracts trainer parameters from args. In particular, we want to search args
-    for task specific training parameters. '''
-    def _get_task_attr(attr_name): return config.get_task_attr(
+    for task spr(ecific training parameters. '''
```
W4ngatang: specific

src/trainer.py
```diff
@@ -27,10 +27,10 @@
...
+    def _get_task_attattr_name): return config.get_task_attr(
```
W4ngatang: breaking bug

src/trainer.py
```diff
@@ -27,10 +27,10 @@
 from .utils import config
 
 
-def build_trainer_params(args, task_names):
+def build_trainer_params(args, task_names, phase='pretrain'):
```
W4ngatang: how come we didn't need to pass this argument in before?

src/trainer.py
```diff
@@ -44,7 +44,13 @@ def _get_task_attr(attr_name): return config.get_task_attr(
...
         params[attr] = _get_task_attr(attr)
     for attr in extra_opts:
-        params[attr] = getattr(args, attr)
+        if attr == 'training_data_fraction':
+            if phase == 'pretrain':
+                params[attr] = getattr(args, 'pretrain_data_fraction')
+            else:
+                params[attr] = getattr(args, 'target_train_data_fraction')
+        else:
+             params[attr] = getattr(args, attr)
```
iftenney: nit: `args[attr]` should also work, and strongly prefer `args.pretrain_data_fraction` to using getattr with static arguments.

config/defaults.conf
pruksmhc: It seems to always be set to 1.0

src/trainer.py
pruksmhc: because before, we used the same value for both pretraining and target task training stages, so if we want to separate it out we have to pass the phase in

src/preprocess.py
pruksmhc: Yeah I agree actually

pruksmhc: Another issue to make! 
I think we should have a separate PR for this, because this will change logging/ the way we're logging our metrics, and will require further testing.

sleepinyourhat: I vaguely remember that this happens during data loading, and that they both use the same mechanism. Should be an easy-ish fix, but I'm not sure why we need to support this.

PR #560, opened 2019-04-22
Improve main.py codeflow

pruksmhc:
Improved code flow of main.py file. 
Main changes: 
-Modularizing:
   -setup for the main loop and target task training 
   -checking for consistent configerations 

Tested on demo.conf


sleepinyourhat:
Looking now

main.py
```diff
@@ -126,16 +250,26 @@ def evaluate_and_write(args, model, tasks, splits_to_write):
...
-    args = config.params_from_file(cl_args.config_file, cl_args.overrides)
-
-    # Raise error if obsolete arg names are present
-    check_arg_name(args)
-
-    # Logistics #
+def setup(args, cl_args):
+    """
+    Sets up email hook, cuda settings, and various initial sanity checks such as 
+    not mixing tasks with minimizing and maximizing validation set objective functions.
```
sleepinyourhat: Can we put these sanity checks with the other sanity checks in the earlier method?

main.py
```diff
@@ -195,56 +329,22 @@ def main(cl_arguments):
...
-                       "allow_reuse_of_pretraining_parameters = 1(risky), or train in two steps:\n"
-                       "  train with do_pretrain = 1, do_target_task_training = 0, stop, and restart with\n"
-                       "  do_pretrain = 0 and do_target_task_training = 1.")
-
-    if args.do_full_eval:
-        assert_for_log(args.target_tasks != "none",
-                       "Error: Must specify at least one eval task: [%s]" % args.target_tasks)
-        steps_log.append("Evaluating model on tasks: %s" % args.target_tasks)
-
+def main(cl_arguments):
```
sleepinyourhat: Since this is the piece of code that we expect people to edit most often, let's comment it _to death_. Explain every block of code, at least one comment for every two or three lines, unless it's _blindingly_ obvious what's going on.

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+                         args.cuda, task_names_to_avoid_loading, strict=strict)
+    else:
+        # Look for eval checkpoints (available only if we're restoring from a run that already
+        # finished), then look for training checkpoints.
+
+        if args.transfer_paradigm == "finetune":
+            # Save model so we have a checkpoint to go back to after each task-specific finetune.
+            model_state = model.state_dict()
+            model_path = os.path.join(args.run_dir, "model_state_untrained_prefinetune.th")
+            torch.save(model_state, model_path)
```
W4ngatang: Not something we have to address this PR, but something seems off here (likely my fault). We should only need to create this checkpoint if we don't find a best checkpoint later. If we do find a best checkpoint to load, we're extraneously creating this checkpoint.

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+        if best_path:
+            load_model_state(model, best_path, args.cuda, task_names_to_avoid_loading,
+                             strict=strict)
+        else:
+            assert_for_log(args.allow_untrained_encoder_parameters,
+                           "No best checkpoint found to evaluate.")
+            log.warning("Evaluating untrained encoder parameters!")
+    return task_names_to_avoid_loading, strict
+
+def check_configeration(args, pretrain_tasks, target_tasks):
```
W4ngatang: configurations

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+    """
+    Checks configerations for any obvious logical flows
```
W4ngatang: configurations

W4ngatang: also, "flaws"?

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+            "Error: Attempting to train a model and then replace that model with one from a checkpoint.")
+        steps_log.append("Loading model from path: %s" % args.load_eval_checkpoint)
+
+    assert_for_log(args.transfer_paradigm in ["finetune", "frozen"],
+                   "Transfer paradigm %s not supported!" % args.transfer_paradigm)
+
+    if args.do_pretrain:
+        assert_for_log(args.pretrain_tasks != "none",
+                       "Error: Must specify at least on training task: [%s]" % args.pretrain_tasks)
+        assert_for_log(
```
W4ngatang: similar to Sam's point below, would be fine just to get rid of `bpp` param

W4ngatang: Probably not this PR

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+        steps_log.append("Re-training model for individual eval tasks")
+        assert_for_log(
+            args.eval_val_interval %
+            args.bpp_base == 0, "Error: eval_val_interval [%d] must be divisible by bpp_base [%d]" %
+            (args.eval_val_interval, args.bpp_base))
+        assert_for_log(len(set(pretrain_tasks).intersection(target_tasks)) == 0
+                       or args.allow_reuse_of_pretraining_parameters
+                       or args.do_pretrain == 0,
+                       "If you're pretraining on a task you plan to reuse as a target task, set\n"
+                       "allow_reuse_of_pretraining_parameters = 1(risky), or train in two steps:\n"
```
W4ngatang: worth explaining why this is risky?

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+                       "  train with do_pretrain = 1, do_target_task_training = 0, stop, and restart with\n"
+                       "  do_pretrain = 0 and do_target_task_training = 1.")
+
+    if args.do_full_eval:
+        assert_for_log(args.target_tasks != "none",
```
W4ngatang: I actually don't understand this. What if I want to do a full evaluation on the test sets of my pretraining tasks?

main.py
```diff
@@ -260,49 +360,9 @@ def main(cl_arguments):
...
-    if not args.load_eval_checkpoint == "none":
-        # This is to load a particular eval checkpoint.
-        log.info("Loading existing model from %s...", args.load_eval_checkpoint)
-        load_model_state(model, args.load_eval_checkpoint,
-                         args.cuda, task_names_to_avoid_loading, strict=strict)
-    else:
-        # Look for eval checkpoints (available only if we're restoring from a run that already
-        # finished), then look for training checkpoints.
-
-        if args.transfer_paradigm == "finetune":
```
W4ngatang: We've moved this into the loop over all the target tasks, so we're now saving a new model for each target task when we should not be (this model has been fine-tuned for a particular target task and shouldn't be used for another task).

main.py
```diff
@@ -260,49 +360,9 @@ def main(cl_arguments):
...
-            load_model_state(model, best_path, args.cuda, task_names_to_avoid_loading,
-                             strict=strict)
-        else:
-            assert_for_log(args.allow_untrained_encoder_parameters,
-                           "No best checkpoint found to evaluate.")
-            log.warning("Evaluating untrained encoder parameters!")
-
     # Train just the task-specific components for eval tasks.
     if args.do_target_task_training:
+        task_names_to_avoid_loading, strict = setup_target_task_training(args, target_tasks, model)
```
W4ngatang: Actually I think it would be correct if we just moved this outside of the loop?

main.py
```diff
@@ -126,16 +250,26 @@ def evaluate_and_write(args, model, tasks, splits_to_write):
...
-def main(cl_arguments):
-    ''' Train or load a model. Evaluate on some tasks. '''
-    cl_args = handle_arguments(cl_arguments)
-    args = config.params_from_file(cl_args.config_file, cl_args.overrides)
-
-    # Raise error if obsolete arg names are present
-    check_arg_name(args)
-
-    # Logistics #
+def setup(args, cl_args):
```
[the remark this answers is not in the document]
W4ngatang: This seems like a really odd way to break stuff up. It makes sense to have all the folder creation and logging in one function (including tensorboard?) but it seems like people would want to be able to see where the model is being created and where the tasks are being created without having to go through this extra layer of code. I think it'd be easier to follow if we had a `setup_logging()`, `setup_model()`, `setup_tasks()` functions.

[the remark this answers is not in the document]
pruksmhc: Yeah, on second though the model = build_model should be moved out of this initial setup function. There's a lot of initial setup though (like making the directories, email notifier hooks, etc), which should be moved out of the huge main.py function

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+    target_tasks: list of target Task objects
+    mdoel: a MultTaskModel object
+
+    Returns
+    ----------------
+    task_names_to_avoid_loading: list of strings, if we don't allow for 
+    use of pretrained target specific module parameters, then this list will 
+    cosnist of all the task names so that we avoid loading the 
+    pretrained parameters. Else, it will be an empty list.
+    strict: bool, indicates whether to exit if the loaded model has some expected
```
iftenney: This name is a bit weird - it seems like something you should be passing _to_ this function, not returning.

Looking at the code below, the control flow is also a bit odd - since this value can be computed from `args` alone, would it be cleaner to just compute outside in the calling code and pass as an argument here?

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+    ----------------
+    args: Params object
+    pretrain_tasks: list of pretraining Task objects
+    target_tasks: list of target task training Task objects
+
+    Returns
+    ----------------
+    None
+    """
+    steps_log = []
```
iftenney: Why not just log as you go?

Otherwise, prefer using [`io.StringIO`](https://docs.python.org/3/library/io.html) to build up a string incrementally.

main.py
pruksmhc: We should actually have a eval_tasks list too, since right now, if you want to do what you said, we have to do another run where you create load_eval_checkpoint to a pretraining task, make the target_tasks the pretrain task you want to do an evaluation on. (I actually haven't tried this myself, but this is what I would think you'd have to do).

main.py
pruksmhc: It's not in a loop?

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
     return parser.parse_args(cl_arguments)
 
 
-def _try_logging_git_info():
+def setup_target_task_training(args, target_tasks, model):
+    """
+    Save model states from pretrianing if applicable, and
```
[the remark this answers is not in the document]
sleepinyourhat: pretrianing. Please proofread more carefully!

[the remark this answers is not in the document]
sleepinyourhat: Also, save...loads should be saves...loads.

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+    ----------------
+    args: Params object
+    target_tasks: list of target Task objects
+    mdoel: a MultTaskModel object
+
+    Returns
+    ----------------
+    task_names_to_avoid_loading: list of strings, if we don't allow for 
+    use of pretrained target specific module parameters, then this list will 
+    cosnist of all the task names so that we avoid loading the
```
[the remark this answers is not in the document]
sleepinyourhat: ...

main.py
sleepinyourhat: I agree. Not strongly opposed to merging this way, but it's definitely not making the code more readable/editable.

main.py
```diff
@@ -53,7 +53,131 @@ def handle_arguments(cl_arguments):
...
+    if args.do_target_task_training and not args.allow_reuse_of_pretraining_parameters:
+        # If we're training models for evaluation, which is always done from scratch with a fresh
+        # optimizer, we shouldn't load parameters for those models.
+        # Usually, there won't be trained parameters to skip, but this can happen if a run is killed
+        # during the do_target_task_training phase.
+        task_names_to_avoid_loading = [task.name for task in target_tasks]
+    else:
+        task_names_to_avoid_loading = []
+
+    if not args.load_eval_checkpoint == "none":
```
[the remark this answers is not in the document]
sleepinyourhat: Small point, but I agree.

main.py
sleepinyourhat: Add an issue if not fixing now?

main.py
sleepinyourhat: (Also, space before '(')

main.py
```diff
@@ -53,7 +53,136 @@ def handle_arguments(cl_arguments):
...
+                       or args.allow_reuse_of_pretraining_parameters
+                       or args.do_pretrain == 0,
+                       "If you're pretraining on a task you plan to reuse as a target task, set\n"
+                       "allow_reuse_of_pretraining_parameters = 1(risky), or train in two steps:\n"
+                       "  train with do_pretrain = 1, do_target_task_training = 0, stop, and restart with\n"
+                       "  do_pretrain = 0 and do_target_task_training = 1.")
+
+    if args.do_full_eval:
+        assert_for_log(args.target_tasks != "none",
+                       "Error: Must specify at least one eval task: [%s]" % args.target_tasks)
```
sleepinyourhat: => target task

main.py
```diff
@@ -126,16 +255,24 @@ def evaluate_and_write(args, model, tasks, splits_to_write):
...
-    cl_args = handle_arguments(cl_arguments)
-    args = config.params_from_file(cl_args.config_file, cl_args.overrides)
-
-    # Raise error if obsolete arg names are present
-    check_arg_name(args)
-
-    # Logistics #
+def initial_setup(args, cl_args):
+    """
+    Sets up email hook, setting up seed, and cuda settings.
```
sleepinyourhat: Restructure: sets up ... setting up doesn't really make sense.

sleepinyourhat: Add a newline, too.

main.py
```diff
@@ -53,7 +53,136 @@ def handle_arguments(cl_arguments):
...
+    ----------------
+    args: Params object
+    pretrain_tasks: list of pretraining Task objects
+    target_tasks: list of target task training Task objects
+
+    Returns
+    ----------------
+    None
+    """
+    log.info("Will run the following steps:")
```
sleepinyourhat: This should probably go below all the warnings, right?

main.py
```diff
@@ -179,72 +313,42 @@ def main(cl_arguments):
...
             args.cuda = -1
 
-    # Prepare data #
+    return args, seed
+
+def main(cl_arguments):
+    ''' Train or load a model. Evaluate on some tasks. '''
+    cl_args = handle_arguments(cl_arguments)
+    args = config.params_from_file(cl_args.config_file, cl_args.overrides)
+    # Raise error if obsolete arg names are present
```
sleepinyourhat: Restore the newline that was above this comment.

main.py
```diff
@@ -179,72 +313,42 @@ def main(cl_arguments):
                 "GPU access failed. You might be using a CPU-only installation of PyTorch. Falling back to CPU.")
             args.cuda = -1
 
-    # Prepare data #
+    return args, seed
+
+def main(cl_arguments):
```
sleepinyourhat: Try to do a bit of cleanup to the comments that define the major sections of main—they don't seem especially consistent.

main.py
```diff
@@ -260,49 +355,17 @@ def main(cl_arguments):
                           to_train, opt_params, schd_params,
                           args.shared_optimizer, args.load_model, phase="main")
 
-    # Select model checkpoint from main training run to load
+    # For checkpointing logic 
     if not args.do_target_task_training:
         log.info("In strict mode because do_target_task_training is off. "
                  "Will crash if any tasks are missing from the checkpoint.")
         strict = True
```
pruksmhc: @W4ngatang  why would we want strict to equal False if target_Task_Training = 0?

main.py
W4ngatang: I think that's right but I just find it annoying and not clear why it's the case (@sleepinyourhat do you remember why we require this?).

main.py
W4ngatang: Oh sorry! I misread the code down below I think.

main.py
W4ngatang: Hmm, I didn't author this (I think Sam did?), but I would guess it's because there's not really a good reason to evaluate an untrained target task classifier. More often than not, it's probably a configuration error?

main.py
W4ngatang: done

main.py
```diff
@@ -154,17 +282,14 @@ def main(cl_arguments):
...
     log.info("Parsed args: \n%s", args)
 
     config_file = os.path.join(args.run_dir, "params.conf")
     config.write_params(args, config_file)
     log.info("Saved config to %s", config_file)
 
-    seed = random.randint(1, 10000) if args.random_seed < 0 else args.random_seed
-    random.seed(seed)
-    torch.manual_seed(seed)
     log.info("Using random seed %d", seed)
```
W4ngatang: how does this not error out? Seed seems to be no longer defined here and not a global?

pruksmhc: Oh whoops that's weird - seems like Git didn't update all the changes here

pruksmhc: figuring out what's going on

main.py
```diff
@@ -53,7 +53,127 @@ def handle_arguments(cl_arguments):
...
+
+    if args.do_pretrain:
+        assert_for_log(args.pretrain_tasks != "none",
+                       "Error: Must specify at least on training task: [%s]" % args.pretrain_tasks)
+        assert_for_log(
+            args.val_interval %
+            args.bpp_base == 0, "Error: val_interval [%d] must be divisible by bpp_base [%d]" %
+            (args.val_interval, args.bpp_base))
+        steps_log.write("Training model on tasks: %s \n" % args.pretrain_tasks)
+    import pdb; pdb.set_trace()
```
W4ngatang: delete

main.py
```diff
@@ -179,74 +309,38 @@ def main(cl_arguments):
...
             args.cuda = -1
 
-    # Prepare data #
+    return args, seed
+
+def main(cl_arguments):
+    ''' Train a model for multitask-training.'''
+    cl_args = handle_arguments(cl_arguments)
+    args = config.params_from_file(cl_args.config_file, cl_args.overrides)
+    # Check for depricated arg names
```
W4ngatang: deprecated

main.py
pruksmhc: I'll make an issue for this

main.py
sleepinyourhat: That sounds right...

pruksmhc: In that case, we should have an assert for this (if it's a configuration error) rather than letting this pass by (even if there are a bunch of warning and exclamation points in the warning)

main.py
```diff
@@ -53,7 +53,127 @@ def handle_arguments(cl_arguments):
...
+    target_tasks: list of target task training Task objects
+
+    Returns
+    ----------------
+    None
+    """
+    steps_log = io.StringIO()
+    if any([t.val_metric_decreases for t in pretrain_tasks]) and any(
+            [not t.val_metric_decreases for t in pretrain_tasks]):
+        log.warn("\tMixing training tasks with i=ncreasing and decreasing val metrics!")
```
iftenney: typo?

main.py
pruksmhc: Made a new issue

main.py
```diff
@@ -53,7 +53,127 @@ def handle_arguments(cl_arguments):
...
+        best_path = get_best_checkpoint_path(args.run_dir)
+        if best_path:
+            load_model_state(model, best_path, args.cuda, task_names_to_avoid_loading,
+                             strict=strict)
+        else:
+            assert_for_log(args.allow_untrained_encoder_parameters,
+                           "No best checkpoint found to evaluate.")
+            log.warning("Evaluating untrained encoder parameters!")
+    return task_names_to_avoid_loading
+
```
sleepinyourhat: Run autopep8. There may need to be another line break here.

main.py
```diff
@@ -53,7 +53,127 @@ def handle_arguments(cl_arguments):
...
+    assert_for_log(args.transfer_paradigm in ["finetune", "frozen"],
+                   "Transfer paradigm %s not supported!" % args.transfer_paradigm)
+
+    if args.do_pretrain:
+        assert_for_log(args.pretrain_tasks != "none",
+                       "Error: Must specify at least on training task: [%s]" % args.pretrain_tasks)
+        assert_for_log(
+            args.val_interval %
+            args.bpp_base == 0, "Error: val_interval [%d] must be divisible by bpp_base [%d]" %
```
sleepinyourhat: Be careful when merging. We already merged in Shikha's PR getting rid of this.

main.py
sleepinyourhat: Mind being consistent about whitespace before/after the comments that introduce new sections of the code?

W4ngatang:
Can you merge master into this? It's going to un-do some fixes we just added.

main.py
W4ngatang: Can you make sure to merge master into this? This was fixed in another earlier PR.

Issue #89, opened 2018-07-03
Include timestamps in log files


sleepinyourhat:
Useful for tracking/predicting the run times of jobs.

PR #577, opened 2019-04-26
Add timestamps to log file

W4ngatang:
^ The timestamps are formatted the same as in the printed messages. Fixes #577 and #89 .


sleepinyourhat:
Mind checking if this breaks extract_results.py and extract_diagnostic_results.py? If it does, feel free to delete/deprecate them if you think that's appropriate.

main.py
```diff
@@ -5,10 +5,12 @@
     python -m ipdb main.py --config_file ...
 """
 # pylint: disable=no-member
+import logging as log
+log.basicConfig(format="%(asctime)s: %(message)s", datefmt="%m/%d %I:%M:%S %p", level=log.INFO)  # noqa # nopep8
```
sleepinyourhat: This import order thing was broken by Black.

Issue #566, opened 2019-04-24
Delete decoder code


sleepinyourhat:
It seems to be abandoned by its authors, and there are major known issues (#371, #375

phu-pmh:
The decoder is still used in `src\models.py` for `RedditSeq2SeqTask, Wiki103Seq2SeqTask` and `MTTask` (related to ACL submission?). It might still be needed.

sleepinyourhat:
Still worth taking a look around to see if there's anywhere that we can/should add deprecation warnings.

sleepinyourhat:
#781

PR #579, opened 2019-04-26
decoder issue #566

phu-pmh:
deprecation warning added for seq2seq_decoder


src/modules/seq2seq_decoder.py
```diff
@@ -23,6 +23,7 @@
 
 from .modules import Pooler
 
+import warnings
```
sleepinyourhat: Could you switch to log.warning? It's what we use elsewhere, and it should work more straightforwardly with our logging tools. Example:

https://github.com/nyu-mll/jiant/blob/master/main.py#L178

phu-pmh: fixed!

Issue #523, opened 2019-03-14
Remove or hide fasttext model compatibility


sleepinyourhat:
The fasttext _model component_ (as opposed to the mere embeddings file) isn't compatible with PyTorch out of the box. If you try to install it normally with conda, you'll cause jiant/PyTorch to segfault on import. That's what @alexwarstadt was hitting a couple of days ago.

It's possible to use it if you do a more complex custom installation, but it's very easy to mess up, and I don't expect that many people will use it. We should definitely remove any mention the fasttext model component from the readme/requirements, and it might be a good idea to remove support for it entirely.

W4ngatang:
Removed with #583

PR #583, opened 2019-04-26
Remove fasttext model

W4ngatang:
Remove fasttext model support, as per #523 . Delete relevant code and update documentation (removing paths and config options).

### DIFF
--- a/src/preprocess.py
+++ b/src/preprocess.py
@@ -24,11 +24,6 @@ from allennlp.data.token_indexers import \
     SingleIdTokenIndexer, ELMoTokenCharactersIndexer, \
     TokenCharactersIndexer
 
-try:
-    import fastText
-except BaseException:
-    log.info("fastText library not found!")
-
 import _pickle as pkl  # :(
 
 from .utils import config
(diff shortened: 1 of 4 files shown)


src/preprocess.py
```diff
@@ -191,18 +186,27 @@ def _find_cached_file(exp_dir: str, global_exp_cache_dir: str,
...
+
+    # Load all the word embeddings based on vocabulary
     log.info("\tBuilding embeddings from scratch")
-    if args.fastText:
-        word_embs, _ = get_fastText_model(vocab, args.d_word,
-                                          model_file=args.fastText_model_file)
-        log.info("\tUsing fastText; no pickling of embeddings.")
-        return word_embs
-
-    word_embs = get_embeddings(vocab, args.word_embs_file, args.d_word)
```
sleepinyourhat: This seemed like a nice helper method. Why get rid of it?

W4ngatang: After getting rid of the optional loading of the fastText model, it was just a wrapper around `get_embeddings` without any extra functionality.

Issue #580, opened 2019-04-26
Rename sent_embs variable


pruksmhc:
This is used elsewhere (in src/models.py) - should probably be changed elsewhere as well (separate PR). The thought behind it is that sent_embs is output of the sentence encoders, but in actually it should be renamed to word_embs_in_context (or something related).

PR #585, opened 2019-04-26
Cleanup: Rename sent_embs #580

phu-pmh:
remaned sent_embs to word_embs_in_context


phu-pmh:
Tested! demo.conf is running fine.

Issue #581, opened 2019-04-26
Support update_metrics for PairOrdinalRegressionTask


pruksmhc:
It's currently not supported after the change in how we handle metrics/trainer refactor.

PR #586, opened 2019-04-26
Add update_metrics to PairOrdinalRegressionTask

HaokunLiu:
This also clean up the _pair_sentence_forward function


HaokunLiu:
#581

sleepinyourhat:
@pruksmhc or @W4ngatang should review this—I'm not as familiar with the metrics code.

src/tasks/tasks.py
```diff
@@ -359,9 +360,9 @@ def process_split(self, split, indexers) -> Iterable[Type[Instance]]:
         ''' Process split text into a list of AllenNLP Instances. '''
         return process_single_pair_task_split(split, indexers, is_pair=True,
                                               classification=False)
-    def update_metrics():
-        # currently don't support metrics for regression task
-        # TODO(Yada): support them! 
+    def update_metrics(self, logits, labels, tagmask=None):
+        self.scorer1(mean_squared_error(logits, labels))
```
pruksmhc: Add a comment here that this is average MSE (which is why you add a MSE here)

HaokunLiu: Done. Actually there is also similar comment in the __init__ of this task.

HaokunLiu:
Once more, this branch is ready.

PR #588, opened 2019-04-26
Converted rest of eval -> target train

pruksmhc:
Went through all references of eval and converted the ones that should be target train. 

Tested on demo.conf, although the naming change shouldn't change anything functionality related.

### DIFF
--- a/scripts/update_config.py
+++ b/scripts/update_config.py
@@ -8,11 +8,17 @@ from src.utils import config # use symlink from scripts to src
 
 # Mapping - key: old name, value: new name
 name_dict = {'task_patience':'lr_patience',
-               'do_train': 'do_pretrain',
-               'train_for_eval':'do_target_task_training',
-               'do_eval': 'do_full_eval',
-               'train_tasks':'pretrain_tasks',
-               'eval_tasks':'target_tasks'}
+				'do_train': 'do_pretrain',
+				'train_for_eval':'do_target_task_training',
+				'do_eval': 'do_full_eval',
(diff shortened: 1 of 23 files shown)


scripts/update_config.py
```diff
@@ -8,11 +8,17 @@
...
+				'train_for_eval':'do_target_task_training',
+				'do_eval': 'do_full_eval',
+				'train_tasks':'pretrain_tasks',
+				'eval_tasks':'target_tasks', 
+				'eval_data_fraction': 'target_train_data_fraction', 
+				'eval_val_interval': 'target_train_val_interval',
+				'eval_max_vals': 'target_train_max_vals',
+				'load_eval_checkpoint': 'load_target_train_checkpoint', 
+				'eval_data_fraction': 'target_train_data_fraction'
+				}
```
sleepinyourhat: Odd indentation...

src/trainer.py
```diff
@@ -995,7 +995,7 @@ def _save_checkpoint(
         training_state: An object containing trainer state (step number, etc.), to be saved.
         phase: Usually 'pretrain' or 'target_train'.
         new_best_macro: If true, the saved checkpoint will be marked with .best_macro, and
-            potentially used later when switching from main to eval training.
+            potentially used later when switching from pretrain to target training.
```
sleepinyourhat: => "pretraining to target task training"

scripts/update_config.py
pruksmhc: Git seems to be displaying the indentation incorrectly here....

PR #591, opened 2019-04-26
Add CommitmentBank task

W4ngatang:
A SuperGLUE task. Data forthcoming. Also exposes quote level as an option for `load_data`.


src/tasks/tasks.py
```diff
@@ -1618,3 +1618,54 @@ def load_data(self, path, max_seq_len):
         self.val_data_text = list(val_data) + [masks[1]]
         self.test_data_text = list(te_data[:2]) + [te_targs] + [te_mask]
         log.info('\tFinished loading CCGTagging data.')
+
+@register_task('commitbank', rel_path='CommitmentBank/')
```
sleepinyourhat: Keep just one of these four or else add much clearer comments. Otherwise, this could throw off SuperGLUE users.

src/tasks/tasks.py
```diff
@@ -1618,3 +1618,54 @@ def load_data(self, path, max_seq_len):
...
+@register_task('commitbank-balanced', rel_path='CommitmentBank_balanced/')
+# all has all the data as the test set, meant for evaluating w/ MNLI
+@register_task('commitbank-balanced-all', rel_path='CommitmentBank_balanced_all/')
+@register_task('commitbank-nice', rel_path='CommitmentBank_nice/')
+class CommitmentTask(PairClassificationTask):
+    ''' NLI-formatted task detecting speaker commitment. '''
```
sleepinyourhat: Add a link to the paper/site.

src/tasks/tasks.py
```diff
@@ -1618,3 +1618,54 @@ def load_data(self, path, max_seq_len):
...
+
+    def __init__(self, path, max_seq_len, name, **kw):
+        ''' There are 1363 supertags in CCGBank. '''
```
sleepinyourhat: ?

src/tasks/tasks.py
```diff
@@ -1618,3 +1618,54 @@ def load_data(self, path, max_seq_len):
...
+        super().__init__(name, n_classes=3, **kw)
+        self.scorer2 = F1Measure(0)
```
sleepinyourhat: Say why we need three.

src/tasks/tasks.py
```diff
@@ -1618,3 +1618,54 @@ def load_data(self, path, max_seq_len):
...
+                           skip_rows=1, delimiter=',', quote_level=csv.QUOTE_ALL)
+        self.train_data_text = tr_data
+        self.val_data_text = val_data
+        self.test_data_text = te_data
+        log.info('\tFinished loading CCGTagging data.')
+
+    def get_metrics(self, reset=False):
+        '''Get metrics specific to the task'''
+        acc = self.scorer1.get_metric(reset)
+        pcs1, rcl1, f11 = self.scorer2.get_metric(reset)
```
sleepinyourhat: Add a comment defining the metrics somewhere near here.

PR #593, opened 2019-04-26
Rename `src` to `jiant`

W4ngatang:
Rename `src` to jiant, and accordingly do the following:

- switch import from `src` in `main.py` to `jiant`

- remove relative imports that go outside the current directory to absolute imports using `jiant`.


zphang:
What is the rationale for: 

>remove relative imports that go outside the current directory to absolute imports using jiant.

jiant/utils/config.py
```diff
@@ -1,17 +1,18 @@
...
+import json
 import types
+import random
+import pyhocon
+import argparse
 import logging as log
+
 log.basicConfig(format='%(asctime)s: %(message)s',
                 datefmt='%m/%d %I:%M:%S %p', level=log.INFO)
+from . import hocon_writer
```
iftenney: should this be `from jiant.utils import hocon_writer` ?

W4ngatang: I didn't change relative imports that were really deep in the repo and importing from the same directory. You think it's better to switch everything to absolute imports?

W4ngatang:
> 
> 
> What is the rationale for:
> 
> > remove relative imports that go outside the current directory to absolute imports using jiant.

@zphang mainly to avoid having really long import lines when it seems likely that those files will stay in the same folder. Do you think it's better to stick to absolute imports always?

sleepinyourhat:
As soon as you merge this, make sure to fix all mentions in .sh scripts, readmes, the site, and the tutorials.

jiant/utils/config.py
iftenney: Yes, for consistency and less brittle if we move anything.

sleepinyourhat:
And fix broken tests. Looks like this probably won't make it into 0.9, though if you're dead sure everything works and you fix the tests, feel free to merge.

sleepinyourhat:
LGTM

iftenney:
Can you fix the `probing/src` symlink so that the scripts in that directory will still work? (and anywhere else something similar is used)

iftenney:
(otherwise, LGTM)

Issue #337, opened 2018-08-07
clean up/remove references to train_dummy and test_dummy files in NLI Probing Task 


epavlick:

sleepinyourhat:
Still on this? If not, close or unassign.

PR #594, opened 2019-04-26
relocate nps and cleanup nli-prob

HaokunLiu:
#337 
It appears, these tasks are probing tasks like mnli-diagnostic. They don't actually have train/test data and are only used alongside with eval_existing.conf. So I just did what people did in mnli-diagnostic.
Since @W4ngatang said that tasks.py aren't expected to contain much project-specific code, I moved nps to nli-prob as well.


src/tasks/nli_probing.py
```diff
@@ -7,123 +7,95 @@
...
+    def load_data(self, path, max_seq_len):
+        targ_map = {'neutral': 0, 'entailment': 1, 'contradiction': 2}
+        prob_data = load_tsv(self._tokenizer_name, os.path.join(path, 'dev.tsv'), max_seq_len,
+                            s1_idx=0, s2_idx=1, label_idx=2, label_fn=targ_map.__getitem__, skip_rows=0)
+        self.train_data_text = self.val_data_text = self.test_data_text = prob_data
+        log.info("\tFinished loading NP/S data.")
+
 
 @register_task('nli-prob', rel_path='NLI-Prob/')
 class NLITypeProbingTask(PairClassificationTask):
```
sleepinyourhat: Why all the structural changes here?

sleepinyourhat: This could be odd diff rendering, but I find it very hard to follow this...

sleepinyourhat: The move makes sense—I could just use more help figuring out what the cleanup is.

src/tasks/nli_probing.py
```diff
@@ -7,123 +7,95 @@
...
         self.load_data(path, max_seq_len, probe_path)
-        #  self.use_classifier = 'mnli'  # use .conf params instead
-        self.sentences = self.train_data_text[0] + self.train_data_text[1] + \
-            self.val_data_text[0] + self.val_data_text[1]
+        self.sentences = self.val_data_text[0] + self.val_data_text[1]
 
     def load_data(self, path, max_seq_len, probe_path):
         targ_map = {'neutral': 0, 'entailment': 1, 'contradiction': 2}
-        label_fn = targ_map.__getitem__
-        tr_data = load_tsv(data_file=os.path.join(path, 'train_dummy.tsv'), max_seq_len=max_seq_len,
```
HaokunLiu: train_dummy.tsv and test_dummy.tsv are dummy. Since the tasks are only used with pretrain = 0, target_tasks_train = 0, full_eval = 1 and write_pred = val, the only split used in the program is the self.val_data_text. (this is similar to mnli-diagnostic)
What I did is just like mnli-diagnostic, pointing all splits to the only data that exists.

src/tasks/nli_probing.py
```diff
@@ -7,123 +7,95 @@
...
-        self.val_data_text = val_data
-        self.test_data_text = te_data
+        self.train_data_text = self.val_data_text = self.test_data_text = prob_data
         log.info("\tFinished loading NLI-type probing data.")
 
 
 @register_task('nli-prob-negation', rel_path='NLI-Prob/')
 class NLITypeProbingTaskNeg(PairClassificationTask):
 
-    def __init__(self, path, max_seq_len, name, probe_path="probe_dummy.tsv",
```
HaokunLiu: probe_path is not used in this task.

src/tasks/nli_probing.py
```diff
@@ -7,123 +7,95 @@
...
 
-        self.train_data_text = tr_data
-        self.val_data_text = val_data
-        self.test_data_text = te_data
+        self.train_data_text = self.val_data_text = self.test_data_text = prob_data
         log.info("\tFinished loading preposition swap data.")
 
 
 @register_task('nli-alt', rel_path='NLI-Prob/')
-class NLITypeProbingAltTask(NLITypeProbingTask):
```
HaokunLiu: The author declared this to be a subclass of NLITypeProbingTask, but it didn't really inherit anything. Nor is this subclass relationship used anywhere where (like in an if-clause) in jiant. 
So I thought, better keep the conformity.

src/tasks/nli_probing.py
HaokunLiu: There are five tasks here, they are of the same kind. So I did this five times.

sleepinyourhat: Looks good!

HaokunLiu:
Do we still have the data used in final_nliprobing_runs.sh

sleepinyourhat:
@najoungkim would know. Do we need them for this change?

HaokunLiu:
Yes, if we need to confirm we can evaluate them.

sleepinyourhat:
If you don't hear back right away, mind just putting in some dummy data to test it? It should be easy to reverse-engineer from the specification.

HaokunLiu:
Hmm, that will do.

najoungkim:
Is it nps that you need? I think Tom has that data

HaokunLiu:
Never mind. If it is not conveniently available, we could use dummy data instead.

HaokunLiu:
Tested

Issue #491, opened 2019-02-18
Replace OpenAI Transformer implementation with newer version


iftenney:
Latest version of [pytorch_pretrained_bert](https://github.com/huggingface/pytorch-pretrained-BERT#pytorch-pretrained-bert-the-big-and-extending-repository-of-pre-trained-transformers) package contains an implementation of the OpenAI GPT model. This package is much cleaner than the earlier one we were using, so we might save a lot of code by switching our support for the GPT model to this implementation.

See https://github.com/huggingface/pytorch-pretrained-BERT/releases/tag/v0.5.0

iftenney:
Note: let's not do this until after the ACL deadline (March 4).

pruksmhc:
Closed with #611

PR #611, opened 2019-04-27
replace OpenAI with hugging face version #491

phu-pmh:
Replaced old Tensorflow version with hugging face's version. Tested using `config/spring19_seminar/cola_gpt.conf`, and it ran fine.

- Note: Need to download pre-trained model folder [here](https://github.com/openai/finetune-transformer-lm/tree/master/model) and place it under `src/openai_transformer_lm/pytorch_huggingface/` folder. (Should I add this to README?)


phu-pmh:
Added info on open-ai setup in README.

sleepinyourhat:
Taking a look...

README.md
```diff
@@ -156,7 +156,7 @@ To use the PRPN sentence encoder from [***Neural language modeling by jointly le
 
 We also include an experimental option to use a shared [Transformer](https://arxiv.org/abs/1706.03762) in place of the shared BiLSTM by setting ``sent_enc = transformer``. When using a Transformer, we use the [Noam learning rate scheduler](https://github.com/allenai/allennlp/blob/master/allennlp/training/learning_rate_schedulers.py#L84), as that seems important to training the Transformer thoroughly. 
 
-We also support using pretrained Transformer language models. To use the OpenAI transformer model, set `openai_transformer = 1`.
+We also support using pretrained Transformer language models. To use the OpenAI transformer model, set `openai_transformer = 1`, download the [model](https://github.com/openai/finetune-transformer-lm) folder that contains pre-trained models, and place it under `src/openai_transformer_lm/pytorch_huggingface/`.
```
sleepinyourhat: Why do we need people to manually download these? The huggingface code seems like it should be able to automatically download any needed weights (which is how we use BERT).

phu-pmh:
> I'm surprised you needed to change so little. Is everything in defaults.conf correct?

Yes, I went through the codes yesterday. Huggingface uses the same module names and function names as OpenAI so I only had to change the import statements. `defaults.conf` is correct.

README.md
phu-pmh: Huggingface does not automatically download the OpenAI models. They ask to manually download the OpenAI models in their README.

sleepinyourhat: Got it, thanks! The readme makes it sound possible, but looking at the code, I don't see any URL.

sleepinyourhat:
Run a short test, maybe with demo.conf (with GPT turned on) and make sure you get the same better-than-chance results on some task with old GPT and new GPT. Then we can merge.

phu-pmh:
> Run a short test, maybe with demo.conf (with GPT turned on) and make sure you get the same better-than-chance results on some task with old GPT and new GPT. Then we can merge.

I'll do that!

PR #615, opened 2019-04-29
Get rid of the NFS vs JIANT environment variable distinction

sleepinyourhat:
This distinction is confusing in most cases, and it's still easy to recreate it with custom paths/configs.

Resolves #613 

@iftenney - Mind reviewing?

PR #616, opened 2019-04-29
Move more scripts to scripts/

sleepinyourhat:
@iftenney, mind taking a look?


sleepinyourhat:
Resolves #606.

demo.with_docker.sh
```diff
@@ -40,7 +40,6 @@ COMMAND+=( -o "exp_name=jiant-demo" )
 # Run demo.conf in the docker container.
 sudo docker run --runtime=nvidia --rm -v "$TEMP_DIR:/nfs/jsalt" \
   -v "$JIANT_PATH:/share/jiant" \
-  -e "NFS_PROJECT_PREFIX=/nfs/jsalt/exp" \
```
iftenney: Does this assume #615 ?

scripts/demo.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-# Quick-start: run this
+# Quick-start: run this from the base jiant/ directory
```
iftenney: Optional, but I like to add a `pushd` to automatically jump to this directory - just smooths out the experience a bit and avoids annoying path errors.

Something like:
```
pushd $(dirname $0)
pushd $(git rev-parse --show-toplevel)
```

scripts/demo.with_docker.sh
```diff
@@ -0,0 +1,49 @@
...
+#
+# TODO to replace this with a properly-agnostic mount point inside the
+# container.
+#
+# TODO: turn this into a real integration test...
+
+set -eux
+
+# Get the path to this repo; we'll mount it from the container later.
+JIANT_PATH=$(readlink -f $(dirname $0))
```
iftenney: need to change this so it still points to the top-level jiant/ repo. Something like:
```
# temporarily cd to jiant/
pushd $(dirname $0)
pushd $(git rev-parse --show-toplevel)

# get absolute path (I think this works)
JIANT_PATH=$(readlink -f .)
```

demo.with_docker.sh
sleepinyourhat: Yes, good catch. I'll revert that here.

sleepinyourhat:
Take another look—demo.sh works, but I have no easy access to a docker environment to test your proposed solution there.

PR #618, opened 2019-04-30
Created new branch for fixing  data fractioning issue

BordiaS:
Will delete the other PR once this is accepted.


sleepinyourhat:
@pep8speaks pep8ify

src/preprocess.py
```diff
@@ -366,64 +364,24 @@ def build_tasks(args):
             task.name, "test", preproc_dir)
         # When using pretrain_data_fraction, we need modified iterators for use
         # only on training datasets at pretraining time.
-        if args.pretrain_data_fraction < 1 and task.name in train_task_names:
```
sleepinyourhat: Just checking: This was simply broken before, right? (Since train_task_names doesn't exist.)

BordiaS: Yes, it was broken. `train_task_names` should be `pretrain_task_names` unless I am missing something.

BordiaS:
> This looks basically reasonable. Two comments:
> 
> * This is touching really important code that I don't fully understand, so it'd be best if @pruksmhc or @W4ngatang could give it a second look.
> * Could you add a comment in defaults.conf and here explaining what happens if you set both settings to <1. In particular, if you set both to 0.1, will you see the same training set both times, will you get two disjoint samples, or will you get two overlapping samples?

In that case, the training set will be same. Added the explanation for data fractions in default.conf.

config/defaults.conf
```diff
@@ -87,8 +87,8 @@ reload_indexing = 0  // If true, force the rebuilding of the index files in prep
                      // exist.
 target_train_data_fraction = 1  // Use only the specified fraction of the training data in the do_target_task_training phase.
                         // should not impact pretraining, even for the same task.
-                        // pretrain_data_fraction should not be used together with                 // target_train_data_fraction
-                        // both should not be less then 1 at a same time
+                        // pretrain_data_fraction can be used together with target_train_data_fraction. Both the fractions can be less then 1 at a same time.
```
sleepinyourhat: The second half of this line is confusing—just delete it. Also, try to keep lines under 100 chars.

config/defaults.conf
```diff
@@ -87,8 +87,8 @@ reload_indexing = 0  // If true, force the rebuilding of the index files in prep
...
+                        // The data is sorted and serialized.  For e.g., if both are set to 0.1, the training set will be same for both.
```
sleepinyourhat: Say more—this might give the impression that it's sorted by text or by length, which would be a serious problem.

PR #626, opened 2019-04-30
Superglue RTE

W4ngatang:
RTE is staying in SuperGLUE, but due to format changes, we still need to make changes.

This moves the old RTE data loader / prediction writer to `rte-glue` and updates task `rte` to expect the JSONL format of SuperGLUE.


src/evaluate.py
```diff
@@ -163,12 +166,13 @@ def write_preds(tasks: Iterable[tasks_module.Task], all_preds, pred_dir, split_n
                  'mrpc': 'MRPC',
                  'qnli': 'QNLI',
                  'qqp': 'QQP',
-                 'rte': 'RTE',
+                 'rte-glue': 'RTE',
```
sleepinyourhat: It seems like it'd be more robust to keep the 'rte' object name for the GLUE version and use super-rte for the SuperGLUE version. No

W4ngatang: Cut off? I'm fine either way.

src/evaluate.py
```diff
@@ -10,7 +10,7 @@
 import torch
 from allennlp.data.iterators import BasicIterator
 from . import tasks as tasks_module
-from .tasks.tasks import CommitmentTask, WiCTask
+from .tasks.tasks import CommitmentTask, RTESuperGLUETask, WiCTask
```
sleepinyourhat: Optional suggestion: SuperGLUERTETask

src/tasks/tasks.py
```diff
@@ -1274,7 +1274,38 @@ def load_data(self, path, max_seq_len):
         self.train_data_text = tr_data
         self.val_data_text = val_data
         self.test_data_text = te_data
-        log.info("\tFinished loading RTE.")
+        log.info("\tFinished loading RTE (from GLUE formatted data).")
+
+
+@register_task('rte-superglue', rel_path='RTE/')
+class RTESuperGLUETask(RTETask):
+    ''' Task class for Recognizing Textual Entailment 1, 2, 3, 5 '''
```
pruksmhc: nit: "3, and 5"

PR #628, opened 2019-05-01
Add pre-commit for code style hooks

Iwontbecreative:
Standardizes the code to follow black style with line length 100 (same as what was used before AFAICT).


Iwontbecreative:
Ideally we should mention somewhere that people should run `pre-commit install` prior to coding to install the git hooks, but not sure where that should be since there's no `CONTRIBUTING.MD`

sleepinyourhat:
Initial thoughts:
- Do you know why pep8speaks is still reporting some unresolved issues?
- For now, just add a note to README.md. @pruksmhc is doing some reorganization and tutorial building.

Iwontbecreative:
My understanding from looking at a few of the cases above:
* Line too long: a lot of times this is because of urls where black doesn't split (which is the right behavior). When it goes over by a small amount (<20 chars) sometimes it seems to just fail to split. Does not seem to be that many exceptions considering the whole codebase changed. 
* For the whitespace before ":", this is apparently not actually a pep8 infraction https://github.com/ambv/black (see slices). Black will have `list[i:j]` and  `list[i + 1 : i + 3]` whereas autopep8 seems to want `list[i:j]` and `list[i+1:i+3]`
* For the trailing whitespace, this is not a check that black runs. I added a pre-commit hook to add it as an extra check

sleepinyourhat:
Thanks! Mind tweaking the pep8speaks config to turn off warnings about things that we're not fixing. (pep8speaks.yml, IIRC.) Or, if you're confident this is a robust strategy, just turning off pep8speaks.

Iwontbecreative:
It's robust if installed, but it doesn't enforce new contributors actually run `pre-commit install` so checking on github is likely still required for MRs. Maybe black should be integrated as part of the CI but meanwhile I think pep8speaks might make sense.

I will just turn off E203, the line too long warnings are legitimate and so are the whitespace ones.

sleepinyourhat:
Cool. Can give a proper review once the readme is set up.

Iwontbecreative:
Test to see whether the ":" are ignored
@pep8speaks suggest diff

Iwontbecreative:
@pep8speaks suggest dif

Iwontbecreative:
@pep8speaks suggest diff

Iwontbecreative:
Missed the comment on README, updated to mention it.

Iwontbecreative:
@pruksmhc Hey!
Do you know whether pep8speaks updates config if updates to it are after initial commit? It seems it is still throwing E203 errors even though I added those to the ignore list

sleepinyourhat:
This works for me. It looks like pep8speaks re-ran, but it may not have re-read the yml. I remember seeing that deleting the pep8speaks comment and making another commit can help.

Iwontbecreative:
Updated to fix the merge conflict but by the nature of the commit it will likely create many others.

sleepinyourhat:
Aaand now there are some more. Mind fixing those and messaging me/Alex/Yada to merge as soon as you're done?

Iwontbecreative:
How should we deal with this MR @sleepinyourhat given merge conflicts will appear for each merge in master? Should we quick merge it at one point and let others fix the conflicts on their end?

The other solution:
We only merge the pre-commit but don't blackify the files yet. They will progressively be fixed as people touch parts of the code. Think this is worse

On a related note, `black` now has PSF support which bodes well for its evolution: https://twitter.com/llanga/status/1123980466292445190?s=20

PR #629, opened 2019-05-01
Continuous Integration

zphang:
- Mild clean-up of tests
- Added CircleCI configuration

Will work with @pruksmhc to integrate into repository with GitHub permissions.


zphang:
- CircleCI is now configured to run on every PR (jiant, or forked), as well as every jiant branch. Ideally we would want to add additional filters (e.g. build only when approved), but this should serve as a good testing phase. We might quickly burn through our free minutes for now, though.
- The status of the builds can be seen [here](https://circleci.com/gh/nyu-mll/jiant), or also e.g. on the PR page, with little icons at the end of the PR names.
- The badge on the README.md points to the master branch (not the PR branch). It says Fail because the master branch has not yet merged in the configuration for running CI.
- Likewise, all branches that have not merged in the changes from this branch will fail, as they will not have the CircleCI config files.
- This PR **is ready to be merged**. The exact CI workflow can still undergo some optimization, but this is ready to show if the master branch is currently passing tests. 
- Currently this just runs tests, not style checks.

PR #634, opened 2019-05-01
WiC

W4ngatang:
Add Words in Context task, including:

- extend pair classification to optionally extract and include specific representations of tokens
- task definition and prediction code


src/models.py
```diff
@@ -653,14 +653,16 @@ def build_pair_attn(d_in, d_hid_attn):
...
+            d_out *= 3 # also pass the two contextual word representations
         classifier = Classifier.from_params(d_out, n_classes, params)
         module = SingleClassifier(pooler, classifier)
     else:
+        d_out = d_out + d_inp if isinstance(task, WiCTask) else d_out
         classifier = Classifier.from_params(4 * d_out, n_classes, params)
         module = PairClassifier(pooler, classifier, pair_attn)
     return module
 
-
```
sleepinyourhat: autopep8

src/models.py
```diff
@@ -904,11 +906,17 @@ def _pair_sentence_forward(self, batch, task, predict):
         classifier = self._get_classifier(task)
         if self.use_bert:
             sent, mask = self.sent_encoder(batch['inputs'], task)
-            logits = classifier(sent, mask)
+            if isinstance(task, WiCTask):
```
sleepinyourhat: Ideally, we should just create different fwd functions for different tasks to minimize the number of places we need these special cases. This is totally fine/standard for now, though, but try to add more comments explaining why each of these special cases is necessary.

src/modules/modules.py
```diff
@@ -411,28 +411,70 @@ def __init__(self, pooler, classifier):
         self.pooler = pooler
         self.classifier = classifier
 
-    def forward(self, sent, mask):
+    def forward(self, sent, mask, idxs=[]):
+        """ Assumes batch_size x seq_len x d_emb """
         emb = self.pooler(sent, mask)
-        logits = self.classifier(emb)
+
+        # append any additional representations we want
```
sleepinyourhat: Like what?

src/modules/modules.py
```diff
@@ -411,28 +411,70 @@ def __init__(self, pooler, classifier):
...
+        logits = self.classifier(final_emb)
         return logits
 
 
 class PairClassifier(nn.Module):
-    ''' Thin wrapper around a set of modules. For sentence pair classification. '''
+    ''' Thin wrapper around a set of modules.
+    For sentence pair classification.
+    Pooler specifies how to aggregate inputted sequence of vectors.
+    Also allows for additional token representations to be used via idx{1,2}'''
```
sleepinyourhat: What does 'additional token representations' mean?

src/models.py
W4ngatang: Yea I agree that having these special cases aren't great long term :/ I'll add a comment explaining for now.

PR #636, opened 2019-05-01
Add COPA and SWAG

W4ngatang:
Add implementation of COPA and SWAG tasks


src/evaluate.py
```diff
@@ -140,18 +141,21 @@ def write_preds(tasks: Iterable[tasks_module.Task], all_preds, pred_dir, split_n
...
             _write_commitment_preds(task, preds_df, pred_dir, split_name,
                                     strict_glue_format=strict_glue_format)
+        elif isinstance(task, COPATask):
+            _write_copa_preds(task, preds_df, pred_dir, split_name,
+                              strict_glue_format=strict_glue_format)
         else:
             log.warning("Task '%s' not supported by write_preds().",
                         task.name)
             continue
+        log.info("Task '%s': Wrote predictions to %s", task.name, pred_dir)
```
sleepinyourhat: It looks like this'll appear twice now for Edge Probing, no?

src/models.py
```diff
@@ -1110,6 +1122,40 @@ def _lm_forward(self, batch, task, predict):
             pass
         return out
 
+    def _mc_forward(self, batch, task, predict):
+        ''' Forward for a multiple choice question answering task '''
```
sleepinyourhat: How does this handle the context vs. question distinction? Are they just concatenated?

src/tasks/tasks.py
```diff
@@ -2224,3 +2225,152 @@ def get_metrics(self, reset=False):
...
+
+        def _load_split(data_file):
+            contexts, questions, choicess, targs = [], [], [], []
+            data = [json.loads(l) for l in open(data_file, encoding="utf-8")]
+            for example in data:
+                context = example["premise"]
+                choice1 = example["choice1"]
+                choice2 = example["choice2"]
+                question = example["question"]
+                question = "What was the cause of this ?" if question == "cause" else "What happened as a result ?"
```
sleepinyourhat: Is the space before '?' intentional? (i.e., is this valid/expected with all the tokenizers we might use?)

src/tasks/tasks.py
```diff
@@ -2224,3 +2225,152 @@ def get_metrics(self, reset=False):
...
+        instances = map(_make_instance, *split)
+        return instances
+
+    def get_metrics(self, reset=False):
+        '''Get metrics specific to the task'''
+        acc = self.scorer1.get_metric(reset)
+        return {'accuracy': acc}
+
+
+@register_task('swag', rel_path='SWAG/')
```
sleepinyourhat: Delete for now.
(Or just combine these PRs.)

src/models.py
W4ngatang: Yea they're concatenated during preprocessing. The question and the answer choice get the same segment ID, which seems standard for BERT.

src/evaluate.py
W4ngatang: fixed

PR #638, opened 2019-05-01
Remove grounded tasks

zphang:
Removed grounded tasks, methods for grounded tasks in models.py, references to grounded tasks in configs, and models used for grounded tasks.

**Note**: I modified `scripts/final_glue_runs.sh` to remove the references, let me know if you prefer them to be handled another way.


scripts/final_glue_runs.sh
```diff
@@ -132,14 +132,6 @@ JIANT_OVERRIDES="pretrain_tasks = dissentwikifullbig, run_name = dissent-elmo, e
 
 ## MSCOCO ##
 
-# Sam is running.
```
sleepinyourhat: Probably better to leave this file as a record, though we will have to move it elsewhere eventually. How about commenting these out and marking them as deprecated?

zphang: Restored + commented out with deprecation comment.

W4ngatang:
I think one paper that uses Jiant uses these tasks. @najoungkim could you confirm? If so, Najoung, could you make a branch with the version of jiant you used for the paper and point to that where needed?

najoungkim:
Yes the function words paper uses the grounded task. I've been using the naacl_probingpaper branch, which will be cleaned up & given pointers to soon.

zphang:
In that case, should we keep the grounded tasks?

sleepinyourhat:
I think the plan was to make that branch separate/independent. If no one is currently willing to maintain Grounded, let's still cut it.

sleepinyourhat:
Reddit is in a similar position, but my understanding is that we're much more confident that Reddit is working/bug-free.

zphang:
Merged in style changes, also did some mild clean-up of unused imports in the relevant files.

sleepinyourhat:
@zphang Is there a reason this didn't remove `src/modules/cnns`?

zphang:
Hm, I think I missed it after merging in the style changes. I'll submit a PR for removal.

sleepinyourhat:
Tx!

PR #640, opened 2019-05-02
Adding Winograd Coreference

pruksmhc:
Added components: 
-A preprocessing script for winograd with tests
-Added superglue to the list in tasks/__init__.py 
-Adding to tasks.py


scripts/winograd/preprocess_winograd.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env python
+
+"""
+This file will preprocess Winograd, aligning the span indices to the tokenizaer of
```
sleepinyourhat: 'preprocess the SuperGLUE Winograd Schema Challenge data,'
There's no standard format for Winograd schemas, so need to specify.

src/tasks/tasks.py
```diff
@@ -2224,3 +2224,26 @@ def get_metrics(self, reset=False):
...
+class WinogradCoreferenceTask(SpanClassificationTask):
+    def __init__(self, path,  **kw):
+        self._files_by_split = {'train': "train.json", 'val': "val.json",'test': "test_final.json"}
+        self.num_spans = 2
+        super().__init__(files_by_split=self._files_by_split, label_file="labels.txt", path=path, **kw)
+        self.val_metric = "%s_acc" % self.name 
+
+    def update_metrics(self, logits, labels, tagmask=None):
+        logits, labels = logits.detach(), labels.detach()
+        def one_hot_v(batch, depth=2):
```
sleepinyourhat: Comment: What does this do?

src/utils/utils.py
```diff
@@ -188,7 +188,7 @@ def split_data(data, ratio, shuffle=1):
         splits[1].append(col[split_pt:])
     return tuple(splits[0]), tuple(splits[1])
     
-def unbind_predictions(self, preds: torch.Tensor) -> Iterable[np.ndarray]:
+def unbind_predictions(preds: torch.Tensor) -> Iterable[np.ndarray]:
```
sleepinyourhat: What's going on here? This method does get called in a few places, so this change should break the repo...

src/tasks/tasks.py
```diff
@@ -2224,3 +2224,26 @@ def get_metrics(self, reset=False):
         f1 = (f11 + f12 + f13) / 3
         return {'accuracy': acc, 'f1': f1, 'precision': pcs, 'recall': rcl}
 
+@register_task('winograd-coreference', rel_path = 'winograd-coref')
+class WinogradCoreferenceTask(SpanClassificationTask):
+    def __init__(self, path,  **kw):
+        self._files_by_split = {'train': "train.json", 'val': "val.json",'test': "test_final.json"}
```
W4ngatang: these should be jsonl

src/utils/utils.py
W4ngatang: It was previously a task method (that didn't need to be) and is pulled out, so this deletes `self`, but idk why it wasn't raising issues earlier.

src/tasks/tasks.py
pruksmhc: After running the preprocessing script, you will get JSON data instead of jsonl. I can change this, but don't think it's too urgent.

src/tasks/tasks.py
pruksmhc: This creates one hot embeddings given the logits , which is what the F1 scorer and acc scorer expects.

sleepinyourhat: Just suggesting you comment in the code.

src/utils/utils.py
sleepinyourhat: But isn't it called with [something].unbind_predictions() elsewhere? Won't that break?

sleepinyourhat: @pruksmhc Can you confirm that all the calls to this are non-broken?

src/tasks/__init__.py
```diff
@@ -26,6 +26,8 @@
 
 ALL_SUPERGLUE_TASKS = ["commitbank", "wic"]
 
+ALL_SUPERGLUE_TASKS = ["winograd-coreference"]
```
sleepinyourhat: !!

src/tasks/tasks.py
```diff
@@ -2651,7 +2651,41 @@ def _make_instance(question, choices, label, idx):
...
+            ones = torch.sparse.torch.eye(depth).cuda()
+            return ones.index_select(0, batch)
+
+        binary_preds = make_one_hot(logits, depth=2)
+        # Make label_ints a batch_size list of labels
+        label_ints = torch.argmax(labels, dim=1)
+        self.f1_scorer(binary_preds, label_ints)
+        self.acc_scorer(binary_preds.long(), labels.long())
+
     def get_metrics(self, reset=False):
```
sleepinyourhat: It looks like get_metrics for the previous task got cut off in a merge fix.

src/utils/utils.py
sleepinyourhat: ...confirmed?

PR #646, opened 2019-05-02
Fix fat finger on self.path breaking tests

Iwontbecreative:
Fix issue introduced by my manually fixing a merge conflict. 
```
"self.path" -> self.path
```


Iwontbecreative:
@zphang might have the fix for this into an exiting MR, close if not relevant anymore

sleepinyourhat:
Incorporated into #638.

Issue #602, opened 2019-04-26
Our logs are a nightmare hellscape


sleepinyourhat:
We need to log less, and structure the logs a bit more clearly. We don't need to fully fix this, but we should at least deal with the easy cases before the launch.

All of these lines are either unhelpful or fully redundant with other lines in the logs:

- 04/26 05:26:16 PM: Waiting on git info....
- All the task-specific overrides that aren't used. (Things like "edges-coref-ontonotes": { ... })
- 04/26 05:26:16 PM: Saved config to /Users/Bowman/Drive/JSALT/jiant-demo/mtl-sst-mrpc/params.conf
- 04/26 05:26:25 PM: Your label namespace was 'idxs'. We recommend you use a namespace ending with 'labels' or 'tags', so we don't add UNK and PAD tokens by default to your vocabulary.  See documentation for `non_padded_namespaces` parameter in Vocabulary.
- 04/26 05:26:31 PM: 	Task 'sst': cleared in-memory data.
- 04/26 05:26:31 PM: 	 Tasks: ['mrpc', 'sst']
- 04/26 05:26:31 PM: 	Not using character embeddings!
- 04/26 05:26:31 PM: Done initializing parameters; the following parameters are using their default initialization from their code
04/26 05:26:31 PM:    _text_field_embedder.token_embedder_words.weight
- 04/26 05:26:31 PM: Using BoW architecture for shared encoder!  
- 04/26 05:26:31 PM: Converting Params object to dict; logging of default values will not occur when dictionary parameters are used subsequently.
- 04/26 05:26:31 PM: CURRENTLY DEFINED PARAMETERS: ... [followed by a ton of redundant lines]
- 04/26 05:26:31 PM: Will run the following steps: Evaluating model on tasks: mrpc
- 04/26 05:26:31 PM: In strict mode because do_target_task_training is off. Will crash if any tasks are missing from the checkpoint.
- 04/26 05:26:31 PM: Task 'mrpc': sorting predictions by 'idx'
- 04/26 05:25:40 PM: Sampling tasks proportional to number of training examples.

These are helpful, but need a cleaner format:

04/26 05:26:31 PM: >> Trainable param mrpc_mdl.pooler.project.weight: torch.Size([256, 50]) = 12800

-> Just show the name and size, with some indentation since this is a long list.

04/26 05:25:47 PM: Statistic: sst_accuracy
04/26 05:25:47 PM: 	training: 0.542500
[...]

PR #650, opened 2019-05-03
Log cleanup

sleepinyourhat:
A few changes to try to make the logs a bit less bizarre in average use cases. They're still pretty bad.

Goes 5% of the way to addressing #602.

Nearly all of these changes are string edits only.


sleepinyourhat:
Merge when done.

Issue #495, opened 2019-02-20
Download glue data, MNLI doesn't do diagnostic download


anhad13:
When downloading only MNLI through the download glue script, the diagnostic file doesn't get downloaded, without which an error come up during training.

W4ngatang:
Closed with #655

PR #655, opened 2019-05-03
Update downloader script to download both diagnostic files

W4ngatang:
Two fixes:

- downloading "diagnostic" downloads both diagnostic files we use (one for MNLI and one for GLUEDiagnostic)
- when downloading MNLI, also download the diagnostic data

Addresses #495

### DIFF
--- a/scripts/download_glue_data.py
+++ b/scripts/download_glue_data.py
@@ -1,5 +1,8 @@
 ''' Script for downloading all GLUE data.
 
+Example usage:
+    python download_glue_data.py --data_dir data --tasks all
+
 Note: for legal reasons, we are unable to host MRPC.
 You can either use the version hosted by the SentEval team, which is already tokenized,
 or you can download the original data from (https://download.microsoft.com/download/D/4/6/D46FF87A-F6B9-4252-AA8B-3604ED519838/MSRParaphraseCorpus.msi) and extract the data from it manually.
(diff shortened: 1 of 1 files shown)


scripts/download_glue_data.py
```diff
@@ -1,47 +1,46 @@
-""" Script for downloading all GLUE data.
+''' Script for downloading all GLUE data.
 
-Example usage:
-  python download_glue_data.py --data_dir data --tasks all
```
sleepinyourhat: Why delete this?

scripts/download_glue_data.py
```diff
@@ -1,47 +1,46 @@
...
+rm MSRParaphraseCorpus.msi
 
-  mkdir MRPC
-  cabextract MSRParaphraseCorpus.msi -d MRPC
-  cat MRPC/_2DEC3DBE877E4DB192D17C0256E90F1D | tr -d $'\r' > MRPC/msr_paraphrase_train.txt
-  cat MRPC/_D7B391F9EAFF4B1B8BCE8F21B20B1B61 | tr -d $'\r' > MRPC/msr_paraphrase_test.txt
-  rm MRPC/_*
-  rm MSRParaphraseCorpus.msi
-"""
+1/30/19: It looks like SentEval is no longer hosting their extracted and tokenized MRPC data, so you'll need to download the data from the original source for now.
```
sleepinyourhat: Delete these.

scripts/download_glue_data.py
W4ngatang: because I have fat fingers. Reverting

PR #665, opened 2019-05-04
Fix task loading test

zphang:


sleepinyourhat:
This seems to have been resolved without a change in this repo.

Issue #666, opened 2019-05-04
COPA expects test set labels


sleepinyourhat:
Trying to load COPA after running `download_superglue_data.py` yields:

```
05/04 03:09:19 PM: Fatal error in main():
Traceback (most recent call last):
  File "main.py", line 553, in <module>
    main(sys.argv[1:])
  File "main.py", line 378, in main
    pretrain_tasks, target_tasks, vocab, word_embs = build_tasks(args)
  File "/Users/Bowman/Drive/JSALT/jiant/src/preprocess.py", line 276, in build_tasks
    tasks, pretrain_task_names, target_task_names = get_tasks(args)
  File "/Users/Bowman/Drive/JSALT/jiant/src/preprocess.py", line 458, in get_tasks
    task = _get_task(name, args, data_path=data_path, scratch_path=scratch_path)
  File "/Users/Bowman/Drive/JSALT/jiant/src/preprocess.py", line 415, in _get_task
    task.load_data()
  File "/Users/Bowman/Drive/JSALT/jiant/src/tasks/tasks.py", line 2413, in load_data
    self.test_data_text = _load_split(os.path.join(self.path, "test_ANS.jsonl"))
  File "/Users/Bowman/Drive/JSALT/jiant/src/tasks/tasks.py", line 2389, in _load_split
    data = [json.loads(l) for l in open(data_file, encoding="utf-8")]
FileNotFoundError: [Errno 2] No such file or directory: '/Users/Bowman/Drive/JSALT/jiant/glue_data/COPA/test_ANS.jsonl'
Traceback (most recent call last):
  File "main.py", line 564, in <module>
    raise e  # re-raise exception, in case debugger is attached.
  File "main.py", line 553, in <module>
    main(sys.argv[1:])
  File "main.py", line 378, in main
    pretrain_tasks, target_tasks, vocab, word_embs = build_tasks(args)
  File "/Users/Bowman/Drive/JSALT/jiant/src/preprocess.py", line 276, in build_tasks
    tasks, pretrain_task_names, target_task_names = get_tasks(args)
  File "/Users/Bowman/Drive/JSALT/jiant/src/preprocess.py", line 458, in get_tasks
    task = _get_task(name, args, data_path=data_path, scratch_path=scratch_path)
  File "/Users/Bowman/Drive/JSALT/jiant/src/preprocess.py", line 415, in _get_task
    task.load_data()
[...]

PR #667, opened 2019-05-04
COPA fix.

sleepinyourhat:
#666


W4ngatang:
This happens with a few tasks, so I'll cover the rest in a PR soon.

PR #669, opened 2019-05-04
COPA example config

sleepinyourhat:
Gets ~61% on validation after a bit of tuning. Is that any good? (We don't report validation numbers anywhere yet. We really should.)

Incorporates #667, so merge that first.


W4ngatang:
61 is a little low, I got high 60s / low 70s on dev. Will add my configs soon. Will add dev results soon.

PR #671, opened 2019-05-04
Removing CNN models

zphang:


sleepinyourhat:
592 fewer lines to maintain!

PR #672, opened 2019-05-04
Add come some comments/clarafications on our use of 'epoch'

sleepinyourhat:
This is an awkward partial-fix, but it fixes everything that's visible in the logs, and I think it does more good than harm.

### DIFF
--- a/src/trainer.py
+++ b/src/trainer.py
@@ -189,10 +189,10 @@ class SamplingMultiTaskTrainer:
             their ``forward`` method returns a dictionary with a "loss" key, containing a
             scalar tensor representing the loss function to be optimized.
         patience , optional (default=2)
-            Number of epochs to be patient before early stopping.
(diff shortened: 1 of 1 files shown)


src/trainer.py
```diff
@@ -384,8 +384,8 @@ def get_scaling_weights(self, scaling_method, num_tasks, task_names, task_n_trai
             scaling_weights = 1 / np.log(task_n_train_examples)
         elif scaling_method == "max_inverse":
             scaling_weights = 1 / task_n_train_examples
-        # Weighting losses based on best epochs for each task from a previous uniform run,
-        # normalizd by max epoch
+        # Weighting losses based on best validation for each task from a previous uniform run,
+        # normalizd by the maximum
```
W4ngatang: The maximum what? (I'm actually not sure what this does)

Issue #601, opened 2019-04-26
Double check ccg-bert.conf


sleepinyourhat:
https://github.com/nyu-mll/jiant/blob/master/config/ccg-bert.conf

This uses BERT, but also uses the default sent_enc = rnn setup, which doesn't really make sense...

PR #675, opened 2019-05-04
Adding CCG Bert Config

pruksmhc:
Changes:
-Adding a config file.
BERT with a none sentence encoder and MLP classifier. 

-Fixing self.path reference in CCG. 
https://github.com/nyu-mll/jiant/issues/601


config/ccg_bert.conf
```diff
@@ -0,0 +1,33 @@
+include "defaults.conf"	
+include
```
sleepinyourhat: ?

config/ccg_bert.conf
```diff
@@ -0,0 +1,33 @@
+include "defaults.conf"
```
sleepinyourhat: If this is meant to be a useful examples, add at least a few comments. (Follow demo.conf or the new copa_bert.conf.)

config/ccg_bert.conf
```diff
@@ -0,0 +1,33 @@
+include "defaults.conf"	
+include 
+pretrain_tasks = ccg	
+target_tasks = ccg	
+project_dir = ccg_exps
+bert_model_name = bert-large-uncased	
+tokenizer = ${bert_model_name}	
+exp_dir = ${bert_model_name}
```
sleepinyourhat: Why this setup?

config/ccg_bert.conf
```diff
@@ -0,0 +1,33 @@
+include "defaults.conf"	
+include 
+pretrain_tasks = ccg	
+target_tasks = ccg	
+project_dir = ccg_exps
+bert_model_name = bert-large-uncased	
+tokenizer = ${bert_model_name}	
+exp_dir = ${bert_model_name}	
+run_dir = ""
```
sleepinyourhat: ???

config/ccg_bert.conf
```diff
@@ -0,0 +1,33 @@
...
+bert_model_name = bert-large-uncased	
+tokenizer = ${bert_model_name}	
+exp_dir = ${bert_model_name}	
+run_dir = ""	
+bert_fine_tune = 1	
+do_target_task_training = 0	
+
+batch_size = 16	
+max_seq_len = 510	
+reload_task = 1
```
sleepinyourhat: Why override these in the config file?

pruksmhc: max_seq_len definitely needs to be overriden because we add CLS and SEP tags, the others are not necessary so I'll remove.

config/ccg_bert.conf
pruksmhc: This isn't going to contribute anything that the COPA example doesn't so I don't think this will be an example (it would be more for people who are running CCG, and also to address the issue opened about the config file).

sleepinyourhat: Sounds good. Just add a one-line header explaining what it's for, at least?

sleepinyourhat: Just a one-liner saying what this is?

Issue #527, opened 2019-03-18
Clean up ipynb files


pruksmhc:
Self-explanatory, move to scripts folder?

pruksmhc:
https://github.com/nyu-mll/jiant/commit/15b46d350f8a8bee50b6ac0995044ae503c07b9b

PR #676, opened 2019-05-04
Clean up scripts and config dirs.

sleepinyourhat:
Move a few things out of the way, and delete some things that are clearly stale/deprecated. Not perfect, but probably net helpful.


sleepinyourhat:
#609 #527

PR #681, opened 2019-05-05
Clean up and add configs+scripts

W4ngatang:
- add config files used by BoW and BERT baselines
- move edgeprobe configs to their own folder and adjust paths throughout


config/superglue-bow.conf
```diff
@@ -0,0 +1,86 @@
...
+    attention = none
+}
+transfer_paradigm = finetune
+
+// Training settings
+dropout = 0.1
+optimizer = adam
+batch_size = 16
+max_epochs = 10
+lr = .00001
```
sleepinyourhat: FWIW, this seems pretty low for non-BERT.

config/superglue-bow.conf
```diff
@@ -0,0 +1,86 @@
...
+rte_classifier_dropout = 0.1
+rte_val_interval = 100
+rte_lr = 0.00001
+
+wnli = {}
+wnli_classifier_dropout = 0.1
+wnli_val_interval = 100
+wnli_lr = 0.00001
+
+mrpc = {}
```
sleepinyourhat: Delete these non-SuperGLUE tasks?

sleepinyourhat: ...and add the SuperGLUE ones?

config/superglue-bert.conf
```diff
@@ -0,0 +1,89 @@
...
+sent_enc = "none"
+sep_embs_for_skip = 1
+classifier = log_reg // following BERT paper
+transfer_paradigm = finetune // finetune entire BERT model
+
+// Training settings
+dropout = 0.1 // following BERT paper
+optimizer = bert_adam
+batch_size = 4
+max_epochs = 10
```
sleepinyourhat: Make sure this is in the paper.

config/superglue-bert.conf
```diff
@@ -0,0 +1,89 @@
...
+do_pretrain = 1
+do_target_task_training = 1
+do_full_eval = 1
+write_preds = val,test  // Somewhat slow, but should be useful for analysis.
+write_strict_glue_format = 1
+
+// These override the task-specific configs set in defaults.conf
+cola = {}
+cola_classifier_dropout = 0.1
+cola_val_interval = 100
```
sleepinyourhat: For the smallest SuperGLUE tasks, this is pretty high—potentially over an epoch, even with our tiny batch size.

sleepinyourhat: It's also a bit odd that this doesn't match the BOW setup due to the batch size diff, though not fatal.

config/superglue-bert.conf
W4ngatang: It is.

config/superglue-bow.conf
W4ngatang: Will delete, but not sure what you mean about adding (there aren't any task-specific hyperparams).

config/superglue-bert.conf
W4ngatang: Oh these settings aren't actually used, I forget why I copied them here.
I was planning on adding the validation intervals to the actual script, because it depends directly on batch size and I wanted to make that obvious.

W4ngatang: The BOW point is fair though...

config/superglue-bow.conf
W4ngatang: :/

config/superglue-bow.conf
sleepinyourhat: Got it. Assumed there were some that would be specified in the script, but makes sense.

W4ngatang:
Ah good catch.

@pruksmhc can you take a look at this and make sure the command I wrote for WSC is correct?

scripts/superglue-baselines.sh
```diff
@@ -0,0 +1,48 @@
...
+function rte() {
+    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = rte, pretrain_tasks = \"rte-superglue\", target_tasks = \"rte-superglue,superglue-diagnostic\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 625"
+}
+
+function wic() {
+    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = wic, pretrain_tasks = \"wic\", target_tasks = \"wic\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 1000"
+}
+
+function wsc() {
+    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = wsc, pretrain_tasks = \"winograd-coreference\", target_tasks = \"winograd-coreference\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 125"
```
pruksmhc: Add optimizer="adam" here since that's what we report in the paper (BERT Adam performance was acting a bit funky, I'll open up an issue).

pruksmhc: Also val_interval should be 139 (554/4, which is train size/batch size)

PR #682, opened 2019-05-05
Moving Example Configs to a Dedicated Folder, adding STILTS example

pruksmhc:
Changes:
* Adding examples for STILTS
* Moving examples configs to example folder


config/examples/stilts_example/mrpc_mnli_phase1.conf
```diff
@@ -0,0 +1,36 @@
+include "../../defaults.conf"
+
+// For STILTS training, first run this experiment, before 
+// mrpc_mnli_phase2.conf.In this STILTS experiment, we add first train on MRPC,
```
sleepinyourhat: Fix spacing

sleepinyourhat: Why MRPC then QQP then MRPC? This doesn't make sense (MRPC is way too small to be useful for pretraining), and it doesn't match the filename...

config/examples/stilts_example/mrpc_mnli_phase2.conf
```diff
@@ -0,0 +1,17 @@
+include "./bert_mrpc.conf"
+
+// First, run mrpc_mnli_phase1.conf first, then this config file.
```
sleepinyourhat: first ... first

config/examples/stilts_example/mrpc_mnli_phase1.conf
sleepinyourhat: Maybe replace this with something like MNLI => RTE? That was the simplest success case for STILTs.

If you'd rather stick closer to this, maybe switch to just BERT+QQP => MRPC, which is at least fairly likely to help on the target task.

pruksmhc: Just to make sure, this is RTE pretraining, then intermediate task of MNLI, and then finetuning on RTE?

sleepinyourhat: Simper: Train BERT on MNLI, then on RTE.

sleepinyourhat: Probably doable with a single conf.

W4ngatang: Definitely should be a single run / configuration. Something like:
`pretrain_tasks = mnli`
`target_tasks = rte`
Would do it.

config/examples/copa_bert.conf
```diff
@@ -0,0 +1,37 @@
+// An example configuration for the CommitmentBank task with BERT.
```
sleepinyourhat: CommitmentBank != COPA

config/examples/stilts_example.conf
```diff
@@ -0,0 +1,38 @@
...
+//BERT-specific parameters
+bert_embeddings_mode = "top"    
+bert_fine_tune = 1
+sep_embs_for_skip = 1
+sent_enc = "none"
+classifier = log_reg // following BERT paper
+
+dropout = 0.1 // following BERT paper
+optimizer = bert_adam
+max_epochs = 1
```
sleepinyourhat: Should be 3 or 4 for standard BERT training.

config/examples/stilts_example.conf
```diff
@@ -0,0 +1,38 @@
+// This config is for STILTS training for BERT -> MNLI -> RTE
```
sleepinyourhat: Mention that we're using base. Most users will want to mess with this then switch to large.

PR #683, opened 2019-05-08
Very minor fixes

zphang:


zphang:
Yes. These would have thrown errors before if they were being used, but I believe that they are mostly unused options.

PR #690, opened 2019-05-21
Moving WSC span realignment to loading time

pruksmhc:
This PR currently holds: 
-Moving the span realignment to loading time as opposed to a separate preprocessing step that needs to be run before running the experiment.
-Fixing the WSC class so that it matches the format of the final version of the data 
-Some fixes to the WSC configuration so that it matches the experiments

Left to do:
-Point the test for preprocessing to the new location of realign_spans
-A test for loading checkpoints


config/superglue-bert.conf
```diff
@@ -8,7 +8,6 @@ exp_name = "bert-large-cased"
 max_seq_len = 256 // Mainly needed for MultiRC, to avoid over-truncating
                   // But not 512 as that is really hard to fit in memory.
 tokenizer = "bert-large-cased"
-
```
sleepinyourhat: Restore?

config/superglue-bert.conf
```diff
@@ -42,3 +41,7 @@ do_full_eval = 1
 write_preds = "val,test"
 write_strict_glue_format = 1
 
+// For WSC
```
sleepinyourhat: Why is this in this PR?

src/tasks/tasks.py
```diff
@@ -2092,11 +2081,11 @@ def _stream_records(self, filename):
     def load_data(self):
         iters_by_split = collections.OrderedDict()
         for split, filename in self._files_by_split.items():
-            iter = list(self._stream_records(filename))
-            iters_by_split[split] = iter
+            if "test" in filename:
```
sleepinyourhat: This seems brittle. What if someone tries to evaluate on a span task with a public test set?

src/trainer.py
```diff
@@ -1229,6 +1229,7 @@ def _restore_checkpoint(self, search_phases_in_priority_order=["pretrain"]):
 
         metric_states = torch.load(metric_state_path)
         for metric_name, metric_state in metric_states.items():
+
```
sleepinyourhat: Cut?

src/utils/data_loaders.py
```diff
@@ -11,11 +11,40 @@
...
+    Returns:
+        List of dictionaries of the aligned spans and tokenized text.
+    """
+    rows = pd.read_json(file_name, lines=True)
+    # realign spans
+    rows = rows.apply(lambda x: realign_spans(x, tokenizer_name), axis=1)
+
+    if has_labels is False:
+        rows["label"] = False
+    return list(rows.T.to_dict().values())
```
sleepinyourhat: Cut?

src/utils/data_loaders.py
```diff
@@ -11,11 +11,40 @@
...
+        target: dict that contains the spans  
+    Args:
+        tokenizer_name: str, 
+        file_name: str, 
+        label_fn: function that expects a row and outputs a transformed row with labels tarnsformed.
+    Returns:
+        List of dictionaries of the aligned spans and tokenized text.
+    """
+    rows = pd.read_json(file_name, lines=True)
+    # realign spans
```
sleepinyourhat: Explain what this does, either here or in the header comment.

config/superglue-bert.conf
pruksmhc: Not sure why it's appearing in github as removed because it isn't

src/tasks/tasks.py
sleepinyourhat: This still seems dangerously brittle...

src/utils/data_loaders.py
```diff
@@ -11,11 +11,40 @@
...
 
+def load_span_data(
+        tokenizer_name, 
+        file_name,
+        label_fn=None,
+        has_labels=True
+    ):
+    """
+    Load a span-related task file in .jsonl format, does 
+    re-alignment of spans, and tokenizes the text.
```
sleepinyourhat: Do we explain what this means anywhere?

PR #692, opened 2019-05-22
Tests for checkpoint logic + some fixes to checkpointing

pruksmhc:
This PR is dedicated to creating tests for checkpointing and fixing some checkpointing errors. 
The general flow is the below:
-We save the best metric thus far. For pretraining, we do multitask training, and save during each epoch the model states (both the sentence encoder and the task-specific modules).
-For finetuning, we must also save task-specific versions of the sentence encoder. If we don't allow pretrained weights for the task-specific modules, we do not load the parameter weights for a module if the sentence encoder has not been trained on a specific task yet. 

Changes:
-Making the code less specific to target_train and adding asserts to make sure that the checkpoints loaded are the best ones rather than the most recent. For example, removing redundant model_target_state_best.th checkpoint. 
-Adding the demo.sh run to .circleCi build since running the script from tests seems odd to make sure that the output of demo.sh is the same.


sleepinyourhat:
Looks sane so far, but I don't have time today to fully reverse engineer what this is doing.

General considerations for checkpointing:
- During a training phase, we should maintain the best checkpoint and the most recent checkpoint. (We don't fully do this during the target task training phase, but we should.)
- If we restart training after a crash, we should load the most recent checkpoint, which should lead to training behaving _exactly_ as it would have without the crash (except for random variation).
- If we're evaluating a model or using it as a starting point for transfer, we should load the best checkpoint for that model, not the most recent.
- When fine-tuning the same model on multiple tasks, each task should behave independently. This requires repeatedly restoring from the same checkpoint.

main.py
```diff
@@ -118,22 +113,31 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
+                else:
+                    assert_for_log(
+                        args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
+                    )
+
+                if args.transfer_paradigm == "finetune":
+                    # Save model so we have a checkpoint to go back to after each
+                    # task-specific finetune.
+                    model_state = model.state_dict()
+                    model_path = os.path.join(
```
pruksmhc: @W4ngatang by the way get_best_checkpoint is currently structured, if we do pretraining then target task training, then wouldn't we during target training always load and use the best pretrained model for finetuning? 
If so, then why is this untrained_pre_target_train necessary (as opposed to just always loading the best pretrained model before each finetuning target training for each task?

W4ngatang: Sometimes we don't want to do any pretraining (e.g. we just want to use BERT for a set of target tasks), so we to create a checkpoint to refer back to at the start of each fine-tuning.

sleepinyourhat:
And of course, make sure tests pass.

main.py
```diff
@@ -118,22 +113,32 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
-                torch.save(model_state, model_path)
+            if args.do_pretrain == 1:
+                best_pretrain = get_best_checkpoint_path(args.run_dir, "pretrain")
+                if best_pretrain:
+                    load_model_state(
+                        model, best_pretrain, args.cuda, task_names_to_avoid_loading, strict=strict
+                    )
+                else:
+                    assert_for_log(
+                        args.allow_untrained_encoder_parameters,
```
W4ngatang: I think this check should occur _outside_ of the pretraining check, the logic being that if you're not doing pretraining, you are using an un(pre)trained model.

is the name "allow_untrained_encoder_parameters" accurate here? At the end of pretraining, even if you don't load any checkpoints, you still have trained parameters.

main.py
W4ngatang: So I think this should happen outside of the pretrain check, i.e. you create this if you are targe training but not pretraining.

main.py
```diff
@@ -248,30 +251,38 @@ def _kill_tb_child():
...
+        1) best checkpoint from the phase so far
+        2) nothing found (empty string) """
+    macro_best = glob.glob(os.path.join(run_dir, "model_state_%s_epoch_*.best_macro.th" % phase))
     if len(macro_best) > 0:
         assert_for_log(len(macro_best) == 1, "Too many best checkpoints. Something is wrong.")
         return macro_best[0]
+    return ""
+
 
-    pre_target_train = glob.glob(os.path.join(run_dir, "model_state_untrained_pre_target_train.th"))
```
W4ngatang: I think this should be kept in, see above comments about this

main.py
```diff
@@ -483,30 +491,31 @@ def main(cl_arguments):
...
-                # Save this fine-tune model with a task specific name.
-                finetune_path = os.path.join(args.run_dir, "model_state_%s_best.th" % task.name)
-                os.rename(layer_path, finetune_path)
-
                 # Reload the original best model from before target-task
-                # training.
-                pre_finetune_path = get_best_checkpoint_path(args.run_dir)
+                # training since we specifically finetune for each task.
+                pre_target_train = get_best_checkpoint_path(args.run_dir, "pretrain")
+                if len(pre_target_train) > 0:
```
W4ngatang: might as well move this check into `get_best_ckpt_path` if we're going to / want to check it every time?

tests/test_checkpointing.py
```diff
@@ -0,0 +1,199 @@
...
+from allennlp.data.token_indexers import SingleIdTokenIndexer
+from allennlp.data import Instance, Token, vocabulary
+from ..allennlp_mods.numeric_field import NumericField
+from allennlp.data.fields import LabelField, ListField, MetadataField, TextField
+from allennlp.common.params import Params
+from allennlp.data.iterators import BasicIterator, BucketIterator
+from allennlp.training.learning_rate_schedulers import (  # pylint: disable=import-error
+    LearningRateScheduler,
+)
+from allennlp.training.optimizers import Optimizer
```
W4ngatang: do you mind cleaning this imports up a bit? (default libraries on top, group sources together, etc.)

main.py
```diff
@@ -248,30 +251,38 @@ def _kill_tb_child():
...
+        2) nothing found (empty string) """
+    macro_best = glob.glob(os.path.join(run_dir, "model_state_%s_epoch_*.best_macro.th" % phase))
     if len(macro_best) > 0:
         assert_for_log(len(macro_best) == 1, "Too many best checkpoints. Something is wrong.")
         return macro_best[0]
+    return ""
+
 
-    pre_target_train = glob.glob(os.path.join(run_dir, "model_state_untrained_pre_target_train.th"))
+def do_finetune_evaluation(args, target_tasks, model, splits_to_write):
```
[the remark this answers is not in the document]
pruksmhc: I was first modularizing this out in order to simmplify testing for checkpoinnting but then found another better way to test. Revertin gthis

main.py
pruksmhc: Realized it's already a part of the refactored function so deleting this redundant check.

main.py
pruksmhc: This allows for random sentence encoders (skipping pretraining step to only run do_target_task_training). Moving this to if args.do_pretrianing = 0.

main.py
```diff
@@ -80,20 +78,17 @@ def setup_target_task_training(args, target_tasks, model, strict):
     Saves model states from pretraining if applicable, and
     loads the correct model state for the target task training
     stage.
-
```
sleepinyourhat: Why cut these?

sleepinyourhat: Here and elsewhere.

pruksmhc: Those were extra changes-reverting.

main.py
```diff
@@ -118,22 +117,32 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
+                )
 
-            if args.transfer_paradigm == "finetune":
-                # Save model so we have a checkpoint to go back to after each
-                # task-specific finetune.
-                model_state = model.state_dict()
-                model_path = os.path.join(args.run_dir, "model_state_untrained_pre_target_train.th")
-                torch.save(model_state, model_path)
+                if args.transfer_paradigm == "finetune":
+                    # We want to do target training without pretrianing, thus
```
W4ngatang: typo

main.py
```diff
@@ -483,30 +476,28 @@ def main(cl_arguments):
                 phase="target_train",
             )
 
-            # Now that we've trained a model, revert to the normal checkpoint
-            # logic for this task.
+            # Now that we've trained the task specific module for the task,
+            # since we are accumulating the best parameters for
```
W4ngatang: I'm confused by this comment. I guess it doesn't apply in the finetune setting, but in the frozen setting, we're accumulating all of the best parameters in the checkpoint `model_state_target_train_best.th`, right? Where is that happening here? I only see us loading that checkpoint.

W4ngatang: Ohhh, is this comment saying we should load the best checkpoint from a particular task's target training?

W4ngatang: That makes sense.

pruksmhc: I had basically deleted target_train_best.th checkpoint since there is also model_state_%s_epoch_target_test.best_macro.th checkpoint. Is there a reason why the best parameters are  in both variables?

PR #696, opened 2019-05-24
Cola var

PterosDiacos:


sleepinyourhat:
Take a look at the test failure—it looks like a ' vs " issue in __init__.

PterosDiacos:
> Take a look at the test failure—it looks like a ' vs " issue in **init**.

Fixed

PR #697, opened 2019-05-27
Update Metrics/Write Preds Testing

pruksmhc:
Tests to determine that updating metrics and writing predictions works as expected.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -16,6 +16,7 @@ dependencies:
         - tensorboard
         - tensorboardX==1.2
         - sendgrid==5.4.1
+        - pyhocon==0.3.35
         - nose2
         - python-Levenshtein
         - pre-commit==1.15.2
--- a/tests/test_update_metrics.py
+++ b/tests/test_update_metrics.py
@@ -1,4 +1,3 @@
-
 import csv
 import os
 import shutil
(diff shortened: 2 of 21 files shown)


tests/test_write_preds.py
```diff
@@ -0,0 +1,183 @@
...
+        self.path = os.path.join(self.temp_dir, "temp_dataset.tsv")
+        self.stsb = tasks.STSBTask(self.temp_dir, 100, "sts-b", tokenizer_name="MosesTokenizer")
+        self.mrpc = tasks.MRPCTask(self.temp_dir, 100, "mrpc", tokenizer_name="MosesTokenizer")
+        self.wic = tasks.WiCTask(self.temp_dir, 100, "wic", tokenizer_name="MosesTokenizer")
+        stsb_val_preds = pd.DataFrame(
+            data={
+                "idx": [0],
+                "labels": [0],
+                "preds": [0],
+                "sent1_str": ["When Tommy dropped his ice cream , Timmy"],
```
sleepinyourhat: Odd example...

tests/test_write_preds.py
```diff
@@ -0,0 +1,183 @@
...
+    return {"n_exs": 4, "preds": [0, 1, 1, 1]}
+
+
+class TestWritePreds(unittest.TestCase):
+    def sentence_to_text_field(self, sent, indexers):
+        """ Helper function to map a sequence of tokens into a sequence of
+        AllenNLP Tokens, then wrap in a TextField with the given indexers """
+        return TextField(list(map(Token, sent)), token_indexers=indexers)
+
+    def setUp(self):
```
sleepinyourhat: Add some comments explaining this one, since it's not that easy to follow.

tests/test_write_preds.py
```diff
@@ -0,0 +1,183 @@
...
+        self.temp_dir = tempfile.mkdtemp()
+        self.path = os.path.join(self.temp_dir, "temp_dataset.tsv")
+        self.stsb = tasks.STSBTask(self.temp_dir, 100, "sts-b", tokenizer_name="MosesTokenizer")
+        self.mrpc = tasks.MRPCTask(self.temp_dir, 100, "mrpc", tokenizer_name="MosesTokenizer")
+        self.wic = tasks.WiCTask(self.temp_dir, 100, "wic", tokenizer_name="MosesTokenizer")
+        stsb_val_preds = pd.DataFrame(
```
sleepinyourhat: Why only one example for some tasks? Worth including at least two.

tests/test_update_metrics.py
```diff
@@ -0,0 +1,53 @@
...
+import src.tasks.tasks as tasks
+import torch
+
+
+class TestUpdateMetricsAccuracy(unittest.TestCase):
+    def setUp(self):
+        self.temp_dir = tempfile.mkdtemp()
+        self.path = os.path.join(self.temp_dir, "temp_dataset.tsv")
+        self.task = tasks.WiCTask("", 100, "winograd", tokenizer_name="MosesTokenizer")
+        predictions = torch.Tensor(
```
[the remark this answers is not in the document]
pruksmhc: I'll decrease it to 2 examples per batch - I just wanted to make sure there was more than 1 per batch since there can be edge cases that can break dimension-wise with only 1 per batch. 
And the predictions are 4d vectors rather than 2d because the predictions are given in 1-hot vector form for each batch.

pruksmhc:
Ready for another review @sleepinyourhat

tests/test_update_metrics.py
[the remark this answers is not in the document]
sleepinyourhat: Add a comment explaining what the three axes are?

tests/test_update_metrics.py
```diff
@@ -0,0 +1,45 @@
...
+        self.task.update_metrics(predictions[0], true_labels[0], tagmask=None)
+        self.task.update_metrics(predictions[1], true_labels[1], tagmask=None)
+        self.imperfect_metrics = self.task.get_metrics(reset=True)
+        self.task.update_metrics(perfect_predictions[0], true_labels[0], tagmask=None)
+        self.task.update_metrics(perfect_predictions[1], true_labels[1], tagmask=None)
+        self.perfect_metrics = self.task.get_metrics(reset=True)
+
+    def test_accuracy(self):
+        # match predictions and labels to be the same format as in src/models.py
+        # Only measures accuracy
```
sleepinyourhat: Be consistent about caps within a file.

tests/test_write_preds.py
```diff
@@ -0,0 +1,202 @@
...
+        evaluate.write_preds(
+            [self.wic], self.val_preds, self.temp_dir, "test", strict_glue_format=True
+        )
+        wic_predictions = pd.read_json(self.temp_dir + "/WIC.jsonl", lines=True)
+        assert "idx" in wic_predictions.columns and "label" in wic_predictions.columns
+        assert wic_predictions.iloc[0]["label"] == "false"
+        assert wic_predictions.iloc[1]["label"] == "true"
+
+    @mock.patch("src.models.MultiTaskModel.forward", side_effect=model_forward)
+    def test_evaluate_and_write(self, model_forward_function):
```
sleepinyourhat: Add a comment?

sleepinyourhat:
Last bit of cleanup, plus make sure tests pass.

tests/test_update_metrics.py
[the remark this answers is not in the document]
W4ngatang: In response to your point about batch size; also run a test to make sure we're catching batch size = 1 bugs (either here or in a separate test) ?

tests/test_write_preds.py
```diff
@@ -0,0 +1,202 @@
...
+    return {"n_exs": 4, "preds": [0, 1, 1, 1]}
+
+
+class TestWritePreds(unittest.TestCase):
+    def sentence_to_text_field(self, sent, indexers):
+        """ Helper function to map a sequence of tokens into a sequence of
+        AllenNLP Tokens, then wrap in a TextField with the given indexers """
+        return TextField(list(map(Token, sent)), token_indexers=indexers)
+
+    def setUp(self):
```
W4ngatang: use snake_case, not camelCase

tests/test_write_preds.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+
+    def test_write_preds_glue(self):
+        evaluate.write_preds(
+            self.glue_tasks, self.val_preds, self.temp_dir, "test", strict_glue_format=True
+        )
+        stsb_predictions = pd.read_csv(self.temp_dir + "/STS-B.tsv", sep="\t")
+        assert "index" in stsb_predictions.columns and "prediction" in stsb_predictions.columns
+        assert stsb_predictions.iloc[0]["prediction"] == 5.00
+        import pdb; pdb.set_trace()
```
W4ngatang: shouldn't be here?

tests/test_write_preds.py
pruksmhc: setUp is predefined by the unittest library - it doesn't recognize set_up()

pruksmhc: I'm following testing code style from examples from the unittest documentation 
https://docs.python.org/3/library/unittest.html

environment.yml
```diff
@@ -16,6 +16,7 @@ dependencies:
         - tensorboard
         - tensorboardX==1.2
         - sendgrid==5.4.1
+        - pyhocon==0.3.35
```
pruksmhc: Not sure why this is appearing in this diff - this is the same as master.

tests/test_update_metrics.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import shutil
+import tempfile
+import unittest
+import src.tasks.tasks as tasks
+import torch
+
+
+class TestUpdateMetricsAccuracy(unittest.TestCase):
+    def setUp(self):
```
W4ngatang: we can't use snake case here? :(

tests/test_write_preds.py
```diff
@@ -0,0 +1,248 @@
...
+from src import evaluate
+import os.path
+from main import evaluate_and_write
+import pandas as pd
+from src.models import MultiTaskModel
+from unittest import mock
+from allennlp.data.token_indexers import SingleIdTokenIndexer
+from allennlp.data import Instance, Token, vocabulary
+from ..allennlp_mods.numeric_field import NumericField
+from allennlp.data.fields import LabelField, ListField, MetadataField, TextField
```
W4ngatang: nit: organize these

tests/test_write_preds.py
```diff
@@ -0,0 +1,248 @@
...
+                    "sent2": "He hooked a snake accidentally , and was so scared he dropped his rod into the water .",
+                    "labels": 1,
+                },
+            ]
+        )
+        indexers = {"bert_wpm_pretokenized": SingleIdTokenIndexer("bert-xe-cased")}
+        self.wic.val_data = [
+            Instance(
+                {
+                    "sent1_str": MetadataField("Room and board yo."),
```
W4ngatang: is this a real example??

tests/test_write_preds.py
```diff
@@ -0,0 +1,248 @@
...
+                    "sent1_str": "A young child is riding a horse.",
+                    "sent2_str": "A child is riding a horse.",
+                },
+            ]
+        )
+        wic_val_preds = pd.DataFrame(
+            data=[
+                {
+                    "idx": 0,
+                    "sent1": "Room and board. ",
```
W4ngatang: The spacing and punctuation seem a little all over the place. It probably doesn't matter, but maybe worth cleaning up?

W4ngatang: e.g. "Hook a fish"  and "He hooked a snake accidentally , ..." below

tests/test_update_metrics.py
```diff
@@ -0,0 +1,41 @@
...
+        self.temp_dir = tempfile.mkdtemp()
+        self.path = os.path.join(self.temp_dir, "temp_dataset.tsv")
+        self.task = tasks.QQPTask("", 100, "qqp", tokenizer_name="MosesTokenizer")
+        predictions = torch.Tensor([[[1, 0], [1, 0]], [[1, 0], [0, 1]]])
```
sleepinyourhat: Explain the shape of this tensor. It's impossible to follow the test otherwise, and tests really need to be self-explanatory.

tests/test_write_preds.py
pruksmhc: Oh nope :P , deleting the yo

tests/test_update_metrics.py
pruksmhc: unittest has a specific namespace for testing,so setUp and tearDown has to use camel case in order to be run before and right after every test.

PR #710, opened 2019-06-20
Adding winogender

pruksmhc:
Changes: 
-Fixing flow of diagnostics in light of recent jiant changes from https://github.com/nyu-mll/jiant/issues/792.
-Added winogender metrics + calculation


scripts/superglue-baselines.sh
```diff
@@ -22,7 +22,7 @@ function multirc() {
 }
 
 function rte() {
-    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = rte, pretrain_tasks = \"rte-superglue\", target_tasks = \"rte-superglue,superglue-diagnostic\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 625"
+    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = rte, pretrain_tasks = \"rte-superglue\", target_tasks = \"rte-superglue,superglue-diagnostic,winograd-diagnostic\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 625"
```
sleepinyourhat: We can't use this name for the task: It's not in Winograd format and it's not associated with the Winograd task in jiant. How about superglue_bias_diagnostic?

src/evaluate.py
```diff
@@ -82,8 +91,8 @@ def evaluate(
                 out = model.forward(task, batch, predict=True)
 
             # We don't want diagnostic tasks to affect the micro and macro average.
-            # Accuracy on diagnostic tasks is hardcoded to 0.
-            if not isinstance(task, GLUEDiagnosticTask):
+            # Accuracy on diagnostic tasks is hardcoded to 0 except for winogender.
```
sleepinyourhat: Make sure we're consistent in how we refer to the new task in comments.

scripts/superglue-baselines.sh
sleepinyourhat: Winogender works, too.

src/evaluate.py
```diff
@@ -123,6 +132,10 @@ def evaluate(
 
         # Combine task_preds from each batch to a single DataFrame.
         task_preds = pd.concat(task_preds, ignore_index=True)
+        if task.name.startswith("winogender"):
```
sleepinyourhat: This seems like an odd place for a special case. If this really needs to be here, add a comment to the code saying why.

src/metrics/winogender_metrics.py
```diff
@@ -0,0 +1,44 @@
+class GenderParity:
+    """
+  Gender parity metric from [Gender Bias in Coreference Resolution](https://arxiv.org/pdf/1804.09301.pdf).
```
sleepinyourhat: I can't find any mention of this in the paper—it might just be on the site. If so, link to the site instead.

src/metrics/winogender_metrics.py
```diff
@@ -0,0 +1,44 @@
+class GenderParity:
```
sleepinyourhat: Can we add a short test case for this?

src/metrics/winogender_metrics.py
```diff
@@ -0,0 +1,44 @@
...
+      Calculate gender parity. 
+      Parameters
+      -------------------
+      predictiosn: list of dicts 
+
+      Returns
+      -------------------
+      None
+      """
+        assert len(predictions) == 464
```
sleepinyourhat: Is there nowhere else we can put this? Seems odd to have it in the metric.

sleepinyourhat: For example, this prevents anyone from filtering/modifying the data themselves.

src/metrics/winogender_metrics.py
```diff
@@ -0,0 +1,44 @@
...
+        for idx in range(int(len(predictions) / 2)):
+            pred1 = predictions[idx * 2]
+            pred2 = predictions[(idx * 2) + 1]
+            try:
+                assert pred1["sent2_str"] == pred2["sent2_str"], (
+                    "Mismatched hypotheses for ids %s and %s"
+                    % (str(pred1["pair_id"]), str(pred2["pair_id"]))
+                )
+            except:
```
sleepinyourhat: Do we need this here? We don't do this for asserts anywhere else.

scripts/superglue-baselines.sh
sleepinyourhat: Or winogender-diagnostic.

src/tasks/tasks.py
```diff
@@ -1389,6 +1393,70 @@ def collect_metrics(ix_to_tag_dict, tag_group):
...
+        self.train_data_text = None
+        self.val_data_text = None
+        self.test_data = None
+        self.acc_scorer = BooleanAccuracy()
+        self.gender_parity_scorer = GenderParity()
+        self.val_mtric = "winogender-diagnostic_accuracy"
+
+    def load_data(self):
+        rows = pd.read_csv(os.path.join(self.path, "winogender.tsv"), sep="\t")
+        # if is_bert, then we will add [CLS], [SEP]
```
sleepinyourhat: What is this comment referring to?

src/tasks/tasks.py
```diff
@@ -1389,6 +1393,70 @@ def collect_metrics(ix_to_tag_dict, tag_group):
...
+            d["labels"] = LabelField(
+                int(record["label"]), label_namespace="labels", skip_indexing=True
+            )
+            return Instance(d)
+
+        instances = map(_make_instance, split)
+        return instances
+
+    def get_metrics(self, reset=False):
+        return {"accuracy": self.acc_scorer.get_metric(reset)}
```
sleepinyourhat: Won't this skip the new metric you introduced?

pruksmhc: Because the scorer depends on the index of the predictions (it depends on calculating the equality of two predictions at a time), which means we would have to a) either restrict what the batch_size is (must be divisible by 2) or we can calculate gender parity at the end (after calculating all the predictions for all the batches).

main.py
```diff
@@ -266,6 +266,10 @@ def get_best_checkpoint_path(args, phase, task_name=None):
             checkpoint = glob.glob(
                 os.path.join(args.run_dir, task_name, "model_state_target_train_val_*.best.th")
             )
+            if len(checkpoint) == 0:
```
sleepinyourhat: Why is this in this PR?

sleepinyourhat: Make sure this matches the comment header.

main.py
```diff
@@ -558,13 +562,28 @@ def main(cl_arguments):
     if args.do_full_eval:
         log.info("Evaluating...")
         splits_to_write = evaluate.parse_write_preds_arg(args.write_preds)
+        diagnostic_tasks = []
+        target_tasks_remove = []
```
sleepinyourhat: _to_remove?

main.py
```diff
@@ -558,13 +562,28 @@ def main(cl_arguments):
...
+        for index, task in enumerate(target_tasks):
+            if task.name in task_modules.ALL_DIAGNOSTICS:
+                diagnostic_tasks.append(task)
+                target_tasks_remove.append(index)
+        # To avoid evaluating twice on a diagnostic task.
+        for index in sorted(target_tasks_remove, reverse=True):
+            del target_tasks[index]
+        if len(diagnostic_tasks) > 0:
+            for task in diagnostic_tasks:
+                ckpt_path = get_best_checkpoint_path(args, "eval", task.name)
```
sleepinyourhat: Does this handle the diagnostic-to-main-task mapping properly?

main.py
```diff
@@ -558,13 +562,28 @@ def main(cl_arguments):
...
+                load_model_state(model, ckpt_path, args.cuda, skip_task_models=[], strict=strict)
+                evaluate_and_write(args, model, [task], splits_to_write)
         if args.do_target_task_training or (args.load_model and not args.do_pretrain):
             # If we either do target task training, or if we only evaluate
             # without pretraining or target task training
             # then we evaluate on the target tasks.
             for task in target_tasks:
                 # Find the task-specific best checkpoint to evaluate on.
-                ckpt_path = get_best_checkpoint_path(args, "eval", task.name)
+                ckpt_path = get_best_checkpoint_path(args, "target_train", task.name)
```
sleepinyourhat: I don't understand why we use 'eval' checkpoints only if there are diagnostic tasks...

scripts/superglue-baselines.sh
```diff
@@ -22,7 +22,7 @@ function multirc() {
 }
 
 function rte() {
-    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = rte, pretrain_tasks = \"rte-superglue\", target_tasks = \"rte-superglue,superglue-diagnostic\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 625"
+    python main.py --config config/superglue-bert.conf --overrides "random_seed = ${seed}, cuda = ${gpuid}, run_name = rte, pretrain_tasks = \"rte-superglue\", target_tasks = \"rte-superglue,superglue-diagnostic,winogender-diagnostic\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 4, val_interval = 625"
```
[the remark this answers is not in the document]
sleepinyourhat: (Before merging.)

src/tasks/tasks.py
```diff
@@ -1389,6 +1393,86 @@ def collect_metrics(ix_to_tag_dict, tag_group):
         return collected_metrics
 
 
+@register_task("winogender-diagnostic", rel_path="RTE/diagnostics", n_classes=2)
+class WinogenderTask(GLUEDiagnosticTask):
+    """Supported wnogender task """
```
sleepinyourhat: Typo. Also, why 'Supported'?

src/metrics/winogender_metrics.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+  Gender parity metric from https://github.com/decompositional-semantics-initiative/DNC.
+  """
+
+    def __init__(self):
+        self.same_preds = 0.0
+        self.diff_preds = 0.0
+
+    def get_metric(self, reset=False):
+        gender_parity = float(self.same_preds) / float(self.same_preds + self.diff_preds)
```
W4ngatang: catch `ZeroDivisionError`?

src/metrics/winogender_metrics.py
```diff
@@ -0,0 +1,42 @@
...
+            sent1_str: str, context sentence,
+            preds: int,
+            pair_id: int
+
+        Returns
+        -------------------
+        None
+        """
+        for idx in range(int(len(predictions) / 2)):
+            pred1 = predictions[idx * 2]
```
W4ngatang: Can you give a comment pointer to where this is being enforced? Seems like it'd be very easy to accidentally have a bug here.

W4ngatang: e.g. what is `len(predictions)` is odd? What if they don't line up as we're assuming here?

src/tasks/tasks.py
```diff
@@ -1389,6 +1393,86 @@ def collect_metrics(ix_to_tag_dict, tag_group):
...
+        self.path = path
+        self.max_seq_len = max_seq_len
+        self.n_classes = n_classes
+
+        self.train_data_text = None
+        self.val_data_text = None
+        self.test_data = None
+        self.acc_scorer = BooleanAccuracy()
+        self.gender_parity_scorer = GenderParity()
+        self.val_mtric = "%s_accuracy" % name
```
W4ngatang: typo

src/tasks/tasks.py
```diff
@@ -1389,6 +1393,86 @@ def collect_metrics(ix_to_tag_dict, tag_group):
...
+        instances = map(_make_instance, split)
+        return instances
+
+    def get_metrics(self, reset=False):
+        return {
+            "accuracy": self.acc_scorer.get_metric(reset),
+            "gender_parity": self.gender_parity_scorer.get_metric(reset),
+        }
+
+    def update_diagnostic_metrics(self, logits, labels, batch):
```
W4ngatang: where is this getting called?

main.py
```diff
@@ -558,13 +562,28 @@ def main(cl_arguments):
...
+        diagnostic_tasks = []
+        target_tasks_remove = []
+        for index, task in enumerate(target_tasks):
+            if task.name in task_modules.ALL_DIAGNOSTICS:
+                diagnostic_tasks.append(task)
+                target_tasks_remove.append(index)
+        # To avoid evaluating twice on a diagnostic task.
+        for index in sorted(target_tasks_remove, reverse=True):
+            del target_tasks[index]
+        if len(diagnostic_tasks) > 0:
```
W4ngatang: why do you need this? If `len(diagnostic_tasks) == 0` then the for loop just terminates

main.py
W4ngatang: :+1:, should it check/use `task.use_classifier`?

main.py
W4ngatang: :+1:, what's the difference between `eval` and `target_train` ?

src/tasks/tasks.py
```diff
@@ -1389,6 +1393,86 @@ def collect_metrics(ix_to_tag_dict, tag_group):
...
+        super().__init__(path, max_seq_len, name, n_classes, **kw)
+        self.path = path
+        self.max_seq_len = max_seq_len
+        self.n_classes = n_classes
+
+        self.train_data_text = None
+        self.val_data_text = None
+        self.test_data = None
+        self.acc_scorer = BooleanAccuracy()
+        self.gender_parity_scorer = GenderParity()
```
W4ngatang: do these `scorers` need to be assigned as `self.scorers` so that `self.update_metrics()` will work with them?

W4ngatang:
Shouldn't this have a specific prediction file name (in `SUPERGLUE_PRED_MAP` or something like that in `src/evaluate.py` ?

main.py
pruksmhc: Yeah I'll put this in a separate PR to do with diagnostics.

main.py
pruksmhc: Oh yes good point - will do.

main.py
pruksmhc: Actually, this should still be eval not target_train - was doing a slightly different way of testing on diagnostics before the current way. I'll revert the change. 
The difference between eval and target_task is where the function looks for checkpoints.

src/metrics/winogender_metrics.py
pruksmhc: batch_size needs to be even for winogender-diagnostic because predictions has to be even.

src/tasks/tasks.py
pruksmhc: There's also another DNC task that's dnc-recast-winogender, which is not completely supported.

src/tasks/tasks.py
pruksmhc: This gets called at src/models.py instead of update_metrics() for normal non-diagnostic tasks.

src/tasks/tasks.py
pruksmhc: Not in this case because diagnostic tasks don't call update_metrics, they call update_diagnostic_metrics()

pruksmhc:
Update @sleepinyourhat I tried putting the winogender dataset in as the RTE's test set and it is performing equally bad.

sleepinyourhat:
Thanks! To confirm, that RTE model is getting good performance on the real
RTE data? (Also worth trying the BERT++ RTE model.)

sleepinyourhat:
Also, double check that the labels are correct in the jsonl version, and
that they're in the exact same format as the RTE labels.

pruksmhc:
Yep, the RTE model is getting good performance on the real RTE data.

pruksmhc:
Also the labels are correct in the jsonl version, and I double checked that the parsing/processing is correct.

sleepinyourhat:
Wow. Even with BERT++?

sleepinyourhat:
It's just very worrying to not be able to get above chance at all... not sure what else to try.

pruksmhc:
Haven't tried with Bert++ yet, doing now.

src/evaluate.py
```diff
@@ -57,6 +57,10 @@ def evaluate(
         "sent1_str",
         "sent2_str",
         "labels",
+        "par_idx",
+        "qst_idx",
```
W4ngatang: these are duplicated and should be removed

scripts/superglue-baselines.sh
```diff
@@ -30,7 +30,7 @@ function record() {
 }
 
 function rte() {
```
W4ngatang: Could you also add `winogender-diagnostic` to the other `rte` commands? (`rte_plus` and `rte_bow`)

W4ngatang:
Aggregating some comments that I think still need to be addressed:
- Could you also rename superglue-diagnostic in this PR to make it clear that it's not the only SuperGLUE diagnostic? Coordinate with how we're describing it in the paper.
- Shouldn't this have a specific prediction file name (in SUPERGLUE_PRED_MAP or something like that in src/evaluate.py ?

src/tasks/tasks.py
```diff
@@ -1390,6 +1395,77 @@ def collect_metrics(ix_to_tag_dict, tag_group):
...
+            self.max_seq_len,
+            targ_map,
+        )
+        self.sentences = (
+            self.train_data_text[0]
+            + self.train_data_text[1]
+            + self.val_data_text[0]
+            + self.val_data_text[1]
+        )
+        log.info("\tFinished loading RTE (from SuperGLUE formatted data).")
```
W4ngatang: update

tests/test_winogender_metric.py
```diff
@@ -0,0 +1,58 @@
...
+import unittest
+import pandas as pd
+
+from src.metrics.winogender_metrics import GenderParity
+
+
+class TestWinogenderMetric(unittest.TestCase):
+    def setUp(self):
+        self.temp_dir = tempfile.mkdtemp()
+        self.path = os.path.join(self.temp_dir, "temp_ccg_dataset.tsv")
```
W4ngatang: why ccg?

src/utils/data_loaders.py
```diff
@@ -44,6 +44,18 @@ def load_span_data(tokenizer_name, file_name, label_fn=None, has_labels=True):
     return list(rows.T.to_dict().values())
 
 
+def load_jsonl(data_file, tokenizer_name, max_seq_len, targ_map):
```
W4ngatang: can you give a docstring and maybe change the function name? This is a somewhat specific version of loading from JSONL.

tests/test_winogender_metric.py
pruksmhc: Oh this actually isn't used and is copy-pasta, deleting.

src/evaluate.py
```diff
@@ -118,11 +119,11 @@ def evaluate(
         # task_preds will be a DataFrame with columns
         # ['preds'] + FIELDS_TO_EXPORT
         # for GLUE tasks, preds entries should be single scalars.
-
         # Update metrics
         task_metrics = task.get_metrics(reset=True)
         for name, value in task_metrics.items():
             all_metrics["%s_%s" % (task.name, name)] = value
+
         all_metrics["micro_avg"] += all_metrics[task.val_metric] * n_examples
```
W4ngatang: can you use the `task.contributes_micro_macro_avg` here as a check ? Also, consider renaming to something like `task.contributes_to_aggregate_score`?

src/evaluate.py
```diff
@@ -243,6 +245,7 @@ def write_preds(
     "wic": "WiC",
     "superglue-diagnostic": "AX",
     "winograd-coreference": "WSC",
+    "winogender-diagnostic": "winogender",
```
W4ngatang: switch to similar style as rest of files, i.e. `Winogender`

src/evaluate.py
pruksmhc: Yeah, that should make it clearer.

PR #720, opened 2019-06-21
A few easy checkpointing logic fixes.

sleepinyourhat:
There are other bigger issues to fix, but here are a few easy improvements:

- Allow wildcards in 'load_target_train_checkpoint', for the common case where you want to use the best pretraining epoch from some other run, but you don't know in advance which epoch that'll be.
- Fix some logic that breaks when restoring fine-tuning jobs.
- A bit more cleanup/commenting.

### DIFF
--- a/main.py
+++ b/main.py
@@ -113,7 +113,7 @@ def setup_target_task_training(args, target_tasks, model, strict):
     else:
         task_names_to_avoid_loading = []
 
-    if not args.load_target_train_checkpoint in ("none", ""):
+    if args.load_target_train_checkpoint not in ("none", ""):
         # This is to load a particular target train checkpoint.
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -180,11 +181,6 @@ def check_configurations(args, pretrain_tasks, target_tasks):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
 
     if args.load_target_train_checkpoint != "none":
-        assert_for_log(
-            os.path.exists(args.load_target_train_checkpoint),
```
pruksmhc: why are we deleting this?

sleepinyourhat: It breaks with wildcards, which we now support. I can fix it, but it'd add a fair bit of code complexity that's probably not worth it.

PR #721, opened 2019-06-21
Adding Separate Best Checkpoints for Target Train tasks

pruksmhc:
Changes:
In target task training, we no longer keep only one checkpoint path - instead, for each target task, we store in separate directories. This applies to both transfer_paradigm=frozen and finetune. 

For pretraining, we still save the checkpoints in the run_dir directory - pretraining checkpoints should not be affected by this PR.

### DIFF
--- a/main.py
+++ b/main.py
@@ -137,13 +137,16 @@ def setup_target_task_training(args, target_tasks, model, strict):
                 # We want to do target training without pretraining, thus
                 # we need to first create a checkpoint to come back to for each of
                 # the target tasks to finetune.
-                assert_for_log(
-                    args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
-                )
                 model_state = model.state_dict()
                 model_path = os.path.join(args.run_dir, "model_state_untrained_pre_target_train.th")
                 torch.save(model_state, model_path)
-            log.warning("Using untrained encoder parameters!")
+            else:
+                # If do_pretrain = 0 and transfer_paradigm=frozen, then do_eval will evaluate on 
+                # untrained encoder parameters.
+                assert_for_log(
+                    args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
+                )  
+                log.warning("Using untrained encoder parameters!")
     return task_names_to_avoid_loading
 
 
--- a/main.py
+++ b/main.py
@@ -488,7 +488,7 @@ def main(cl_arguments):
             if args.transfer_paradigm == "finetune":
                 # Reload the original best model from before target-task
                 # training since we specifically finetune for each task.
-                pre_finetune_path = get_best_checkpoint_path(args.run_dir, "pretrain")
+                pre_target_train = get_best_checkpoint_path(args.run_dir, "pretrain")
 
                 load_model_state(
                     model, pre_target_train, args.cuda, skip_task_models=[], strict=strict
@@ -529,7 +529,7 @@ def main(cl_arguments):
                         phase = "pretrain"
                     else:
                         phase = "target_train"
-                    # find the task-specific best checkpoint to evaluat eon
+                    # Find the task-specific best checkpoint to evaluate on.
                     ckpt_path = get_best_checkpoint_path(args.run_dir, phase, task.name)
 
                 assert "best" in ckpt_path
--- a/main.py
+++ b/main.py
@@ -124,37 +124,24 @@ def setup_target_task_training(args, target_tasks, model, strict):
             strict=strict,
         )
     else:
-        # Look for target train checkpoints (available only if we're restoring from a run that
-        # already finished), then look for training checkpoints.
-
-        best_path = get_best_checkpoint_path(args.run_dir, "target_train")
-        if best_path:
-            load_model_state(
-                model, best_path, args.cuda, task_names_to_avoid_loading, strict=strict
-            )
+        if args.do_pretrain == 1:
+            best_pretrain = get_best_checkpoint_path(args.run_dir, "pretrain")
+            if best_pretrain:
+                load_model_state(
+                    model, best_pretrain, args.cuda, task_names_to_avoid_loading, strict=strict
+                )
         else:
-            if args.do_pretrain == 1:
-                best_pretrain = get_best_checkpoint_path(args.run_dir, "pretrain")
-                if best_pretrain:
-                    load_model_state(
-                        model, best_pretrain, args.cuda, task_names_to_avoid_loading, strict=strict
-                    )
-            else:
+            if args.transfer_paradigm == "finetune":
+                # We want to do target training without pretraining, thus
+                # we need to first create a checkpoint to come back to for each of
+                # the target tasks to finetune.
                 assert_for_log(
                     args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
                 )
-
-                if args.transfer_paradigm == "finetune":
-                    # We want to do target training without pretraining, thus
-                    # we need to first create a checkpoint to come back to for each of
-                    # the target tasks to finetune.
-                    model_state = model.state_dict()
-                    model_path = os.path.join(
-                        args.run_dir, "model_state_untrained_pre_target_train.th"
-                    )
-                    torch.save(model_state, model_path)
-
-            log.warning("Evaluating untrained encoder parameters!")
+                model_state = model.state_dict()
+                model_path = os.path.join(args.run_dir, "model_state_untrained_pre_target_train.th")
+                torch.save(model_state, model_path)
+            log.warning("Using untrained encoder parameters!")
     return task_names_to_avoid_loading
 
 
@@ -264,23 +251,22 @@ def _run_background_tensorboard(logdir, port):
 def get_best_checkpoint_path(run_dir, phase, task_name=None):
     """ Look in run_dir for model checkpoint to load.
     Hierarchy is
-        1) best task-specific checkpoint
-        2) best checkpoint from the phase so far
-        3) if we do only target training without pretraining, then load checkpoint before 
+        1) best task-specific checkpoint for target_train, used when evaluating
+        2) if we do only target training without pretraining, then load checkpoint before 
         target training
+        3) if we're doing pretraining, then load the overall best model state
         4) nothing found (empty string) 
     """
     checkpoint = []
-    if task_name is not None:
+    if phase == "target_train":
+        assert task_name is not None, "Specify a task checkpoint to evaluate from."
         checkpoint = glob.glob(
-            os.path.join(run_dir + "/" + task_name + "/", "model_state_%s_epoch_*.best.th" % phase)
+            os.path.join(run_dir, task_name, "model_state_%s_epoch_*.best_macro.th" % phase)
         )
     if len(checkpoint) == 0:
-        checkpoint = glob.glob(
-            os.path.join(run_dir, "model_state_%s_epoch_*.best_macro.th" % phase)
-        )
-    if len(checkpoint) == 0 and phase == "target_train":
         checkpoint = glob.glob(os.path.join(run_dir, "model_state_untrained_pre_target_train.th"))
+    if len(checkpoint) == 0 and phase == "pretrain":
+        checkpoint = glob.glob(os.path.join(run_dir, "model_state_pretrain_epoch_*.best_macro.th"))
     if len(checkpoint) > 0:
         assert_for_log(len(checkpoint) == 1, "Too many best checkpoints. Something is wrong.")
         return checkpoint[0]
@@ -512,7 +498,7 @@ def main(cl_arguments):
                 )
             else:  # args.transfer_paradigm == "frozen":
                 # Load the current overall best model.
-                layer_path = get_best_checkpoint_path(args.run_dir, "target_train")
+                layer_path = get_best_checkpoint_path(args.run_dir, "target_train", task.name)
                 assert layer_path, "No best checkpoint found."
                 load_model_state(
                     model,
--- a/main.py
+++ b/main.py
@@ -464,9 +464,6 @@ def main(cl_arguments):
                 task.val_metric_decreases,
                 phase="target_train",
             )
-            import pdb
-
-            pdb.set_trace()
             _ = trainer.train(
                 tasks=[task],
                 stop_metric=task.val_metric,
--- a/main.py
+++ b/main.py
@@ -141,11 +141,11 @@ def setup_target_task_training(args, target_tasks, model, strict):
                 model_path = os.path.join(args.run_dir, "model_state_untrained_pre_target_train.th")
                 torch.save(model_state, model_path)
             else:
-                # If do_pretrain = 0 and transfer_paradigm=frozen, then do_eval will evaluate on 
+                # If do_pretrain = 0 and transfer_paradigm=frozen, then do_eval will evaluate on
                 # untrained encoder parameters.
                 assert_for_log(
                     args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
-                )  
+                )
                 log.warning("Using untrained encoder parameters!")
     return task_names_to_avoid_loading
 
@@ -544,7 +544,9 @@ def main(cl_arguments):
                 for task in pretrain_tasks:
                     ckpt_path = get_best_checkpoint_path(args.run_dir, "pretrain", task.name)
                     assert ".best" in ckpt_path
-                    load_model_state(model, ckpt_path, args.cuda, skip_task_models=[], strict=strict)
+                    load_model_state(
+                        model, ckpt_path, args.cuda, skip_task_models=[], strict=strict
+                    )
                     evaluate_and_write(args, model, [task], splits_to_write)
 
         elif args.transfer_paradigm == "frozen":
--- a/main.py
+++ b/main.py
@@ -258,12 +258,12 @@ def get_best_checkpoint_path(run_dir, phase, task_name=None):
     if phase == "target_train":
         assert task_name is not None, "Specify a task checkpoint to evaluate from."
         checkpoint = glob.glob(
-            os.path.join(run_dir, task_name, "model_state_%s_epoch_*.best_macro.th" % phase)
+            os.path.join(run_dir, task_name, "model_state_%s_epoch_*.best.th" % phase)
         )
     if len(checkpoint) == 0:
         checkpoint = glob.glob(os.path.join(run_dir, "model_state_untrained_pre_target_train.th"))
     if len(checkpoint) == 0 and phase == "pretrain":
-        checkpoint = glob.glob(os.path.join(run_dir, "model_state_pretrain_epoch_*.best_macro.th"))
+        checkpoint = glob.glob(os.path.join(run_dir, "model_state_pretrain_epoch_*.best.th"))
     if len(checkpoint) > 0:
         assert_for_log(len(checkpoint) == 1, "Too many best checkpoints. Something is wrong.")
         return checkpoint[0]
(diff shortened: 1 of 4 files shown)


main.py
```diff
@@ -495,7 +505,7 @@ def main(cl_arguments):
             if args.transfer_paradigm == "finetune":
                 # Reload the original best model from before target-task
                 # training since we specifically finetune for each task.
-                pre_target_train = get_best_checkpoint_path(args.run_dir, "pretrain")
+                pre_finetune_path = get_best_checkpoint_path(args.run_dir, "pretrain")
```
sleepinyourhat: This'll revert break a fix I recently merged, since the fix makes a change in the opposite direction. Look out when updating.

main.py
```diff
@@ -261,14 +261,24 @@ def _kill_tb_child():
...
+        3) if we do only target training without pretraining, then load checkpoint before 
         target training
-        3) nothing found (empty string) """
-    checkpoint = glob.glob(os.path.join(run_dir, "model_state_%s_epoch_*.best_macro.th" % phase))
+        4) nothing found (empty string) 
+    """
+    checkpoint = []
+    if task_name is not None:
+        checkpoint = glob.glob(
+            os.path.join(run_dir + "/" + task_name + "/", "model_state_%s_epoch_*.best.th" % phase)
```
sleepinyourhat: use join for everything, not '+'.

main.py
```diff
@@ -261,14 +261,24 @@ def _kill_tb_child():
...
+        1) best task-specific checkpoint
+        2) best checkpoint from the phase so far
+        3) if we do only target training without pretraining, then load checkpoint before 
         target training
-        3) nothing found (empty string) """
-    checkpoint = glob.glob(os.path.join(run_dir, "model_state_%s_epoch_*.best_macro.th" % phase))
+        4) nothing found (empty string) 
+    """
+    checkpoint = []
+    if task_name is not None:
```
sleepinyourhat: This doesn't seem like the right test. Even if it works for now, it'd be easy to break in future editing.
- In pretraining, we never want to do this, since there's only ever one task, and it's helpful to have a common filename standard for pretrained models.
- In target task training, we always want to do this, even if there's only one target task, since it'll make our checkpoints robust in case we come back and add more target task laters.

Let's make this depend on the phase.

main.py
```diff
@@ -261,14 +261,24 @@ def _kill_tb_child():
...
+        checkpoint = glob.glob(
+            os.path.join(run_dir + "/" + task_name + "/", "model_state_%s_epoch_*.best.th" % phase)
+        )
+    if len(checkpoint) == 0:
+        checkpoint = glob.glob(
+            os.path.join(run_dir, "model_state_%s_epoch_*.best_macro.th" % phase)
```
sleepinyourhat: Make a note that we'll only see these for 'target_task' when loading older runs.

src/trainer.py
```diff
@@ -303,6 +303,11 @@ def _setup_training(
         task_infos = {task.name: {} for task in tasks}
         for task in tasks:
             task_info = task_infos[task.name]
+            if (
+                os.path.exists(self._serialization_dir + "/" + task.name) is False
+                and phase == "target_train"
```
sleepinyourhat: At some point we should start treating this as something other than a string.

src/trainer.py
```diff
@@ -303,6 +303,11 @@ def _setup_training(
         task_infos = {task.name: {} for task in tasks}
         for task in tasks:
             task_info = task_infos[task.name]
+            if (
+                os.path.exists(self._serialization_dir + "/" + task.name) is False
```
sleepinyourhat: Use join.

sleepinyourhat: Here and below.

src/trainer.py
```diff
@@ -350,9 +355,13 @@ def _setup_training(
             task_info["last_log"] = time.time()
 
         # Metric bookkeeping
-        all_metrics = [task.val_metric for task in tasks] + ["micro_avg", "macro_avg"]
```
sleepinyourhat: This change seems like it could break quite a lot. Split it off if you can, otherwise run as many tests as you can, and read through the full trainer file for possible issues.

src/trainer.py
```diff
@@ -1016,23 +1025,34 @@ def _description_from_metrics(self, metrics):
         """ format some metrics as a string """
         return ", ".join(["%s: %.4f" % (name, value) for name, value in metrics.items()])
 
-    def _unmark_previous_best(self, phase, epoch):
+    def _unmark_previous_best(self, phase, epoch, task=""):
+        if task == "":
+            best_str = "best_macro"
```
sleepinyourhat: Simpler/better: Just use the string 'best' everywhere.

src/trainer.py
```diff
@@ -1016,23 +1025,34 @@ def _description_from_metrics(self, metrics):
...
+            best_str = "best"
         candidates = glob.glob(
-            os.path.join(self._serialization_dir, "*_state_{}_epoch_*.th".format(phase))
+            os.path.join(self._serialization_dir + task, "*_state_{}_epoch_*.th".format(phase))
         )
         for file in candidates:
             # Skip the best, because we'll need it.
             # Skip the just-written checkpoint.
-            if ".best_macro" not in file and "_{}.".format(epoch) not in file:
+            if "best" not in file and "_{}.".format(epoch) not in file:
```
sleepinyourhat: Better to keep in the '.', in case someone uses 'best' in a run name.

src/trainer.py
```diff
@@ -1041,6 +1061,8 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
         ----------
         training_state: An object containing trainer state (step number, etc.), to be saved.
         phase: Usually 'pretrain' or 'target_train'.
+        all_val_metrics If this is not none, then we loop through and save each task's _best
```
sleepinyourhat: Missing colon.

src/trainer.py
```diff
@@ -1041,6 +1061,8 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
         ----------
         training_state: An object containing trainer state (step number, etc.), to be saved.
         phase: Usually 'pretrain' or 'target_train'.
+        all_val_metrics If this is not none, then we loop through and save each task's _best 
+        in addiiton to the other checkpoints.
```
sleepinyourhat: Typo. Please proofread before sending for review.

src/trainer.py
```diff
@@ -1115,14 +1137,67 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
                 "metric_state_{}_epoch_{}{}.th".format(phase, epoch, best_str),
             ),
         )
-        log.info("Saved to %s", self._serialization_dir)
+        log.info("Saved general checkpoints to %s", self._serialization_dir)
```
sleepinyourhat: I don't follow the logic here. We always save exactly one checkpoint, either the general ones for pretraining or the task-specific one in target task training.

src/trainer.py
```diff
@@ -1115,14 +1137,67 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
+                )
+
+                torch.save(
+                    model_state,
+                    os.path.join(
+                        self._serialization_dir,
+                        task_name + "/model_state_{}_epoch_{}.best.th".format(phase, epoch),
+                    ),
+                )
+                # delte previosu best.
```
sleepinyourhat: Three typos.

src/trainer.py
```diff
@@ -1115,14 +1137,67 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
+            - epoch: int
+            - task_states: dict, dictionary with task experiment parameters such as optimizer and scheduler. 
+            - model_state: experiment model
+        
+        Returns
+        --------------------
+        None
+        """
+        for metric_name, metric_info in self._metric_infos.items():
+            # if the best checpoint is thsi epoch
```
sleepinyourhat: Typo.

src/trainer.py
```diff
@@ -1115,14 +1137,67 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
+        --------------------
+            - epoch: int
+            - task_states: dict, dictionary with task experiment parameters such as optimizer and scheduler. 
+            - model_state: experiment model
+        
+        Returns
+        --------------------
+        None
+        """
+        for metric_name, metric_info in self._metric_infos.items():
```
sleepinyourhat: I don't follow the logic here—it seems like we're marking everything as best.

src/trainer.py
```diff
@@ -1115,14 +1137,67 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
+            # if the best checpoint is thsi epoch
+            if metric_info["best"][0] == epoch and metric_info["task_name"] is not None:
+                task_name = metric_info["task_name"]
+                torch.save(
+                    metric_info,
+                    os.path.join(
+                        self._serialization_dir,
+                        task_name + "/metric_state_{}_epoch_{}.best.th".format(phase, epoch),
```
sleepinyourhat: We weren't previously saving metric/task state for target tasks, right? It's reasonable to fix that, but if we fix it here, make sure we can actually use the stuff we save.

tutorials/setup_tutorial.md
```diff
@@ -241,7 +241,7 @@ You should see something like this:
 After running this experiment, you should have in your run directory:
 
 * a checkpoint of the best model state (based on your scores)
-* a `log.log` file which contains all the logs
```
sleepinyourhat: Why delete this?

main.py
pruksmhc: What do you mean that there is only one task in pretraining? 
Also, we only pass in task_name during target_train phase (and we don't' during pretrain), but I can move that logic inside this function.

src/trainer.py
pruksmhc: We were previously saving metric/task state for target tasks.

main.py
sleepinyourhat: Er, sorry. There's only ever one model and set of checkpoints in pretraining, regardless of how many tasks we use.

I know the presence of the task variable depends on phase, but it's easier to maintain if we directly check the phase in this if statement.

src/trainer.py
pruksmhc: We only mark a task's metric_infos as best only if the validation metrics during that epoch is better than the previous validation passes.

main.py
pruksmhc: In this PR, we still have checkpoints named model_state_target_train_epoch_{}{}.best_macro.th, that store the checkpoints of models that maximize the macro_avg (which might be different than the best checkpoints for each task).

src/trainer.py
sleepinyourhat: All of the strings below specify 'best'. Where are the regular non-bests saved?

src/trainer.py
sleepinyourhat: Ah, yeah. Do we ever use them, though?

main.py
sleepinyourhat: Ah, that's broken, then—we can't and shouldn't save the best checkpoints for the individual tasks during pretraining. (They're not guaranteed to be good 'bests', since we might early stop before some tasks are well trained.) Just track a single 'best' in pretraining.

src/trainer.py
pruksmhc: They're stored in the same directory file.

sleepinyourhat: I'm asking for where in the code. The name of this method doesn't match what it does.

In any case, if this is meant for pretraining, we may not need it at all...

pruksmhc: It's here: https://github.com/nyu-mll/jiant/blob/15dd439fd9b998f136a6d8e865d43a6ec21a2c05/src/trainer.py#L1035
We unmark the previous "best" checkpoint before we save a best checkpoint

sleepinyourhat: But what if we don't hit a new best at all, it has just been 1000 steps (or similar) and we want to checkpoint anyway? In the current code, we will save such checkpoints. The title of this method makes me think that should live here, but it doesn't seem to.

main.py
```diff
@@ -126,36 +126,24 @@ def setup_target_task_training(args, target_tasks, model, strict):
         log.info("Loading existing model from %s...", best_path)
         load_model_state(model, best_path, args.cuda, task_names_to_avoid_loading, strict=strict)
     else:
-        # Look for target train checkpoints (available only if we're restoring from a run that
-        # already finished), then look for training checkpoints.
-
-        best_path = get_best_checkpoint_path(args.run_dir, "target_train")
-        if best_path:
-            # Continue target task training.
```
sleepinyourhat: Deleting intentionally? I know this may not have been fully supported, just making sure.

main.py
```diff
@@ -126,36 +126,24 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
-                )
-
+        else:
             if args.transfer_paradigm == "finetune":
                 # We want to do target training without pretraining, thus
                 # we need to first create a checkpoint to come back to for each of
                 # the target tasks to finetune.
+                assert_for_log(
+                    args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
+                )
```
sleepinyourhat: This assert should only fire if paradigm != finetune. If we're fine-tuning, there's no risk that we'll evaluate untrained parameters. So, something's off.

main.py
```diff
@@ -126,36 +126,24 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
                 model_state = model.state_dict()
                 model_path = os.path.join(args.run_dir, "model_state_untrained_pre_target_train.th")
                 torch.save(model_state, model_path)
-
-            log.warning("Evaluating untrained encoder parameters!")
+            log.warning("Using untrained encoder parameters!")
```
sleepinyourhat: We can have some kind of warning here, but this is misleading if we're doing fine tuning.

main.py
```diff
@@ -257,16 +245,25 @@ def _kill_tb_child():
     atexit.register(_kill_tb_child)
 
 
-def get_best_checkpoint_path(run_dir, phase):
+def get_best_checkpoint_path(run_dir, phase, task_name=None):
     """ Look in run_dir for model checkpoint to load.
     Hierarchy is
-        1) best checkpoint from the phase so far
-        2) if we do only target training without pretraining, then load checkpoint before
+        1) best task-specific checkpoint for target_train, used when evaluating
```
sleepinyourhat: Only when evaluating, or also when doing target-task training?

sleepinyourhat: We should make sure *not* to restore a partially-completed target_task training run unless we're sure that that works, and we restore things like step count and optimizer state appropriately.

src/trainer.py
```diff
@@ -1016,23 +1022,26 @@ def _description_from_metrics(self, metrics):
...
 
-    def _unmark_previous_best(self, phase, epoch):
+    def _unmark_previous_best(self, phase, epoch, task=""):
         marked_best = glob.glob(
-            os.path.join(self._serialization_dir, "*_state_{}_epoch_*.best_macro.th".format(phase))
+            os.path.join(
+                self._serialization_dir + task, "*_state_{}_epoch_*.best_macro.th".format(phase)
+            )
         )
+        # and now, we will recursively elete all of these ones.
```
sleepinyourhat: Typo, capitalization.

src/trainer.py
```diff
@@ -1016,23 +1022,26 @@ def _description_from_metrics(self, metrics):
         """ format some metrics as a string """
         return ", ".join(["%s: %.4f" % (name, value) for name, value in metrics.items()])
 
-    def _unmark_previous_best(self, phase, epoch):
+    def _unmark_previous_best(self, phase, epoch, task=""):
         marked_best = glob.glob(
-            os.path.join(self._serialization_dir, "*_state_{}_epoch_*.best_macro.th".format(phase))
+            os.path.join(
+                self._serialization_dir + task, "*_state_{}_epoch_*.best_macro.th".format(phase)
```
sleepinyourhat: There are still a few instances of best_macro floating around. We should only ever write ".best" as a filename, and if we do want to keep reading/searching for ".best_macro" as a backward compatibility measure (optional), we should also read ".best".

src/trainer.py
```diff
@@ -1016,23 +1022,26 @@ def _description_from_metrics(self, metrics):
...
             # Skip the just-written checkpoint.
             if "_{}.".format(epoch) not in file:
-                os.rename(file, re.sub("%s$" % ".best_macro.th", ".th", file))
+                os.rename(file, re.sub("%s$" % (".best_macro.th"), ".th", file))
 
-    def _delete_old_checkpoints(self, phase, epoch):
+    def _delete_old_checkpoints(self, phase, epoch, task=""):
         candidates = glob.glob(
-            os.path.join(self._serialization_dir, "*_state_{}_epoch_*.th".format(phase))
+            os.path.join(self._serialization_dir + task, "*_state_{}_epoch_*.th".format(phase))
```
sleepinyourhat: use join, not +. Necessary for platform future-proofing.

src/trainer.py
```diff
@@ -1056,6 +1065,7 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
         else:
             best_str = ""
 
+        # Save the overall best model for this target train in terms of macro/micro_avg score.
```
sleepinyourhat: Seems misleading, since we'll also run the below line when this checkpoint isn't the best, and best_str == "".

src/trainer.py
```diff
@@ -1094,35 +1095,105 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
-        torch.save(
-            metric_states,
-            os.path.join(
-                self._serialization_dir,
-                "metric_state_{}_epoch_{}{}.th".format(phase, epoch, best_str),
-            ),
-        )
-        log.info("Saved to %s", self._serialization_dir)
+
+        if phase == "pretrain":
```
sleepinyourhat: It'd be way more readable if we split this into two methods, one (new) method for pretraining/mtl, and one (that we have now) for target tasks.

src/trainer.py
sleepinyourhat: I think I see now—this looks reasonable as-is.

src/trainer.py
pruksmhc: I intentionally put in best_macro as the variable name because what we are doing is tracking macro averages (which if you have only one task, it is equal to tracking its performance). If there's a strong reason why you think .best > .best_macro, I can change it

sleepinyourhat: Yeah, we can get more votes from others if you want, but I think 'best' is a lot clearer. It's more standard, and most users likely won't be doing MTL pretraining.

main.py
pruksmhc: Yes, this was not fully supported, and has to be changed now that we're saving our target tasks in multiple places depending on task.

main.py
pruksmhc: Ah, alright - so this assert is for evaluation.

main.py
pruksmhc: This is only used during evaluation, and is a retrieval type function. During target task training, we store the parameters at each epoch for each task being target trained, which is retrieved here.

src/trainer.py
pruksmhc: Since they share some overlap, splitting the parts after this into two functions.

main.py
sleepinyourhat: SGTM!

src/trainer.py
sleepinyourhat: Use join to join _serialization_dir and task.

tests/test_checkpointing.py
```diff
@@ -165,6 +167,15 @@ def __init__(self, requires_grad):
...
+            test_trainer, train_params, opt_params, schd_params = trainer.build_trainer(
+                self.args,
+                ["wic"],  # here, we use WIC twice to reduce the amount of boiler-plate code
+                model,
+                self.args.run_dir,
+                self.wic.val_metric_decreases,
+                phase="pretrain",
+            )
+
             test_trainer, train_params, opt_params, schd_params = trainer.build_trainer(
```
sleepinyourhat: You initialize these variable names twice. Disambiguate or use _ if you really don't need the first set.

Issue #719, opened 2019-06-21
We log weighting strategy during target task training when it's irrelevant


sleepinyourhat:
Another logging line to cut, at least during target task training:
"06/21 06:37:13 PM: Using weighting method: proportional, with normalized sample weights [1.] "

PR #722, opened 2019-06-21
No more weight logging in target

sleepinyourhat:
This is a bit messy, but still hopefully an improvement. This cuts out some of the logic and logging for multitask training when there's only one task to train on.


sleepinyourhat:
Resolves #719.

PR #734, opened 2019-06-24
Fixing MNLI evaluation 

pruksmhc:
Changes:
-Adding change to parse "n/a" as a string 
-Adding change to write_preds to get correct file path to write preds to


src/evaluate.py
```diff
@@ -511,21 +511,15 @@ def _add_default_column(df, name: str, val):
         _apply_pred_map(preds_df, pred_map, "prediction")
         _write_preds_with_pd(
             preds_df.iloc[:9796],
-            os.path.join(
-                pred_dir, _get_pred_filename("mnli-m", pred_dir, split_name, strict_glue_format)
-            ),
+            _get_pred_filename("diagnostic", pred_dir, split_name, strict_glue_format),
```
sleepinyourhat: Thy do all three of these lines say 'diagnostic' now? What changed?

Issue #748, opened 2019-06-25
Fixing evaluation (Especially when restoring from a previous checkpoint 


pruksmhc:
If you want to evaluate on a previous checkpoint without training it further (do_pretrain = 0, do_target_train = 0, do_eval = 1) with args.load_target_train_checkpoint, it doesn't find the correct checkpoint to evaluate on.

Issue #741, opened 2019-06-24
do_eval is broken with args.load_target_train_checkpoint


sleepinyourhat:
We only save the special model_state_untrained_pre_target_train.th checkpoint if there is no load_target_train_checkpoint, but we look for it regardless at the end of the target task training section in main.

save: https://github.com/nyu-mll/jiant/blob/master/main.py#L142

load: https://github.com/nyu-mll/jiant/blob/master/main.py#L525

I can't really figure out what the code around line 525 is meant to do, so I'm not sure what direction is best for a fix. @pruksmhc, @W4ngatang – does this look familiar to either of you?

pruksmhc:
The code around 525 is because for finetuning each task, we want to start off with a fresh copy of BERT. If we pretrain first, the fresh copy would be the weights of the best performing BERT weights in pretraining stage, else it would be model_state_untrained_pre_target_train.th that we save previously.

sleepinyourhat:
Ah. We're doing it one extra time, then, since we do it after every task, even the last. We should fix the path finding and also fix that, since each save/load is a ton of file i/o. Do you have bandwidth to take this on?

pruksmhc:
I'm focusing on winogender and a few other things for the first part of today, and can take this on after. Feel free to take this one if you have time before then!

sleepinyourhat:
Sounds good.

sleepinyourhat:
I took a look at this and it's bigger than I can really take on today/tomorrow. I think we'd need to:
- Add some logic at the start of the pretraining section to identify what checkpoint to load.
- Save model_state_untrained_pre_target_train.th iff no checkpoint is available.
- Add logic to decide when we actually need to reload the checkpoint.
- Rewrite/test the eval checkpoint loading system to ensure that it still properly supports

sleepinyourhat:
Sorry!

pruksmhc:
What is the thinking behind the checking logic at the start of the pretraining section? Don't we always want to train from a fresh start? (unless we have something like load_pretrain_checkpoint).

sleepinyourhat:
The idea is that most of the code/logic only needs to happen once:
- Deciding whether to save a model_state_untrained_pre_target_train.th checkpoint.
- Deciding what checkpoint (if any) to load at the start of target task training.
- Deciding whether we need to reload that checkpoint for future tasks. (I think this is basically just checking transfer_paradigm == finetune.)

[...]

PR #752, opened 2019-06-25
Fixing Target Train 

pruksmhc:
Target train logic + checkpointing surrounding reloading checkpoints and also frozen and finetuning is quite confusing. 
This PR makes some major changes, including: 
-Getting rid of task_names_to_avoid_loading
Refer to this comment thread on the thinking behind this:https://github.com/nyu-mll/jiant/issues/741
-Getting rid of the distinction between paradigm=frozen and finetune. The only difference is that the sentence encoder is trainable in one and not in the other in target train phase. However, the checkpointing logic is now the same between the two cases (each task is saved in its own separate directory) to reduce code complexity. 
-In evaluation phase, we introduce args.load_eval_checkpointing for the case where do_pretrain = 0 and do_target_train = 0 and do_full_eval = 1. 

This refers to:
https://github.com/nyu-mll/jiant/pull/752/
https://github.com/nyu-mll/jiant/issues/748


main.py
```diff
@@ -456,37 +500,13 @@ def main(cl_arguments):
...
-                if "scalar_mix" in n and "scalar_mix_0" not in n
-            ]
-            # Fails when sep_embs_for_skip is 0 and elmo_scalars has nonzero
-            # length.
-            assert_for_log(
-                not elmo_scalars or args.sep_embs_for_skip,
-                "Error: ELMo scalars loaded and will be updated in do_target_task_training but "
-                "they should not be updated! Check sep_embs_for_skip flag or make an issue.",
-            )
+        pre_target_train_model_state = setup_target_task_training(args, target_tasks, model, strict)
```
pruksmhc: Note: We now only load the checkpoint to return back to once instead of every time for each task.

config/defaults.conf
```diff
@@ -93,6 +93,9 @@ transfer_paradigm = "frozen" // How to use pretrained model parameters during ta
 load_target_train_checkpoint = none  // If not "none", load the specified model_state checkpoint
                                      // file when starting do_target_task_training.
                                      // Supports * wildcards.
+load_eval_checkpoint = none  // If not "none", load the specified model_state checkpoint
+                              // file when doing do_full_eval
+
```
sleepinyourhat: Should also support wildcards. If we wind up with the glob-then-assert-then-assert logic in more than two places, it might be worth it to make it a utility method.

main.py
```diff
@@ -148,7 +152,40 @@ def setup_target_task_training(args, target_tasks, model, strict):
                     args.allow_untrained_encoder_parameters, "No best checkpoint found to evaluate."
                 )
                 log.warning("Using untrained encoder parameters!")
-    return task_names_to_avoid_loading
+
+        # The version of the model before target task training for the new task.
+        pre_target_train_model = copy.deepcopy(model)
```
sleepinyourhat: This looks expensive, and it might increase GPU memory usage. Why the deep copy?

sleepinyourhat: I think it's cleanest to just make sure there's a checkpoint we can restore to, and passing around a path to that checkpoint.

main.py
```diff
@@ -89,7 +89,11 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
     loads the correct model state for the target task training
-    stage.
+    stage. 
+    The logic for loading the correct model state for target task training is:
+    1) If load_target_train_checkpoint is used, then load the weights from that checkpoint.
+    2) If we did pretraining, then load the best model from pretraining. 
+    3) Default case: We save untrained encoder weights.
 
     Parameters
     ----------------
```
sleepinyourhat: We should cut task_names_to_avoid_loading here, and generally scrap most of the existing code in this method.

main.py
```diff
@@ -148,7 +152,40 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
+            ]
+            # Fails when sep_embs_for_skip is 0 and elmo_scalars has nonzero
+            # length.
+            assert_for_log(
+                not elmo_scalars or args.sep_embs_for_skip,
+                "Error: ELMo scalars loaded and will be updated in do_target_task_training but "
+                "they should not be updated! Check sep_embs_for_skip flag or make an issue.",
+            )
+
+        if args.transfer_paradigm == "finetune":
```
sleepinyourhat: Merge this with the above. Should only need one if/else for this whole section of code.

main.py
sleepinyourhat: Probably too expensive. :/ 

Suggested logic:
- setup returns a path
- when training on the first task (enumerate and test i == 0), load the path
- when training on later tasks, load the path only if paradigm == finetune

main.py
```diff
@@ -501,89 +513,45 @@ def main(cl_arguments):
...
-            # Don't do any special checkpointing logic here
-            # since model already has all the trained task specific modules.
-            evaluate_and_write(args, model, target_tasks, splits_to_write)
+                    assert len(checkpoint) == 1, (
+                        "Too many checkpoints match pattern: %r" % args.load_eval_checkpoint
+                    )
+                    ckpt_path = checkpoint[0]
+                else:
+                    ckpt_path = get_best_checkpoint_path(args.run_dir, "target_train", task.name)
+                assert ".best" in ckpt_path
```
sleepinyourhat: Might be cleaner to fold this into the get_best_checkpoint_path method, even if it means another keyword arg.

main.py
```diff
@@ -501,89 +513,45 @@ def main(cl_arguments):
...
-                # then evaluate on pretraining checkpoints.
-                for task in pretrain_tasks:
-                    ckpt_path = get_best_checkpoint_path(args.run_dir, "pretrain", task.name)
-                    assert ".best" in ckpt_path
-                    load_model_state(
-                        model, ckpt_path, args.cuda, skip_task_models=[], strict=strict
+        if args.do_target_task_training:
+            for task in target_tasks:
+                # Find the task-specific best checkpoint to evaluate on.
+                if args.load_eval_checkpoint:
```
sleepinyourhat: We could fold this into the get_best_checkpoint_path logic by including a 'user_specified' flag. If it's not None or 'none' or '', then assert that it exists and return it before doing anything else.

main.py
pruksmhc: Actually, I realize that we should still keep task_names_to_avoid_loading since we want to avoid loading the task-specific modules before we finetune it.

.circleci/config.yml
```diff
@@ -29,7 +29,10 @@
               python scripts/download_glue_data.py --data_dir data --tasks all
               source activate jiant
               python -m nltk.downloader perluniprops  nonbreaking_prefixes punkt
-              python main.py --config_file config/demo.conf
+              python main.py --config_file config/demo.conf --overrides "transfer_paradigm=finetune"
```
pruksmhc: Ignore these changes - this is just a way to automatically test that all of these cases pass. This will not be pushed to master.

sleepinyourhat: Sounds good. I made a parallel proposal to add one of these cases permanently.

config/defaults.conf
```diff
@@ -93,6 +93,10 @@ transfer_paradigm = "frozen" // How to use pretrained model parameters during ta
 load_target_train_checkpoint = none  // If not "none", load the specified model_state checkpoint
                                      // file when starting do_target_task_training.
                                      // Supports * wildcards.
+load_eval_checkpoint = none  // If not "none", load the specified model_state checkpoint
+                            // file when doing do_full_eval
```
sleepinyourhat: Wonky indentation.

sleepinyourhat: And punctuation—final period.

main.py
```diff
@@ -99,11 +104,9 @@ def setup_target_task_training(args, target_tasks, model, strict):
 
     Returns
     ----------------
-    task_names_to_avoid_loading: list of strings, if we don't allow for
```
sleepinyourhat: Don't we still need to return it? If we're loading the same checkpoint here and in reload_model, we'll need to do the same filtering.

main.py
```diff
@@ -115,41 +118,29 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
-        )
-        assert len(checkpoint) == 1, (
-            "Too many checkpoints match pattern: %r" % args.load_target_train_checkpoint
-        )
-        best_path = checkpoint[0]
-        log.info("Loading existing model from %r...", best_path)
-        load_model_state(model, best_path, args.cuda, task_names_to_avoid_loading, strict=strict)
+        model_path = get_best_checkpoint_path(args, "target_train")
+        log.info("Loading existing model from %r...", model_path)
+        load_model_state(model, model_path, args.cuda, task_names_to_avoid_loading, strict=strict)
```
sleepinyourhat: Why do we have the actual load happen both here and in reload? Why not just do setup once, and then reload at the start of training for each task.

main.py
```diff
@@ -115,41 +118,29 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
-                assert_for_log(
-                    args.allow_untrained_encoder_parameters,
-                    "No best checkpoint found to evaluate. Set `allow_untrained_encoder_parameters` if you really want to use an untrained encoder.",
-                )
-                log.warning("Using untrained encoder parameters!")
-    return task_names_to_avoid_loading
+            # We want to do target training without pretraining, thus
+            # we need to first create a checkpoint to come back to for each of
+            # the target tasks to finetune.
+            model_state = model.state_dict()
```
sleepinyourhat: Could just fold this into the save line below.

main.py
```diff
@@ -115,41 +118,29 @@ def setup_target_task_training(args, target_tasks, model, strict):
...
+            assert_for_log(
+                args.allow_untrained_encoder_parameters,
+                "No best checkpoint found to evaluate. Set `allow_untrained_encoder_parameters` if you really want to use an untrained encoder.",
+            )
```
sleepinyourhat: Won't this fire now even if we're fine-tuning BERT? This was under a conditional before.

main.py
```diff
@@ -263,25 +254,42 @@ def _kill_tb_child():
     atexit.register(_kill_tb_child)
 
 
-def get_best_checkpoint_path(run_dir, phase, task_name=None):
+def get_best_checkpoint_path(args, phase, task_name=None):
     """ Look in run_dir for model checkpoint to load.
     Hierarchy is
-        1) best task-specific checkpoint for target_train, used when evaluating
+        1) user-specified target task checkpoint
```
sleepinyourhat: Write out the full logic, including which of these cases are phase-specific. If the logic looks too hairy, try to simplify the logic?

sleepinyourhat: As-is, this would make me suspect that if I supplied target and eval checkpoints, we'd load the target one regardless of the phase.

main.py
```diff
@@ -263,25 +254,42 @@ def _kill_tb_child():
...
+
     """
     checkpoint = []
-    if phase == "target_train":
-        assert task_name is not None, "Specify a task checkpoint to evaluate from."
-        checkpoint = glob.glob(
-            os.path.join(run_dir, task_name, "model_state_%s_epoch_*.best.th" % phase)
+    if phase == "target_train" and args.load_target_train_checkpoint not in ("none", ""):
+        checkpoint = glob.glob(args.load_target_train_checkpoint)
+        assert len(checkpoint) > 0, (
```
sleepinyourhat: Why do we have this here but not in the next conditional?

main.py
```diff
@@ -263,25 +254,42 @@ def _kill_tb_child():
...
+            "Specified load_target_train_checkpoint not found: %r"
+            % args.load_target_train_checkpoint
         )
-    if len(checkpoint) == 0:
-        checkpoint = glob.glob(os.path.join(run_dir, "model_state_untrained_pre_target_train.th"))
-    if len(checkpoint) == 0 and phase == "pretrain":
-        checkpoint = glob.glob(os.path.join(run_dir, "model_state_pretrain_epoch_*.best.th"))
+
+    if phase == "pretrain":
```
sleepinyourhat: Why this condition? The simplest case (the one in demo.py) is to do both pretraining and fine-tuning, and if we're doing that, we should be loading this checkpoint when starting fine-tuning.

main.py
```diff
@@ -409,6 +416,62 @@ def check_arg_name(args):
...
 
 
+def reload_model(args, ckpt_path, model, index, strict, task):
+    """
+        Function that reloads model if necesary and extracts trainable part 
+        of the model in preparation for target_task training. 
+        It only reloads model after the first task is trained, and only 
+        for finetuning. 
+
+        Parameters
```
sleepinyourhat: These types aren't that helpful, and if we're just going to include types, we should actually write them out in Python in the function signature. Instead, can we say what these params mean? Index is non-obvious.

main.py
```diff
@@ -409,6 +416,62 @@ def check_arg_name(args):
...
+        -------------------
+        args: config.Param object 
+        ckpt_path: str
+        model: MultiTaskmodel object
```
sleepinyourhat: Caps.

main.py
```diff
@@ -409,6 +416,62 @@ def check_arg_name(args):
...
+        Returns 
+        -------------------
+        to_train: List of tuples of (name, weight) for trainable parameters
+
+    """
+    if index > 0 and args.transfer_paradigm == "finetune":
+        # Only reload after target_train for first task, and transfer_paradigm=finetune.
+        load_model_state(model, ckpt_path, args.cuda, skip_task_models=[task.name], strict=strict)
+
+    if args.transfer_paradigm == "frozen":
```
sleepinyourhat: Does this really belong in reload_tasks? Seems like it'd make more sense to include in the setup fn.

main.py
```diff
@@ -409,6 +416,62 @@ def check_arg_name(args):
...
+        # Fails when sep_embs_for_skip is 0 and elmo_scalars has nonzero
+        # length.
+        assert_for_log(
+            not elmo_scalars or args.sep_embs_for_skip,
+            "Error: ELMo scalars loaded and will be updated in do_target_task_training but "
+            "they should not be updated! Check sep_embs_for_skip flag or make an issue.",
+        )
+
+    if args.transfer_paradigm == "finetune":
+        model_state = load_model_state(
```
sleepinyourhat: We already did this above.

main.py
```diff
@@ -470,37 +533,17 @@ def main(cl_arguments):
...
-            # length.
-            assert_for_log(
-                not elmo_scalars or args.sep_embs_for_skip,
-                "Error: ELMo scalars loaded and will be updated in do_target_task_training but "
-                "they should not be updated! Check sep_embs_for_skip flag or make an issue.",
-            )
-        for task in target_tasks:
+        pre_target_train_path = setup_target_task_training(args, target_tasks, model, strict)
+
+        for index in range(len(target_tasks)):
```
sleepinyourhat: for index, task in enumerate(target_tasks):

This lets you skip the extra indexing line.

main.py
```diff
@@ -515,72 +558,34 @@ def main(cl_arguments):
                 batch_size=args.batch_size,
                 weighting_method=args.weighting_method,
                 scaling_method=args.scaling_method,
-                train_params=to_train,
+                train_params=pre_target_train_model_state,
```
sleepinyourhat: This doesn't seem right. Isn't this supposed to be a list of parameters, not a model state?

sleepinyourhat: Might just be a naming issue.

main.py
sleepinyourhat: Also, see the comment in the other PR: Basically, when we're starting a phase, we should always check for checkpoints from the previous phase. That way, we can rerun evaluation for a job by simply setting do_target_train = 0 and do_full_eval = 1. Pretty common, for example, in cases where we have a new eval or analysis file to use.

tests/test_checkpointing.py
```diff
@@ -55,7 +55,7 @@ def build_trainer_params(args, task_names, phase="pretrain"):
     }
 
 
-class TestCheckpionting(unittest.TestCase):
+class TestCheckpointing(unittest.TestCase):
```
sleepinyourhat: 😃

tests/test_checkpointing.py
```diff
@@ -232,7 +232,27 @@ def __init__(self, requires_grad):
...
-            file = open(file_path, "rb")
-            assert file
+    def test_get_best_checkpointing(self):
+        self.args.load_target_train_checkpoint = os.path.join(self.temp_dir, "target_checkpoint")
+        self.args.load_eval_checkpoint = ""
+        open(self.args.load_target_train_checkpoint, "wb").close()
+        target_ckpt = get_best_checkpoint_path(self.args, phase="target_train", task_name=None)
+        assert target_ckpt == self.args.load_target_train_checkpoint
+
+        # Load from teh best target train phase checkpoint.
```
sleepinyourhat: New typo

tests/test_checkpointing.py
```diff
@@ -232,7 +232,27 @@ def __init__(self, requires_grad):
             )
             assert len(pretrain_best_checkpoints) == 1
 
-        def test_does_produce_results(self):
-            file_path = "~/repo/sample_run/jiant-demo/results.tsv"
-            file = open(file_path, "rb")
-            assert file
+    def test_get_best_checkpointing(self):
```
sleepinyourhat: Update as we update the logic.

main.py
pruksmhc: That is exactly what is being done here - we do setup once, and then we reload at the start of each task. We don't reload at index = 0 since we already do that once with setup. 
Actually, I just realized that we could just avoid loading in the setup_target_tasks function and just keep it in reload_tasks

pruksmhc: That would make the code much simpler and there wouldn't be the weird index = 0 type logic

main.py
pruksmhc: If we're finetuning BERT and we don't have pretraining, wouldn't we want to make sure the user is extra  aware of that by switching allow_untrained_encoder_parameters to equal 1?

main.py
pruksmhc: This is necessary in reload_tasks because every time you reload, you need to re-calculate which parameters to pass into the Trainer.

main.py
sleepinyourhat: I don't think so. I expect that >50% of the usage of jiant will just be finetuning BERT, or training and evaluating PRPN, or other setups without a pretraining phase.

sleepinyourhat: If we add a warning, let's add it only in the case where there are pretraining tasks specified, but do_pretrain isn't set, and there are no pretraining checkpoints. For the use cases I described, people would leave pretrain_tasks blank.

main.py
pruksmhc: This is different - I realize it's a bit confusing so I'm going to collapse the above into the paradigm=frozen case.

main.py
pruksmhc: Yep, it's a list of parameters - will change the name.

tests/test_checkpointing.py
pruksmhc: 👍

main.py
sleepinyourhat: Ah, got it. I didn't realize that these were reused later.

main.py
pruksmhc: Good point- i'll put this under target_train phase instead

main.py
pruksmhc: I actually realize we don't need it. In the load_model_for_target_train_run function we can just pass in the specific task we're finetuning on into the skip_task_models param because we're only going to be using the task specific module for that task in that specifc target task run anyways.

.circleci/config.yml
```diff
@@ -29,6 +29,9 @@
               python scripts/download_glue_data.py --data_dir data --tasks all
               source activate jiant
               python -m nltk.downloader perluniprops  nonbreaking_prefixes punkt
```
sleepinyourhat: Cut?

main.py
```diff
@@ -87,9 +87,13 @@ def handle_arguments(cl_arguments):
 
 def setup_target_task_training(args, target_tasks, model, strict):
     """
-    Saves model states from pretraining if applicable, and
-    loads the correct model state for the target task training
-    stage.
+    Saves model states from pretraining if applicable,
```
sleepinyourhat: This can probably be two lines.

main.py
```diff
@@ -263,25 +232,43 @@ def _kill_tb_child():
...
-        3) if we're doing pretraining, then load the overall best model state
-        4) nothing found (empty string)
+def get_best_checkpoint_path(args, phase, task_name=None):
+    """ Look in run_dir for model checkpoint to load when setting up for 
+    phase = target_train or phase = eval.
+    Hierarchy is: 
+        If phase == target_train:
+            1) user-specified target task checkpoint
+            2) best task-specific checkpoint from pretraining stage
+        if phase == eval:
```
pruksmhc: Nit, capitalize if, will fix

sleepinyourhat: Be consistent about if vs If.

main.py
```diff
@@ -409,6 +395,53 @@ def check_arg_name(args):
...
+def load_model_for_target_train_run(args, ckpt_path, model, strict, task):
+    """
+        Function that reloads model if necessary and extracts trainable parts 
+        of the model in preparation for target_task training. 
+        It only reloads model after the first task is trained. 
+
+        Parameters
+        -------------------
+        args: config.Param object,
+        ckpt_path: str,
```
pruksmhc: add description for ckpt_path

main.py
```diff
@@ -409,6 +395,53 @@ def check_arg_name(args):
...
+
+        Returns 
+        -------------------
+        to_train: List of tuples of (name, weight) of trainable parameters
+
+    """
+
+    if args.transfer_paradigm == "finetune":
+        load_model_state(model, ckpt_path, args.cuda, skip_task_models=[task.name], strict=strict)
+        # Train both the task specific models as well as sentence
```
sleepinyourhat: Make this one line.

main.py
```diff
@@ -470,37 +503,16 @@ def main(cl_arguments):
...
-                # Train both the task specific models as well as sentence
-                # encoder.
-                to_train = [(n, p) for n, p in model.named_parameters() if p.requires_grad]
-            else:  # args.transfer_paradigm == "frozen":
-                # Only train task-specific module
-                pred_module = getattr(model, "%s_mdl" % task.name)
-                to_train = [(n, p) for n, p in pred_module.named_parameters() if p.requires_grad]
-                to_train += elmo_scalars
-
+            pre_target_train_model_params = load_model_for_target_train_run(
```
sleepinyourhat: This suggests that it could include all the params. How about 'params_to_train'?

main.py
```diff
@@ -515,72 +527,36 @@ def main(cl_arguments):
...
-            # since model already has all the trained task specific modules.
-            evaluate_and_write(args, model, target_tasks, splits_to_write)
+        if args.do_target_task_training or (args.load_model and not args.do_pretrain):
+            # If we either do target task training, or if we only evaluate
+            # without pretraining or target task training
+            # then we evaluate on the target tasks.
+            for task in target_tasks:
+                # Find the task-specific best checkpoint to evaluate on.
+                ckpt_path = get_best_checkpoint_path(args, "eval", task.name)
+                assert ckpt_path is not None and ".best" in ckpt_path
```
sleepinyourhat: The .best clause here will fail if someone manually specifies a non-best checkpoint, even though there are legitimate use cases for that. Just assert that the path is not none.

main.py
```diff
@@ -515,72 +527,36 @@ def main(cl_arguments):
...
+                load_model_state(model, ckpt_path, args.cuda, skip_task_models=[], strict=strict)
+
+                evaluate_and_write(args, model, [task], splits_to_write)
+        elif args.do_pretrain:
+            # If args.do_target_task_training = 0 and args.do_pretrain = 1
+            # then evaluate on pretraining checkpoints.
+            for task in pretrain_tasks:
+                ckpt_path = get_best_checkpoint_path(args, "target_train", task.name)
```
sleepinyourhat: I think it's worth revising the arguments to the get_best_checkpoint_path method, both for clarity and to fix a bug: Passing in the wrong phase name feels like a hack, and as-is, this means that if you turn off do_target_task_training, then load_eval_checkpoint will stop working.

tests/test_checkpointing.py
```diff
@@ -232,7 +232,34 @@ def __init__(self, requires_grad):
...
```
sleepinyourhat: Add a brief comment saying what this is doing. It's not immediately obvious that this is just testing the flag/path logic—this method name suggests that we might actually be loading stuff.

main.py
pruksmhc: What do you mean by if do_target_task_training is off, then load_eval_checkpoint will not work? load_eval_checpoint is checked in a different logic branch in get_best_checkpoint than target training.

pruksmhc: I actually realize you don't have to call get_best_checkpoint here for simplicity.

main.py
sleepinyourhat: ?

main.py
```diff
@@ -409,6 +395,52 @@ def check_arg_name(args):
...
+def load_model_for_target_train_run(args, ckpt_path, model, strict, task):
+    """
+        Function that reloads model if necessary and extracts trainable parts 
+        of the model in preparation for target_task training. 
+        It only reloads model after the first task is trained. 
+
+        Parameters
+        -------------------
+        args: config.Param object,
+        ckpt_path: str:  path to reload model from,
```
sleepinyourhat: Extra space

tests/test_checkpointing.py
```diff
@@ -232,7 +232,36 @@ def __init__(self, requires_grad):
...
+        """
+        Testing the get_best_checkpointing function for logic surrounding path logic.
```
sleepinyourhat: cut 'logic surrounding' :D

PR #742, opened 2019-06-25
Cleaning up src/modules/modules.py into several files.

Narsil:
Fixes #420

PR #753, opened 2019-06-26
Modules cleanup

Narsil:
2nd version of #742 
This was checked with flake8 for unused + missing imports.


sleepinyourhat:
Thanks! I tried a few more test runs and everything works so far. 

I'm a novice at CircleCI setup. We can try to put something together, but if you see an easy change to make, go ahead and make the PR. I can help get the checks passing once they're in place.

Issue #750, opened 2019-06-25
Add a test to cover an eval-only run.


sleepinyourhat:
Running do_full_eval = 1 with the other two phases off brings up some unique/wonky checkpointing situations. If there's an easy way to run demo.conf twice in the automatic tests, once in this eval-only setup, and test for crashes, that could help.

PR #763, opened 2019-06-26
Add an eval-only run to CI

sleepinyourhat:
Addresses #750. This is essentially an integration test for some of our checkpointing logic.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,7 @@
               source activate jiant
               python -m nltk.downloader perluniprops  nonbreaking_prefixes punkt
               python main.py --config_file config/demo.conf
+              python main.py --config_file config/demo.conf --overrides "do_pretrain = 0, do_target_task_training = 0, load_model = 1"
         # Step 4: run tests
         - run:
             name: run tests


.circleci/config.yml
```diff
@@ -30,6 +30,7 @@
               source activate jiant
               python -m nltk.downloader perluniprops  nonbreaking_prefixes punkt
               python main.py --config_file config/demo.conf
+              python main.py --config_file config/demo.conf --overrides "do_pretrain = 0, do_target_task_training = 0, load_model = 1"
```
pruksmhc: Don't we need to set load_target_train_checkpoint (or with the new change in the Fixing Target Train PR) load_eval_checkpoint here?

sleepinyourhat: Ah, yeah. This is working as is, but producing at-chance results. 

I think that's actually a bug: If there are checkpoints for saved target task models, we should load them when starting evaluation. That was the historic behavior, and I think it's the most user-friendly. Can we try to address that in the final version of the checkpointing rewrite PR?

pruksmhc: Ah got it - so we should automatically load the best target train without having to specify it. 
I think this is addressed in the checkpointing rewrite PR right now.

Issue #760, opened 2019-06-26
WiC broken in demo.conf


sleepinyourhat:
WiC works in superglue-bow.conf, but gives strange errors in demo.conf. It should either run in this setup, or else give a helpful error that tells you what needs to be set:

```
python main.py --config config/demo.conf --overrides "pretrain_tasks = wic"
```

```
06/26 02:48:09 PM: Beginning training with stopping criteria based on metric: wic_accuracy
06/26 02:48:09 PM: Fatal error in main():
Traceback (most recent call last):
  File "main.py", line 590, in <module>
    main(sys.argv[1:])
  File "main.py", line 462, in main
    phase="pretrain",
  File "/Users/srbowman/jiant/src/trainer.py", line 594, in train
    output_dict = self._forward(batch, task=task)
  File "/Users/srbowman/jiant/src/trainer.py", line 1035, in _forward
    model_out = self._model.forward(task, tensor_batch)
  File "/Users/srbowman/jiant/src/models.py", line 808, in forward
    out = self._pair_sentence_forward(batch, task, predict)
  File "/Users/srbowman/jiant/src/models.py", line 936, in _pair_sentence_forward
    logits = classifier(sent1, sent2, mask1, mask2, [batch["idx1"]], [batch["idx2"]])
  File "/Users/srbowman/miniconda3/envs/jiant/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/Users/srbowman/jiant/src/modules/simple_modules.py", line 167, in forward
    s2_ctx_emb = s2.gather(dim=1, index=idx)
RuntimeError: Invalid index in gather at ../aten/src/TH/generic/THTensorEvenMoreMath.cpp:456
Traceback (most recent call last):
  File "main.py", line 601, in <module>
    raise e  # re-raise exception, in case debugger is attached.
  File "main.py", line 590, in <module>
    main(sys.argv[1:])
  File "main.py", line 462, in main
    phase="pretrain",
  File "/Users/srbowman/jiant/src/trainer.py", line 594, in train
    output_dict = self._forward(batch, task=task)
  File "/Users/srbowman/jiant/src/trainer.py", line 1035, in _forward
    model_out = self._model.forward(task, tensor_batch)
  File "/Users/srbowman/jiant/src/models.py", line 808, in forward
    out = self._pair_sentence_forward(batch, task, predict)
[...]

PR #764, opened 2019-06-26
WiC v2

W4ngatang:
Adapt WiC task to use detokenized data. Includes the following changes:

- modify the WiC preprocessing to track the marked word, to ensure we extract the span correponding to the entire word (in cases where we use subword tokenization)
- modify classifiers to extract representations corresponding to a (padded/masked) list of sequence element indices (previously we just took one word)


W4ngatang:
Still need to address #760 but ready for review in meantime @sleepinyourhat

sleepinyourhat:
Taking a look now.

src/modules/modules.py
```diff
@@ -444,19 +444,29 @@ def __init__(self, pooler, classifier):
 
     def forward(self, sent, mask, idxs=[]):
         """ Assumes batch_size x seq_len x d_emb """
-        emb = self.pooler(sent, mask)
 
+        emb = self.pooler(sent, mask)
         # append any specific token representations, e.g. for WiC task
         ctx_embs = []
-        for idx in [i.long() for i in idxs]:
+        for idx in idxs:
```
sleepinyourhat: Add a comment explaining those cases?

src/modules/modules.py
```diff
@@ -444,19 +444,29 @@ def __init__(self, pooler, classifier):
 
     def forward(self, sent, mask, idxs=[]):
         """ Assumes batch_size x seq_len x d_emb """
```
sleepinyourhat: A header could help here. What are idx and sent?

src/tasks/tasks.py
```diff
@@ -2263,23 +2263,36 @@ def load_data(self):
 
         trg_map = {"true": 1, "false": 0, True: 1, False: 0}
 
+        def _process_sent_special(sent, word):
```
sleepinyourhat: More descriptive name? Why 'special'?

src/tasks/tasks.py
```diff
@@ -2263,23 +2263,36 @@ def load_data(self):
...
+                    sent1 = row["sentence1"]
+                    sent2 = row["sentence2"]
+                    sent1, start1, end1 = _process_sent_special(
+                        sent1, sent1[row["start1"] : row["end1"]]
                     )
-                    sent2 = process_sentence(
-                        self._tokenizer_name, row["sentence2"], self.max_seq_len
+                    sent2, start2, end2 = _process_sent_special(
+                        sent2, sent2[row["start2"] : row["end2"]]
                     )
```
sleepinyourhat: Add an assert or test to make sure this does something reasonable (i.e., you're not deleting or doubling the target word).

sleepinyourhat:
Copying over: Add an assert to make sure we don't load the old data file with this code.

src/tasks/tasks.py
W4ngatang: Hmm, any recommendations for how to do this? I'm not sure there's a straightforward way.

src/tasks/tasks.py
W4ngatang: We want to track the indices of the marked word after processing so that we can extract its representation.

src/modules/simple_modules.py
```diff
@@ -103,20 +103,49 @@ def __init__(self, pooler, classifier):
...
+
         emb = self.pooler(sent, mask)
 
         # append any specific token representations, e.g. for WiC task
         ctx_embs = []
-        for idx in [i.long() for i in idxs]:
+        for idx in idxs:
             if len(idx.shape) == 1:
                 idx = idx.unsqueeze(-1)
             if len(idx.shape) == 2:
```
sleepinyourhat: Verify that this case still works, with some demo experiments and some debug printing.

src/modules/simple_modules.py
```diff
@@ -136,11 +165,25 @@ def __init__(self, pooler, classifier, attn=None):
         self.attn = attn
 
     def forward(self, s1, s2, mask1, mask2, idx1=[], idx2=[]):
-        """ s1, s2: sequences of hidden states corresponding to sentence 1,2
-            mask1, mask2: binary mask corresponding to non-pad elements
-            idx{1,2}: indexes of particular tokens to extract in sentence {1, 2}
-                and append to the representation, e.g. for WiC
         """
+        This class applies some type of pooling to get a fixed-size vector,
```
sleepinyourhat: '...for each input sequence'?

src/modules/simple_modules.py
sleepinyourhat: (Not all of these changed cases will be touched by demo.conf, IIRC.)

src/tasks/tasks.py
sleepinyourhat: Do some manual checking/verification for a couple of test tasks, and assert that (end_idx - start_idx) > 0.

src/tasks/tasks.py
```diff
@@ -2265,26 +2265,43 @@ def load_data(self):
...
-                    idxs2.append(int(idx2))
+                    idxs1.append((start1, end1))
+                    idxs2.append((start2, end2))
                     trg = trg_map[row["label"]] if "label" in row else 0
                     trgs.append(trg)
-                return [sents1, sents2, idxs1, idxs2, trgs]
+                    idxs.append(row["idx"])
+                    assert (
+                        row["version"] == 1.1
+                    ), "WiC version is not v1.1; you may be using stale data."
```
sleepinyourhat: stale => something else. Explain why people should update WiC, rather than commenting this out, if they don't actually plan to submit to our server.

sleepinyourhat:
Still can't totally follow, so check very carefully, but merge when ready.

src/modules/modules.py
W4ngatang: They're not really cases; for the `torch.gather` we need the `idx` tensor to be three dimensional and this basically checks and modifies the shape.

src/modules/simple_modules.py
W4ngatang: Hmm I guess I was unclear / lazily programming. These are not cases so much as checks that the `idx` tensor is the shape we need it to be (3-dimensional, matching `sent`).

W4ngatang:
Did some manual checks and added a check: marked word and where our indices say the marked word starts should start the same (doesn't work with ends b/c of added tokenizer cruft).

Issue #709, opened 2019-06-19
Update AllenNLP to >= 0.8.4


sleepinyourhat:
https://github.com/allenai/allennlp/releases

sleepinyourhat:
Necessary for #730, taking this on now.

Issue #730, opened 2019-06-24
Add support for XLNet and the new whole-word-masking variant of BERT.


sleepinyourhat:

sleepinyourhat:
This shouldn't require any code changes, just updating to the next release of pytorch_pretrained_bert once it comes out. We can test it out by pulling pytorch_pretrained_bert at head now.

sleepinyourhat:
This is now out in pytorch_pretrained_bert. Now, does anyone see where/how we're installing pytorch_pretrained_bert?

This doesn't need to go along with 1.0, but as soon as the code is stable, we should add it. Frankly: For the vast majority of new experiments, it doesn't make sense to use plain BERT.

sleepinyourhat:
It looks like this'll break old task pickles, but doesn't require any change to our code.

W4ngatang:
I'm not super familiar with XLNet on a low level; do they use the same BERT modeling tricks (special tokens, concatenating inputs, etc.) ? 

Mostly our BERT-specific code is very BERT-specific, so we'd likely have to rename various variables at a minimum to incorporate XLNet. I'm worried about more drastic changes; I don't think our code is well-abstracted enough to support arbitrary (pretrained LM) model switch. BERT with whole word masking seems like it shouldn't break too much, if anything at all.

sleepinyourhat:
I also misread the HuggingFace readme—XLNet isn't ready yet—so we'll have to wait and see.

I found the requirement, though—we were sneakily installing the HuggingFace repo via Allen. I'll at least try to get whole-word set up.

sleepinyourhat:
The HuggingFace update is out—I'll start looking into adding support...

PR #765, opened 2019-06-26
AllenNLP update/support for whole-word-masking BERT [WIP]

sleepinyourhat:
Addresses half of #730.


environment.yml
```diff
@@ -11,7 +11,7 @@ dependencies:
     - pandas=0.23.0
     - pytz=2017.3
     - pip:
-        - allennlp==0.8.1
+        - allennlp==0.8.4
```
sleepinyourhat: For compatibility with the latest pytorth_pretrained_bert. Doesn't break anything else.

sleepinyourhat:
Also resolves #709.

sleepinyourhat:
Relevant: https://github.com/huggingface/pytorch-pretrained-BERT/issues/659#issuecomment-506210983

environment.yml
sleepinyourhat: False alarm. This does break stuff.

sleepinyourhat:
@W4ngatang, I may need your help on this one. It looks like upgrading AllenNLP 0.8.1 => 0.8.4 breaks our patched version of MultiLabelField, and I don't think I understand the use case for the patch well enough to fix it. 

Mind taking a look? We can wait until after 1.0 if necessary, but it'd be nice to be up to date from launch.

W4ngatang:
Can take a look

environment.yml
pruksmhc: Has something been done to make sure this does not break things?

pruksmhc:
What's the status of this @sleepinyourhat ?

sleepinyourhat:
See last couple of comments—it's in @W4ngatang's hands.

sleepinyourhat:
Sooner is better, but doesn't have to be in 1.0.

W4ngatang:
@sleepinyourhat do you have a MWE for the break with multilabel field? I'm not seeing it running demo.

W4ngatang:
Oh, looks like WSC uses it. NVM.

W4ngatang:
I'm still not getting the error re: `MultiLabelField` using WSC (`winograd-coreference`) to test

sleepinyourhat:
Hrm—it was in a slightly odd setup. If you can't reproduce it and WWM BERT
works, feel free to merge...

W4ngatang:
Caveat: PPB on pip doesn't have the whole word masking models, so users that want to use them will have to install PPB through the git repo for a while.

W4ngatang:
Oh you said that haha, I just can't read.

Issue #773, opened 2019-06-27
WSC is loading is broken again


sleepinyourhat:
```
rm -r $JIANT_PROJECT_PREFIX/jiant-demo;  python main.py --config config/demo.conf --overrides "target_tasks = winograd-coreference"
```
...yields an unhelpful error that we need to fix.
```
Traceback (most recent call last):
  File "main.py", line 566, in <module>
    main(sys.argv[1:])
  File "main.py", line 534, in main
    phase="target_train",
  File "/Users/srbowman/jiant/src/trainer.py", line 594, in train
    output_dict = self._forward(batch, task=task)
  File "/Users/srbowman/jiant/src/trainer.py", line 1035, in _forward
    model_out = self._model.forward(task, tensor_batch)
  File "/Users/srbowman/jiant/src/models.py", line 830, in forward
    out = self._span_forward(batch, task, predict)
  File "/Users/srbowman/jiant/src/models.py", line 917, in _span_forward
    out = module.forward(batch, sent_embs, sent_mask, task, predict)
  File "/Users/srbowman/jiant/src/modules/span_modules.py", line 134, in forward
    out["loss"] = self.compute_loss(logits, batch["labels"].squeeze(dim=1), task)
  File "/Users/srbowman/jiant/src/modules/span_modules.py", line 184, in compute_loss
    raise ValueError("Unsupported loss type '%s' " "for edge probing." % self.loss_type)
ValueError: Unsupported loss type '' for edge probing.
Traceback (most recent call last):
  File "main.py", line 577, in <module>
    raise e  # re-raise exception, in case debugger is attached.
  File "main.py", line 566, in <module>
    main(sys.argv[1:])
  File "main.py", line 534, in main
    phase="target_train",
  File "/Users/srbowman/jiant/src/trainer.py", line 594, in train
    output_dict = self._forward(batch, task=task)
  File "/Users/srbowman/jiant/src/trainer.py", line 1035, in _forward
    model_out = self._model.forward(task, tensor_batch)
  File "/Users/srbowman/jiant/src/models.py", line 830, in forward
    out = self._span_forward(batch, task, predict)
  File "/Users/srbowman/jiant/src/models.py", line 917, in _span_forward
    out = module.forward(batch, sent_embs, sent_mask, task, predict)
  File "/Users/srbowman/jiant/src/modules/span_modules.py", line 134, in forward
    out["loss"] = self.compute_loss(logits, batch["labels"].squeeze(dim=1), task)
[...]

PR #774, opened 2019-06-27
Change classifier_loss_fn name, default

sleepinyourhat:
Fixes a small part of #773

### DIFF
--- a/config/defaults.conf
+++ b/config/defaults.conf
@@ -326,9 +326,9 @@ pair_attn = 1  // If true, use attn in sentence-pair classification/regression t
 d_hid_attn = 512  // Post-attention LSTM state size.
 shared_pair_attn = 0  // If true, share pair_attn parameters across all tasks that use it.
 d_proj = 512  // Size of task-specific linear projection applied before before pooling.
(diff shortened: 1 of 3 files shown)


sleepinyourhat:
@iftenney  -  Will this break any edge probing scripts?

config/defaults.conf
```diff
@@ -326,13 +326,13 @@ pair_attn = 1  // If true, use attn in sentence-pair classification/regression t
 d_hid_attn = 512  // Post-attention LSTM state size.
 shared_pair_attn = 0  // If true, share pair_attn parameters across all tasks that use it.
 d_proj = 512  // Size of task-specific linear projection applied before before pooling.
-classifier_loss_fn = ""  // Classifier loss function. Used only in some tasks (notably
-                         // span-related tasks),  not mlp/fancy_mlp. Currently supports
+classifier_loss_fn = "softmax"  // Classifier loss function. Used only in some tasks (notably
```
iftenney: Maybe for a future PR, but if this is specific to the span classifier, give it a more specific name like `span_classifier_loss` ?

Or just use the same parameter that's used for the rest of tasks, assuming the functionality is more or less the same.

PR #776, opened 2019-06-27
WSC cuda fix

sleepinyourhat:
Meant to fix the second half of #773.

I'm not sure if this is necessary—it may be safe to just delete this entirely and trust that PyTorch will move this over when it moves over the rest of the model. In any case, though, raw cuda() calls break CPU machine support. 

@pruksmhc, you wrote this part—any ideas?


pruksmhc:
This has to be manually moved over to GPUsince batch is in GPU but ones is not.

PR #777, opened 2019-06-27
apex setup note

sleepinyourhat:


sleepinyourhat:
Caveat: I can't actually get apex to install, I'm just passing along a recommendation from HuggingFace.

PR #778, opened 2019-06-27
Fixing labels for WSC

pruksmhc:
In the current version of WSC, the labels were being switched ("False"in the dataset was being mapped to 1, and "True" to 0). This PR fixes it.


sleepinyourhat:
Eek. When did this show up and what results did it impact? Is there a mismatch between the results we got locally and on the server?

pruksmhc:
I was loading labels a bit differently at first (reading from a separate file) and when refactoring the code to be more clean, this subtle label switching was introduced. It basically resulted in majority True class being predicted even though the dataset is majority False.

PR #781, opened 2019-06-27
Delete MT

pruksmhc:
Part 2 of this will involve changing the documentation on jiant.info to include supported/unsupported models. 
https://github.com/nyu-mll/jiant/issues/766


src/tasks/mt.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
         self.files_by_split = {
             split: os.path.join(path, "%s.txt" % split) for split in ["train", "val", "test"]
         }
+        log.warning("Machine translation is not fully supported in Jiant.")
```
sleepinyourhat: => 'jiant'
And maybe add some spacing or exclamation points. This could be easy to miss.

src/tasks/mt.py
```diff
@@ -1,4 +1,8 @@
-"""Task definitions for machine translation tasks."""
+"""
+Task definitions for machine translation tasks.
+NOTE: This is not fully supported currently in jiant.
```
sleepinyourhat: Say more: What parts probably work, and where are people likely to hit issues.

sleepinyourhat:
Looks good to me—ideally, this'll just be a temporary cut to prevent us from releasing anything broken. @Kelina, @anhad13, if you need this in the near future, it's fine to create a new PR to undo this one. We'll just need to do a pass through for serious/obvious bugs before we merge that in.

sleepinyourhat:
Make sure this changes in the docs, too.

Issue #169, opened 2018-07-16
If a run crashes during saving, it won't restore automatically.


sleepinyourhat:
If we write some but not all of the four main training checkpoints for a run and then crash, we won't detect the correct files when restoring. You should be able to get around this by editing filenames.

sleepinyourhat:
Is this still open and claimed?

W4ngatang:
I think this is no longer an issue.

PR #782, opened 2019-06-28
Allowing Restoring Target Train Runs

pruksmhc:
This PR enables restoring target train. 
Changes: 
-Moving checkpoint restoration related functions to utils. 
-Checking for functions to restore
-A trivial change to what is stored in checkpoints during target_tasks to match the API for restore_checkpoints.

To do:
-Tests for check_for_previous_checkpoints() function


main.py
```diff
@@ -503,7 +504,15 @@ def main(cl_arguments):
     if args.do_target_task_training:
         # Train on target tasks
         pre_target_train_path = setup_target_task_training(args, target_tasks, model, strict)
-
+        # Check for previous target train checkpoints
+        task_directory = check_for_previous_checkpoints(
+            args.run_dir, target_tasks, "target_train", args.load_model
+        )
+        if len(task_directory) > 0:
+            # If there is a task to restore form, target train only on target tasks including and following that task.
```
sleepinyourhat: Line len.

main.py
```diff
@@ -503,7 +504,15 @@ def main(cl_arguments):
...
+            name_to_task = collections.OrderedDict([(task.name, task) for task in target_tasks])
+            last_task_index = list(name_to_task.keys()).index(task_directory)
```
sleepinyourhat: Should be able to get rid of collections and use something like target_tasks.index(task_directory)

src/utils/utils.py
```diff
@@ -33,6 +33,51 @@
 _MOSES_DETOKENIZER = MosesDetokenizer()
 
 
+def check_for_previous_checkpoints(serialization_dir, tasks, phase, load_model):
```
sleepinyourhat: The logic here is a bit subtle, so add a comment to spell it out.

src/utils/utils.py
```diff
@@ -33,6 +33,51 @@
...
+    if not load_model:
+        assert_for_log(
+            len(glob.glob(checkpoint_pattern)) == 0,
+            "There are existing checkpoints in %s which will be overwritten. "
+            "If you are restoring from a run, or would like to train from an "
+            "existing checkpoint, Use load_model = 1 to load the checkpoints instead. "
+            "If you don't want them, delete them or change your experiment name."
+            % serialization_dir,
+        )
+    return task_directory
```
sleepinyourhat: More helpful name: most_recent_task_with_ckpt

src/utils/utils.py
```diff
@@ -33,6 +33,51 @@
...
 
 
+def check_for_previous_checkpoints(serialization_dir, tasks, phase, load_model):
+    task_directory = ""
+    for task in tasks[::-1]:
+        last_checkpoint = find_last_checkpoint_suffix(serialization_dir, phase, task.name)
+        # if we haven't found a task directory yet
+        if last_checkpoint is not None and len(task_directory) == 0 and phase == "target_train":
+            task_directory = task.name
+    checkpoint_pattern = os.path.join(serialization_dir, task_directory, "*_{}_*.th".format(phase))
```
sleepinyourhat: To be safer, check that there's a set of four matching checkpoints that are all from the same epoch. That'll help get around #169.

main.py
```diff
@@ -503,7 +504,15 @@ def main(cl_arguments):
     if args.do_target_task_training:
         # Train on target tasks
         pre_target_train_path = setup_target_task_training(args, target_tasks, model, strict)
-
+        # Check for previous target train checkpoints
+        task_directory = check_for_previous_checkpoints(
+            args.run_dir, target_tasks, "target_train", args.load_model
+        )
+        if len(task_directory) > 0:
```
sleepinyourhat: For checkpoints elsewhere, we return None rather than "". Let's use that convention.

src/utils/utils.py
sleepinyourhat: Use something like this here and in main.

src/trainer.py
```diff
@@ -509,25 +511,14 @@ def train(
 
         # define these here b/c they might get overridden on load
         n_pass, should_stop = 0, False
-        if self._serialization_dir is not None and phase == "pretrain":
+        if self._serialization_dir is not None:
             # Resume from serialization path
-            if load_model and any(
```
sleepinyourhat: Where does this logic go?

src/trainer.py
```diff
@@ -1203,15 +1207,15 @@ def _save_target_train_checkpoints(
                 continue
             task_name = self.task_to_metric_mapping[metric_name]
             torch.save(
-                metric_info,
+                metric_states,
```
sleepinyourhat: Why are we saving this, but still iterating over self._metric_infos? This looks broken...

sleepinyourhat: If this code runs, then we should probably be able to just get rid of the loop.

src/utils/utils.py
```diff
@@ -33,6 +33,51 @@
...
+            "If you don't want them, delete them or change your experiment name."
+            % serialization_dir,
+        )
+    return task_directory
+
+
+def find_last_checkpoint_suffix(serialization_dir, search_phase="pretrain", task_name=""):
+
+    """
+    Search for checkpoints to load, looking only for `main` training checkpoints.
```
sleepinyourhat: I think this comment is meant to suggest that this is only for pretraining, which doesn't make sense now that we have a search_phase argument.

src/utils/utils.py
```diff
@@ -33,6 +33,51 @@
...
+            "There are existing checkpoints in %s which will be overwritten. "
+            "If you are restoring from a run, or would like to train from an "
+            "existing checkpoint, Use load_model = 1 to load the checkpoints instead. "
+            "If you don't want them, delete them or change your experiment name."
+            % serialization_dir,
+        )
+    return task_directory
+
+
+def find_last_checkpoint_suffix(serialization_dir, search_phase="pretrain", task_name=""):
```
sleepinyourhat: Alternately, we could put the test to check for a matching set of four checkpoints here, rather than above, and make sure this is called in target_train, too.

src/trainer.py
```diff
@@ -1276,28 +1251,26 @@ def _restore_checkpoint(self, search_phases_in_priority_order=["pretrain"]):
         computation graph, you should use the native Pytorch functions:
         `` model.load_state_dict(torch.load("/path/to/model/weights.th"))``
 
+        We restore based on the phase. If phase=target_train, we look backwards based 
+        on the tasks to find the most recen.
```
sleepinyourhat: typo

src/trainer.py
```diff
@@ -1276,28 +1251,26 @@ def _restore_checkpoint(self, search_phases_in_priority_order=["pretrain"]):
...
         epoch
             The epoch at which to resume training.
         """
-
-        suffix_to_load = self._find_last_checkpoint_suffix(
-            search_phases_in_priority_order=search_phases_in_priority_order
-        )
+        suffix_to_load = None
+        tasks = tasks[::-1]  # reverse the tasks
+        for task in tasks:
```
sleepinyourhat: This feels redundant with check_for_previous_checkpoints. Do we really need both?

sleepinyourhat:
FWIW, I tried this on demo.conf, and it doesn't seem to be working. If I kill a run during pretraining or fine-tuning, and set load_model = 1, I get this error:

```
06/28 02:50:13 PM: Fatal error in main():
Traceback (most recent call last):
  File "main.py", line 576, in <module>
    main(sys.argv[1:])
  File "main.py", line 495, in main
    phase="pretrain",
  File "/nfs/jiant/home/srbowman/jiant/src/trainer.py", line 517, in train
    n_pass, should_stop = self._restore_checkpoint(phase, tasks)
  File "/nfs/jiant/home/srbowman/jiant/src/trainer.py", line 1268, in _restore_checkpoint
    assert suffix_to_load, "No checkpoint found."
AssertionError: No checkpoint found.
Traceback (most recent call last):
  File "main.py", line 587, in <module>
    raise e  # re-raise exception, in case debugger is attached.
  File "main.py", line 576, in <module>
    main(sys.argv[1:])
  File "main.py", line 495, in main
    phase="pretrain",
  File "/nfs/jiant/home/srbowman/jiant/src/trainer.py", line 517, in train
    n_pass, should_stop = self._restore_checkpoint(phase, tasks)
  File "/nfs/jiant/home/srbowman/jiant/src/trainer.py", line 1268, in _restore_checkpoint
    assert suffix_to_load, "No checkpoint found."
AssertionError: No checkpoint found.
```

This shouldn't be a common user-facing error, since the expected behavior with load_model = 1 is just to load a model _if there is one available to load_. It's usually set to 1 for long-running batch jobs, so that the run can start from scratch, but if it gets restarted, it'll restore from checkpoint.

sleepinyourhat:
Thanks for taking this on, by the way!

src/trainer.py
pruksmhc: The logic of searching for model_state checkpoints in the run directory is moved to its own function (check_for_previous_checkpiont in src/utils/utils.py). This part also becomes phase-agnostic as opposed to only for pretrain.

src/trainer.py
pruksmhc: There's enough difference between the two, and check_for_previous_checkpoints callsfind_last_checkpoint_suffix. check_for_previous_checkpoints is used specifically in target_train - otherwise, we would have to put all the code in that function into main.py, which I was trying to avoid.

sleepinyourhat: I guess most of the difference strikes me as unnecessary, unless I'm missing something. There's some logic to prevent us from launching the trainer at all in target_train for finished tasks, but we might as well set that up for both phases if we're going to add it. 

If we check should_stop on each checkpoint to decide whether to launch the trainer, rather than just checking if files exist, we'll get a solution that works for both phases. We'd just need to open the tiny training_state file, not the huge model/optimizer state.

sleepinyourhat: (I know it'd be quicker to just push this as soon as it's working and reasonably well commented, but debugging has gotten so tricky that I'd like to be _really_ careful about increasing code complexity, even if it means a small delay.)

pruksmhc:
Slack seems to be down - but @sleepinyourhat this is ready for a look!

main.py
```diff
@@ -152,10 +152,17 @@ def check_configurations(args, pretrain_tasks, target_tasks):
 
     assert_for_log(
         args.transfer_paradigm in ["finetune", "frozen"],
-        "Transfer paradigm %s not supported!" % args.transfer_paradigm,
+        "Transfer paradigm %s not supported." % args.transfer_paradigm,
```
sleepinyourhat: :D

main.py
```diff
@@ -152,10 +152,17 @@ def check_configurations(args, pretrain_tasks, target_tasks):
...
     )
 
     if args.do_pretrain:
+        task_directory, _, _ = check_for_previous_checkpoints(
+            args.run_dir, target_tasks, "target_train", load_model=1
+        )
+        if task_directory is not None and args.load_model:
+            log.warning(
+                "do_pretrain = 1 but there are currently checkpoints in the target train stage. If you mean to start running from target_train, set do_pretrain=0, do_target_train=1, and load_model=1."
```
sleepinyourhat: This looks like it could backfire. What if someone sets do_pretrain = 1, do_target_train = 1, launches a cluster job, and it crashes during target training? In that case, we should enter into the pretraining code, but then observe that the training job is done and move onto target task training without actually touching any data.

(If we keep this, be consistent about spacing around =.)

main.py
```diff
@@ -152,10 +152,17 @@ def check_configurations(args, pretrain_tasks, target_tasks):
 
     assert_for_log(
         args.transfer_paradigm in ["finetune", "frozen"],
-        "Transfer paradigm %s not supported!" % args.transfer_paradigm,
+        "Transfer paradigm %s not supported." % args.transfer_paradigm,
     )
 
     if args.do_pretrain:
+        task_directory, _, _ = check_for_previous_checkpoints(
```
sleepinyourhat: task_directory still sounds like a path name. Could you use a name that says what this variable is doing here?

main.py
```diff
@@ -503,7 +510,15 @@ def main(cl_arguments):
...
-
+        # Check for previous target train checkpoints
+        task_directory, _, _ = check_for_previous_checkpoints(
+            args.run_dir, target_tasks, "target_train", args.load_model
+        )
+        if task_directory is not None:
+            # If there is a task to restore from, target train only on target tasks
+            # including and following that task.
+            last_task_index = [task.name for task in target_tasks].index(task_directory)
+            target_tasks = target_tasks[last_task_index:]
```
sleepinyourhat: Make sure this filtering operation doesn't impact the eval stage. We still want to eval on everything.

src/trainer.py
```diff
@@ -718,6 +708,7 @@ def clip_function(grad):
                 if should_save:
                     self._save_checkpoint(
                         {"pass": n_pass, "epoch": n_val, "should_stop": should_stop},
+                        tasks=tasks,
                         phase=phase,
                         new_best_macro=new_best_macro,
```
sleepinyourhat: Nice but not crucial: Change this to new_best.

src/trainer.py
```diff
@@ -1118,8 +1071,16 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
         else:
             best_str = ""
 
+        task_directory = ""
+
+        if phase == "target_train":
+            # We only pass in one task at a time during target train phase.
+            task_directory = tasks[0].name
```
sleepinyourhat: assert len(tasks) == 1

sleepinyourhat: (Just for ease of debugging later.)

src/trainer.py
```diff
@@ -1155,149 +1116,71 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
         """
         Restores a model from a serialization_dir to the last saved checkpoint.
         This includes an epoch count and optimizer state, which is serialized separately
-        from  model parameters. This function should only be used to continue training -
+        from model parameters. This function should only be used to continue training -
         if you wish to load a model for inference/load parts of a model into a new
         computation graph, you should use the native Pytorch functions:
         `` model.load_state_dict(torch.load("/path/to/model/weights.th"))``
 
+        We restore based on the phase. If phase=target_train, we start from the lsat
```
sleepinyourhat: lsat => last

src/trainer.py
```diff
@@ -1155,149 +1116,71 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
         -------
-        epoch
-            The epoch at which to resume training.
+        epoch: The epoch at which to resume training.
         """
-
-        suffix_to_load = self._find_last_checkpoint_suffix(
-            search_phases_in_priority_order=search_phases_in_priority_order
+        suffix_to_load = None
+        tasks = tasks[::-1]  # reverse the tasks
```
sleepinyourhat: You reverse tasks inside the next method, too. Intentional?

src/trainer.py
```diff
@@ -1155,149 +1116,71 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
-            self._serialization_dir, "task_state_{}".format(suffix_to_load)
-        )
-        metric_state_path = os.path.join(
-            self._serialization_dir, "metric_state_{}".format(suffix_to_load)
+        assert epoch > -1, "No checkpoint found."
+        log.info("Found checkpoint {}. Loading.".format(suffix))
+        if task_directory is None:
+            task_directory = ""
+        model_path = os.path.join(
+            self._serialization_dir, task_directory, "_".join(["model", suffix])
```
sleepinyourhat: Make sure this behaves properly with task_directory = "" for pretrain.

sleepinyourhat: Might need an if here.

src/utils/utils.py
```diff
@@ -33,6 +33,90 @@
...
+    ---------------------
+    serialization_dir: str, 
+    tasks: List of SamplingMultiTask objects, 
+    phase: str 
+    load_model: bool
+
+    Returns
+    ---------------------
+    task_directory: None or str
+    max_epoch: int, -1 if not found
```
sleepinyourhat: Why not None?

src/utils/utils.py
```diff
@@ -33,6 +33,90 @@
...
+    suffix: None or str, the suffix of the checkpoint
+    """
+    task_directory = None
+    suffix = None
+    ckpt_epoch = -1
+    ckpt_suffix = None
+    if phase == "target_train":
+        for task in tasks[::-1]:
+            max_epoch, suffix = find_last_checkpoint_epoch(serialization_dir, phase, task.name)
+            # if we haven't found a task directory yet
```
sleepinyourhat: Caps

sleepinyourhat: Why > -1? Seems to contradict the comment.

sleepinyourhat: Or I'm missing something.

src/utils/utils.py
```diff
@@ -33,6 +33,90 @@
...
+            )
+    return task_directory, ckpt_epoch, ckpt_suffix
+
+
+def find_last_checkpoint_epoch(serialization_dir, search_phase="pretrain", task_name=""):
+
+    """
+    Search for the last epoch in a directory. 
+    Here, we check that all four checkpoints (model, training_state, task_state, metrics)
+    are loaded and return the most recent epoch with all four checkpoints.
```
sleepinyourhat: 'are loaded' or just exist?

src/utils/utils.py
```diff
@@ -33,6 +33,90 @@
...
+
+    """
+    Search for the last epoch in a directory. 
+    Here, we check that all four checkpoints (model, training_state, task_state, metrics)
+    are loaded and return the most recent epoch with all four checkpoints. 
+
+    """
+    if not serialization_dir:
+        raise ConfigurationError(
+            "serialization_dir not specified - cannot " "restore a model without a directory path."
```
sleepinyourhat: " "

src/utils/utils.py
```diff
@@ -33,6 +33,90 @@
...
+        )
+    max_epoch = -1
```
sleepinyourhat: More consistent to use None here, too.

tests/test_restore_runs.py
```diff
@@ -0,0 +1,128 @@
...
+import unittest
+from unittest import mock
+
+from src import evaluate
+import src.tasks.tasks as tasks
+from src.utils import utils
+from main import evaluate_and_write, get_best_checkpoint_path
+
+
+class TestCheckpointing(unittest.TestCase):
```
sleepinyourhat: Make class name match filename.

tests/test_restore_runs.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        task_directory, max_epoch, suffix = utils.check_for_previous_checkpoints(
+            self.temp_dir, tasks, phase="target_train", load_model=True
+        )
+        assert (
+            task_directory == "mrpc"
+            and max_epoch == 2
+            and suffix == "state_target_train_epoch_2.th"
+        )
+        # Test partial checkpoints.
```
sleepinyourhat: Not sure I follow this part. Add a bit more commentary?

tests/test_restore_runs.py
```diff
@@ -0,0 +1,128 @@
...
+            self.temp_dir, tasks, phase="target_train", load_model=True
+        )
+        assert (
+            task_directory == "sst"
+            and max_epoch == 1
+            and suffix == "state_target_train_epoch_1.best.th"
+        )
+
+    def test_check_for_previous_ckpt_assert(self):
+        # Testing that if args.load_model=1 and there are checkpoints in pretrain or target_train,
```
sleepinyourhat: =0?

main.py
pruksmhc: Right, and we know the training job is done because of the stopped variable that we save in the training state.

pruksmhc: In that case,we can just remove this assert.

src/trainer.py
pruksmhc: Yeah, this works for task_directory = ""

src/utils/utils.py
pruksmhc: Just because usually when it comes to these types of ints, convention is usually -1 if not found.

src/utils/utils.py
pruksmhc: Oh this is a misleading comment - I'll clarify.

src/trainer.py
```diff
@@ -509,26 +511,14 @@ def train(
...
         n_pass, should_stop = 0, False
-        if self._serialization_dir is not None and phase == "pretrain":
+        if self._serialization_dir is not None:
             # Resume from serialization path
-            if load_model and any(
-                ["model_state_" in x for x in os.listdir(self._serialization_dir)]
-            ):
-                n_pass, should_stop = self._restore_checkpoint()
+            if load_model:
+                n_pass, should_stop = self._restore_checkpoint(phase, tasks)
```
sleepinyourhat: We should check that there's a checkpoint before loading: load_model is on by default for good reason, and this crashes if load_model is set but there's no checkpoint. (I had to manually switch this to get a test to run.)

main.py
```diff
@@ -467,8 +467,11 @@ def main(cl_arguments):
         tb_logdir = os.path.join(args.run_dir, "tensorboard")
         _run_background_tensorboard(tb_logdir, cl_args.tensorboard_port)
 
+    if args.do_full_eval:
+        # We evaluate on all target tasks during evaluation stage.
+        eval_tasks = copy.deepcopy(target_tasks)
```
sleepinyourhat: I think it'd be easier to maintain if we did this entirely inside target train. Evaluate on all the target tasks, but create a copy later on with something like `target_tasks_to_train =  copy.deepcopy(target_tasks)`, and only pass those to the trainer.

This is partially just because it's not clear why these two should be different from their names as is.

src/trainer.py
```diff
@@ -1155,149 +1117,70 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
         Returns
         -------
-        epoch
-            The epoch at which to resume training.
+        epoch: The epoch at which to resume training.
         """
-
-        suffix_to_load = self._find_last_checkpoint_suffix(
-            search_phases_in_priority_order=search_phases_in_priority_order
+        suffix_to_load = None
```
sleepinyourhat: This doesn't seem to ever get used...

src/trainer.py
```diff
@@ -1155,149 +1117,70 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
-        metric_state_path = os.path.join(
-            self._serialization_dir, "metric_state_{}".format(suffix_to_load)
+        assert epoch > -1, "No checkpoint found."
+        log.info("Found checkpoint {}. Loading.".format(suffix))
+        if task_directory is None:
+            task_directory = ""
+        model_path = os.path.join(
+            self._serialization_dir, task_directory, "_".join(["model", suffix])
         )
+        training_state_path = model_path.replace("model", "training")
```
sleepinyourhat: Just noticed a major risk here: If someone's run_name or exp_name contains the word 'model' (I could easily imagine a exp name called something like 'sweep_model_size'), that'll get replaced, too. Instead, we should just templatically generate all four names.

main.py
pruksmhc: Sounds good

src/trainer.py
pruksmhc: Alright, I'll just put a log warning here if load_model=1 but there's no checkpoint. 
That problem should only be hit if the user put in load_model=1 though when there is no checkpoint.

sleepinyourhat: There shouldn't be a warning at all. What I'm trying to say is that the most normal/typical/ideal experiment should start with load_model = 1 and no checkpoint. It's only off in defaults.conf since that makes it easier to catch bugs.

src/trainer.py
```diff
@@ -1155,148 +1123,74 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
-                    to_return = x
-            return to_return.split("model_state_")[-1]
-
-    def _restore_checkpoint(self, search_phases_in_priority_order=["pretrain"]):
+    def _restore_checkpoint(self, phase, tasks=None):
         """
         Restores a model from a serialization_dir to the last saved checkpoint.
         This includes an epoch count and optimizer state, which is serialized separately
-        from  model parameters. This function should only be used to continue training -
+        from model parameters. This function should only be used to continue training -
```
sleepinyourhat: It'd help to say why: This will work, but it'll load a bunch of stuff you don't want.

src/trainer.py
```diff
@@ -1155,148 +1123,74 @@ def _save_checkpoint(self, training_state, phase="pretrain", new_best_macro=Fals
...
         if you wish to load a model for inference/load parts of a model into a new
         computation graph, you should use the native Pytorch functions:
         `` model.load_state_dict(torch.load("/path/to/model/weights.th"))``
 
+        We restore based on the phase. If phase=target_train, we start from the last 
+        target task and work backwards, to find the most recent checkpoint in the target 
+        train phase. If phase=pretrain, we check for checkpoints in the main run 
+        directory.
         Returns
```
sleepinyourhat: Whitespace above these blocks?

PR #784, opened 2019-06-28
Deleting old scripts

pruksmhc:
Self-explanatory


sleepinyourhat:
LGTM. @alexwarstadt - We will need to put these scripts up somewhere when the seminar paper fully comes out, but they're probably better off in their own repo, now that jiant is more public-facing. We can just pull them from this commit in the future.

Issue #632, opened 2019-05-01
Test or remove plain transformer support


sleepinyourhat:
From Alex:

> dumb way to support Transformers easily is to allow for resetting the BERT and GPT weights?

PR #786, opened 2019-06-29
Delete transformer

pruksmhc:
https://github.com/nyu-mll/jiant/issues/632


src/trainer.py
```diff
@@ -96,13 +96,11 @@ def build_trainer(
     phase="pretrain",
 ):
     """Build a trainer from params.
-
     Parameters
     ----------
     params: Trainer parameters as built by build_trainer_params.
     model: A module with trainable parameters.
     run_dir: The directory where we save the models.
-
```
sleepinyourhat: Why cut these?

src/trainer.py
```diff
@@ -895,14 +885,12 @@ def _validate(self, epoch, tasks, batch_size, periodic_save=True):
         Validate on all tasks and return the results and whether to save this epoch or not.
         Note/TODO: 'Epoch' here refers to validation passes, not proper epochs over
         any given task's training set.
-
```
pruksmhc: This isn't deleted - not sure why github is showing that it is.

sleepinyourhat: Probably out of sync with master...

src/trainer.py
```diff
@@ -121,29 +121,17 @@ def build_trainer(
         opt_params["warmup"] = 0.1
```
sleepinyourhat: A comment a couple of lines above mentions transformer. Double check that it still applies to BERT or GPT.

pruksmhc: Yep it has to do with BERT.

Issue #767, opened 2019-06-27
Merge bert_model, use_elmo, openai_transformer, and word_embs flags


sleepinyourhat:
This should simplify some documentation quite a bit by making it clearer that we treat all of these as the same glass of objects. The result would probably be a single flag called something like 'input_module' that can take values like 'bert-base-cased', 'gpt', and 'glove'.

pruksmhc:
Probably should be called word_embs since that's what we're treating them as in jiant.

sleepinyourhat:
Could be confusing, too, since that'd probably imply something simple and fixed. I'm not sure the 'contextualized word embedding idea' term on, and even then I'm not sure it applies to BERT.

PR #787, opened 2019-06-29
Merging word_embs, glove, elmo, gpt, bert to input_module

pruksmhc:
Addressing https://github.com/nyu-mll/jiant/issues/767


config/defaults.conf
```diff
@@ -223,11 +223,18 @@ tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task c
                               // Use "" for edge probing tasks to use the original tokenization,
                               // or one of 'OpenAI.BPE', 'MosesTokenizer', 'bert-uncased-base',
                               // etc.
-word_embs = none  // The type of word embedding layer. Set to none when using ELMo, GPT, or BERT.
-                  // Options: none, scratch (i.e., trained from scratch), glove, fastText.
+
+input_module = "" // The word embedding layer, available values are the following:
+                  // gpt, glove, ELMo, scratch (i.e., trained from scratch), fastText. For BERT, 
+                  // specify the version of BERT. Available values for BERT are: bert-base-uncased,
```
sleepinyourhat: Let's keep everything lowercase. It's confusing to have to switch between glove (usually GloVe) and ELMo.

config/defaults.conf
```diff
@@ -223,11 +223,18 @@ tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task c
                               // Use "" for edge probing tasks to use the original tokenization,
                               // or one of 'OpenAI.BPE', 'MosesTokenizer', 'bert-uncased-base',
                               // etc.
-word_embs = none  // The type of word embedding layer. Set to none when using ELMo, GPT, or BERT.
-                  // Options: none, scratch (i.e., trained from scratch), glove, fastText.
+
+input_module = "" // The word embedding layer, available values are the following:
+                  // gpt, glove, ELMo, scratch (i.e., trained from scratch), fastText. For BERT, 
+                  // specify the version of BERT. Available values for BERT are: bert-base-uncased, 
+                  // bert-large-cased. For more, see:
```
sleepinyourhat: You deleted an ellipsis, and now this could read as saying we only support these two BERT versions.

config/defaults.conf
```diff
@@ -223,11 +223,18 @@ tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task c
...
+                  //   https://github.com/huggingface/pytorch-pretrained-BERT
+                  //   #loading-google-ais-pre-trained-weigths-and-pytorch-dump
+                  // For ELMo, BERT, and GPT, there are additional config parameters. 
+
 word_embs_file = ${WORD_EMBS_FILE}  // Path to embeddings file.
 d_word = 300  //  Dimension of word embeddings. Not used by ELMo, GPT, or BERT.
-elmo = 0  // If true, load and use ELMo.
+
 elmo_chars_only = 0  // If true, use *only* the char CNN layer of ELMo. If false but elmo is true,
```
sleepinyourhat: Even better: Make this another input_module option called elmo-chars-only or elmo-char-layer

main.py
```diff
@@ -309,7 +309,8 @@ def initial_setup(args, cl_args):
     pretrain_tasks: list of pretraining tasks
     target_tasks: list of target tasks
     vocab: list of vocab
-    word_embs: loaded word embeddings, may be None if args.word_embs = none
+    word_embs: loaded word embeddings, may be None if 
+    args.input_module = gpt, ELMo, BERT
```
sleepinyourhat: Spacing

src/models.py
```diff
@@ -311,13 +311,13 @@ def build_embeddings(args, vocab, tasks, pretrained_embs=None):
     token_embedders = {}
     # Word embeddings
     n_token_vocab = vocab.get_vocab_size("tokens")
-    if args.word_embs != "none":
-        if args.word_embs in ["glove", "fastText"] and pretrained_embs is not None:
+    if not args.input_module.startswith("bert") and args.input_module not in ["gpt", "ELMo"]:
```
sleepinyourhat: This doesn't need to be a conditional here. We can just replace the else case at the end with an assert that checks that the argument is valid.

src/preprocess.py
```diff
@@ -241,24 +241,22 @@ def build_indexers(args):
...
             " you are using args.tokenizer = {args.tokenizer}"
         )
-    if args.openai_transformer:
-        assert not indexers, (
-            "OpenAI transformer is not supported alongside" " other indexers due to tokenization!"
-        )
-        assert args.tokenizer == "OpenAI.BPE", (
-            "OpenAI transformer is not supported alongside" " other indexers due to tokenization!"
-        )
+    if args.input_module == "gpt":
```
sleepinyourhat: I know this isn't new, but maybe drop the exclamation points in this section.

config/defaults.conf
```diff
@@ -223,11 +223,18 @@ tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task c
...
-                  // Options: none, scratch (i.e., trained from scratch), glove, fastText.
+
+input_module = "" // The word embedding layer, available values are the following:
+                  // gpt, glove, ELMo, scratch (i.e., trained from scratch), fastText. For BERT, 
+                  // specify the version of BERT. Available values for BERT are: bert-base-uncased, 
+                  // bert-large-cased. For more, see:
+                  //   https://github.com/huggingface/pytorch-pretrained-BERT
+                  //   #loading-google-ais-pre-trained-weigths-and-pytorch-dump
+                  // For ELMo, BERT, and GPT, there are additional config parameters. 
+
```
sleepinyourhat: Mind adding some instructions here about tokenization support? (i.e., which of these require specific tokenization, and how that tokenization should be specified.)

Even better would be to have the option of leaving tokenizer blank, and having it be set to some sane default for each option.

src/trainer.py
```diff
@@ -96,13 +96,11 @@ def build_trainer(
     phase="pretrain",
 ):
     """Build a trainer from params.
-
     Parameters
     ----------
     params: Trainer parameters as built by build_trainer_params.
     model: A module with trainable parameters.
     run_dir: The directory where we save the models.
-
```
sleepinyourhat: Why cut these?

sleepinyourhat:
One more thing that'll have to happen here: Update all the checked-in config files, and add an assert to detect/block the use of the old flags.

src/models.py
```diff
@@ -311,19 +310,26 @@ def build_embeddings(args, vocab, tasks, pretrained_embs=None):
...
+        d_word = args.d_word
+        word_embs = nn.Embedding(n_token_vocab, d_word).weight
+    else:
+        assert args.input_module.startswith("bert") or args.input_module in [
+            "gpt",
+            "elmo",
+            "elmo-chars-only",
+        ], "You do not have a valid value for input_module."
+        embeddings = None
+        log.info("\tNot using word embeddings!")
```
sleepinyourhat: Probably safe to cut.

src/models.py
```diff
@@ -333,15 +339,12 @@ def build_embeddings(args, vocab, tasks, pretrained_embs=None):
...
-    else:
-        embeddings = None
-        log.info("\tNot using word embeddings!")
 
     # Handle cove
     cove_layer = None
     if args.cove:
         assert embeddings is not None
-        assert args.word_embs == "glove", "CoVe requires GloVe embeddings."
+        assert args.input_module == "glove", "CoVe requires GloVe embeddings."
```
sleepinyourhat: Does it make sense to fold cove into this logic as well? Not sure I know that part of the code well enough.

config/defaults.conf
```diff
@@ -223,13 +223,24 @@ tokenizer = "MosesTokenizer"  // The name of the tokenizer, passed to the Task c
...
-word_embs = none  // The type of word embedding layer. Set to none when using ELMo, GPT, or BERT.
-                  // Options: none, scratch (i.e., trained from scratch), glove, fastText.
+
+input_module = "" // The word embedding layer, available values are the following:
+                  // gpt, glove, elmo, elmo-chars-only, scratch (i.e., trained from scratch), fastText. For BERT, specify the version of BERT. Available values for BERT are: 
+                  // bert-base-uncased, bert-large-cased, ... 
+                  // For more, see:
+                  //   https://github.com/huggingface/pytorch-pretrained-BERT
+                  //   #loading-google-ais-pre-trained-weigths-and-pytorch-dump
+                  // Set input_module = elmo-chars-only to use *only* the char CNN
```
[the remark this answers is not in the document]
pruksmhc: What do you mean with lists?

src/models.py
pruksmhc: I don't think so since cove is the only one that relies on another word embedding (glove). Plus, i'm also not the most familiar with it and am not as comfortable putting it in input_modules alongside the others atm.

config/defaults.conf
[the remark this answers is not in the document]
sleepinyourhat: A bulleted list of modules, with any special instructions after each one.

main.py
```diff
@@ -309,7 +309,7 @@ def initial_setup(args, cl_args):
     pretrain_tasks: list of pretraining tasks
     target_tasks: list of target tasks
     vocab: list of vocab
-    word_embs: loaded word embeddings, may be None if args.word_embs = none
+    word_embs: loaded word embeddings, may be None if args.input_module = gpt, elmo, BERT
```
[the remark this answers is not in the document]
sleepinyourhat: This is still wrong as is.

src/models.py
```diff
@@ -158,8 +158,7 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
         log.info("Using PRPN sentence encoder!")
     elif any(isinstance(task, LanguageModelingTask) for task in tasks) or args.sent_enc == "bilm":
         assert_for_log(args.sent_enc in ["rnn", "bilm"], "Only RNNLM supported!")
-        if args.elmo:
-            assert_for_log(args.elmo_chars_only, "LM with full ELMo not supported")
+        assert_for_log(args.input_module == "elmo-chars-only", "LM with full ELMo not supported")
```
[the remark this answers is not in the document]
sleepinyourhat: This needs to apply to BERT, too. See comment.

src/models.py
```diff
@@ -242,12 +241,12 @@ def build_model(args, vocab, pretrained_embs, tasks):
...
         embedder = OpenAIEmbedderModule(args)
         d_emb = embedder.get_output_dim()
-    elif args.bert_model_name:
+    elif args.input_module.startswith("bert"):
         # Note: incompatible with other embedders, but logic in preprocess.py
         # should prevent these from being enabled anyway.
         from .bert.utils import BertEmbedderModule
 
-        log.info(f"Using BERT model ({args.bert_model_name}); skipping other embedders.")
+        log.info(f"Using BERT model ({args.input_module}); skipping other embedders.")
```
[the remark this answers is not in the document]
sleepinyourhat: Bump. If our logs are filled with meaningless comments, no one will read them to notice the actual warnings.

src/models.py
sleepinyourhat: Mind opening an issue? Seems okay to leave for now, but it will make the documentation/code harder to follow. Ultimately, we should fold the glove loading logic into cove, and treat cove as the input module.

sleepinyourhat:
Found another open issue: The old option names still exist in most of the checked-in config files, so most experiments are broken. Reverting.

sleepinyourhat:
Can't auto-revert. Please fix ASAP.

PR #788, opened 2019-06-29
WNLI => CB in demo.conf

sleepinyourhat:
WNLI is a somewhat problematic choice for a demo, since our tiny model doesn't do better than chance, so it's hard to see if optimization is working. CB is similarly tiny, but reliably yields better-than-chance results.


W4ngatang:
Have you verified this is actually above chance? I think the label distribution is skewed so it seems worth checking...

sleepinyourhat:
Yep. Val numbers go up over training, and I manually checked the MFC number in the dev set.

W4ngatang:
Really not sure why the test is failing :/

PR #798, opened 2019-07-01
Fix epoch bug

W4ngatang:
Fixes a bug in loading model checkpoints.

Is it worth going through and changing all mentions of `epoch` in variable names and comments to `val` or something similar? @pruksmhc @sleepinyourhat


sleepinyourhat:
Is there a reason this closed? Looks good to me.

W4ngatang:
Accidentally addressed when I merged in BoolQ

PR #799, opened 2019-07-02
Fixing diagnostics run 

pruksmhc:
Right now, if we pretrain on a task and then evaluate on another (where that evaluation task is in target_tasks and args.do_target_train = 0), then it would not evaluate on that evaluation task and only evaluate on the pretraining task. 

Tested on RTE pretraining, superglue-diagnostic evaluation.


main.py
```diff
@@ -266,6 +266,10 @@ def get_best_checkpoint_path(args, phase, task_name=None):
             checkpoint = glob.glob(
                 os.path.join(args.run_dir, task_name, "model_state_target_train_val_*.best.th")
             )
+            if len(checkpoint) == 0:
```
sleepinyourhat: Update the header comment to match this.

src/tasks/__init__.py
```diff
@@ -35,6 +35,7 @@
     "wic",
     "superglue-diagnostic",
 ]
+ALL_DIAGNOSTICS = ["superglue-diagnostic", "winogender-diagnostic", "glue-diagnostic"]
```
sleepinyourhat: This'll be easier to maintain if we make it an attribute of the diagnostic tasks themselves, rather than a list.

main.py
```diff
@@ -558,6 +562,21 @@ def main(cl_arguments):
...
         log.info("Evaluating...")
         splits_to_write = evaluate.parse_write_preds_arg(args.write_preds)
+        diagnostic_tasks = []
+        target_tasks_to_remove = []
+        for index, task in enumerate(target_tasks):
+            if task.name in task_modules.ALL_DIAGNOSTICS:
+                diagnostic_tasks.append(task)
+                target_tasks_to_remove.append(index)
+        # To avoid evaluating twice on a diagnostic task.
+        for index in sorted(target_tasks_to_remove, reverse=True):
```
sleepinyourhat: Why sort and reverse? Why not just del target_tasks(index) in the previous loop?

main.py
```diff
@@ -558,6 +562,21 @@ def main(cl_arguments):
...
+            del target_tasks[index]
+        for task in diagnostic_tasks:
+            ckpt_path = get_best_checkpoint_path(args, "eval", task.name)
```
sleepinyourhat: Where do we substitute task.name for the appropriate trained model? For SuperGLUE diagnostics, we want task.name to be rte-superglue, not the name of the diagnostic.

main.py
pruksmhc: Because if you delete from first to last, then the indices will get messed up (remove index 0, then all the proceeding indices need to be shuffled down by 1).

main.py
pruksmhc: Right now, the only way to run diagnostics is to have a model in pretraining stage, and then superglue-diagnostics in evaluation. We don't need to switch the task.name to be rte-superglue since it's in pretraining stage and the checkpoints for pretraining stage do not depend on task name. 
Ideally, what we would want is to separate target_tasks from eval_tasks entirely so that it would be possible to pretrain, then target train, then use that target train to evaluate on diagnostics, but that will entail a bigger rewrite.

main.py
sleepinyourhat: Ah, fair point. Another easy alternative: target_tasks.remove(task) in the initial loop, with no indexing at all.

main.py
```diff
@@ -558,6 +563,17 @@ def main(cl_arguments):
     if args.do_full_eval:
         log.info("Evaluating...")
         splits_to_write = evaluate.parse_write_preds_arg(args.write_preds)
+
+        # Even if args.do_target_training = 0, evaluate on diagnostic tasks.
```
sleepinyourhat: This comment seems misleading. We should evaluate on all the target_tasks when do_full_eval is set, whether or not they're diagnostics.

src/tasks/tasks.py
```diff
@@ -196,7 +196,7 @@ def __init__(self, name, tokenizer_name):
         self.name = name
         self._tokenizer_name = tokenizer_name
         self.scorers = []
-
+        self.is_diagnostic = False
```
W4ngatang: I'd prefer to rename this to what the property that we actually care about is. "Diagnostic" is a pretty loose term; does this specifically refer to tasks for which we don't train on ? Tasks that use another task's classifier?

main.py
W4ngatang: I think at some point, you'll have to maintain a list of pretraining tasks, target tasks that you train on, target tasks that you don't train on. We probably want to maintain only giving users the choice between pretraining and target tasks (merging the second two categories into one) and handling the task management ourselves, like we're doing here.

src/tasks/tasks.py
W4ngatang: IDK how to leave the comment on the code diff, but shouldn't `main.py:533` use this flag instead of checking on the task class?

main.py
pruksmhc: In that case we can actually cut a lot of the logic here if we're evaluating on all target_tasks.

sleepinyourhat: Yep!

src/tasks/tasks.py
pruksmhc: Good point- changing in the code

main.py
```diff
@@ -558,24 +562,14 @@ def main(cl_arguments):
...
-            # without pretraining or target task training
-            # then we evaluate on the target tasks.
-            for task in target_tasks:
-                # Find the task-specific best checkpoint to evaluate on.
-                ckpt_path = get_best_checkpoint_path(args, "eval", task.name)
-                assert ckpt_path is not None
-                load_model_state(model, ckpt_path, args.cuda, skip_task_models=[], strict=strict)
-
-                evaluate_and_write(args, model, [task], splits_to_write)
-        elif args.do_pretrain:
```
pruksmhc: Note that now, if we pretrain on a task and then evaluate on that task, we have to put that task in both pretrain_tasks and target_tasks.

sleepinyourhat: Sounds right!

config/defaults.conf
```diff
@@ -71,8 +71,12 @@ pretrain_tasks = sst  // Comma-separated list of pretraining tasks or 'glue' or
                       // do_pretrain to 1, and do_target_task_training to 0.
 target_tasks = glue  // Target tasks, for use in both target_task_training and the final
                      // evaluation, in the same list format as pretrain_tasks.
+                     // If args.do_full_eval = 1, then the target_tasks will be used for
```
pruksmhc: args.do_full_eval  -> do_full_eval, also remove "the"

pruksmhc:
@W4ngatang ready for another review

config/defaults.conf
```diff
@@ -71,7 +71,8 @@ pretrain_tasks = sst  // Comma-separated list of pretraining tasks or 'glue' or
                       // do_pretrain to 1, and do_target_task_training to 0.
 target_tasks = glue  // Target tasks, for use in both target_task_training and the final
                      // evaluation, in the same list format as pretrain_tasks.
-
+                     // If do_full_eval = 1, then target_tasks will be used for 
+                     // model evaluation.
```
sleepinyourhat: The previous sentence says the same thing ('for use in the final evaluation'). Merge the two sentences or, preferably, just revert the change. We're going back to the behavior from before last month, so the documentation should mostly be correct.

src/tasks/tasks.py
sleepinyourhat: Agree with Alex's initial suggestion. Maybe use something like 'eval_only_task = 1' or 'can_be_trained = 0'.

Also, this'll likely be temporary—once we finish cleaning up the logic for diagnostic tasks that use classifiers from other tasks, we can simply test whether this task uses its own classifier, and skip training if not.

src/evaluate.py
```diff
@@ -203,7 +202,7 @@ def write_preds(
             _write_winograd_preds(
                 task, preds_df, pred_dir, split_name, strict_glue_format=strict_glue_format
             )
-        elif isinstance(task, GLUEDiagnosticTask):
+        elif task.is_diagnostic:
```
sleepinyourhat: I'd vote to revert this change. For deciding whether to train, it makes sense to check a simple Boolean that we can easily reuse for new tasks. But for deciding how to write preds, we actually need to know something about tho data structures used for the task, so I think it's right to check isinstance or something similarly specific.

src/models.py
```diff
@@ -744,7 +743,7 @@ def forward(self, task, batch, predict=False):
                 self.utilization(get_batch_utilization(batch["input"]))
         if isinstance(task, SingleClassificationTask):
             out = self._single_sentence_forward(batch, task, predict)
-        elif isinstance(task, GLUEDiagnosticTask):
+        elif task.is_diagnostic:
```
sleepinyourhat: As above, I'd vote to revert this change. For deciding whether to train, it makes sense to check a simple Boolean that we can easily reuse for new tasks. But for deciding which model code to run, we actually need to know something about tho data structures used for the task, so I think it's right to check isinstance or something similarly specific.

config/defaults.conf
pruksmhc: I added the specifics about do_full_eval to make it extra clear that even if do_target_train = 0, target_tasks will be evaluated on.

pruksmhc:
Reverted the changes.

Issue #631, opened 2019-05-01
Clean up pool_type logic


sleepinyourhat:
Alex expressed concern that this option may not work.

```
pool_type = max // The type of pooling to aggregate sequences of vectors into a single vector.
                // Options: first, last, max, mean.
```

The easy/temporary fix would be to simply delete the option.

sleepinyourhat:
Not resolved yet, but bumping to needed-for-1.0, since it's no longer a major issue.

sleepinyourhat:
Still on it? This is the last day I can do much reviewing before Mon.

W4ngatang:
Resolved with #802

Issue #532, opened 2019-03-25
Making hidden state pooling a user-controllable option


W4ngatang:
Currently we force people to aggregate sequence encoder hidden states using max pooling (and first-pooling for BERT). It'd be nice to expose this as an option.

sleepinyourhat:
Still on it? This is the last day I can do much reviewing before Mon.

W4ngatang:
resolved with #802

PR #802, opened 2019-07-04
Pool type

W4ngatang:
Addresses #631 and #532 .

Exposes `pool_type` as a cmdline arg and uses it in model construction. Note that for BERT runs, we now need to manually specify `first` pooling. This commit also adds it to BERT configs.

### DIFF
--- a/config/defaults.conf
+++ b/config/defaults.conf
@@ -342,6 +342,8 @@ pair_attn = 1  // If true, use attn in sentence-pair classification/regression t
 d_hid_attn = 512  // Post-attention LSTM state size.
 shared_pair_attn = 0  // If true, share pair_attn parameters across all tasks that use it.
 d_proj = 512  // Size of task-specific linear projection applied before before pooling.
+pool_type = "max" // Type of pooling to reduce sequences of vectors into a single vector.
+                  // Options: "max", "mean", "first", "final"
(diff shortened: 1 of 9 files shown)


src/models.py
```diff
@@ -583,7 +583,9 @@ def build_single_sentence_module(task, d_inp: int, use_bert: bool, params: Param
         - SingleClassifier (nn.Module): single-sentence classifier consisting of
```
sleepinyourhat: Header for this fn. says that pool_type depends on use_bert. Is that still right?

config/defaults.conf
```diff
@@ -342,6 +342,8 @@ pair_attn = 1  // If true, use attn in sentence-pair classification/regression t
 d_hid_attn = 512  // Post-attention LSTM state size.
 shared_pair_attn = 0  // If true, share pair_attn parameters across all tasks that use it.
 d_proj = 512  // Size of task-specific linear projection applied before before pooling.
+pool_type = "max" // Type of pooling to reduce sequences of vectors into a single vector.
+                  // Options: "max", "mean", "first", "final"
```
sleepinyourhat: Idea, though not really a suggestion: Have an 'auto' option, maybe yielded by a blank string here, that uses max for most encoders, first for BERT, and final for GPT.

W4ngatang: That's nice to have but in the interest of getting this through, I'm going to skip for now.

PR #806, opened 2019-07-04
Cleaning up Tensorboard

pruksmhc:
Right now, tensorboard doesn't work for micro_avg and macro_avg (see picture attached), so removing them for now. It correctly displays the task-specific metrics though.

Screenshot 2019-07-04 15 06 45


src/trainer.py
```diff
@@ -1247,8 +1249,10 @@ def _metrics_to_tensorboard_val(self, val_pass, val_metrics):
         metric_names = val_metrics.keys()
 
         for name in metric_names:
+            if name == "micro_avg" or name == "macro_avg":
+                continue
             val_metric = val_metrics.get(name)
-            name = name.split("_")[0] + "/" + name
+            name = os.path.join(name.split("_")[0], name)
```
sleepinyourhat: See if you can rewrite this and 1242 to avoid this delicate string manipulation logic. It seems plausible that that's what's breaking micro/macro...

PR #808, opened 2019-07-05
Update MultiRC to new data format

W4ngatang:
Some changes in MultiRC data format necessitates code changes:

- "isAnswer" -> "label"
- "paragraph" -> "passage"
- removed HTML from passage text
- check versioning

### DIFF
--- a/src/tasks/qa.py
+++ b/src/tasks/qa.py
@@ -109,7 +109,7 @@ class MultiRCTask(Task):
             for example in data_fh:
                 ex = json.loads(example)
 
-                assert "version" in ex and ex["version"] == 1.1, "MultiRC version is invalid!"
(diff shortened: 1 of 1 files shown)


src/tasks/qa.py
```diff
@@ -108,16 +108,16 @@ def load_data_for_path(self, path):
             examples = []
             for example in data_fh:
                 ex = json.loads(example)
-                # each example has a paragraph field -> (text, questions)
-                # text is the paragraph, which requires some preprocessing
+
+                assert "version" in ex and ex["version"] == 1.1, "MultiRC version is invalid!"
```
sleepinyourhat: Say something more helpful: Why shouldn't people use the old version?

sleepinyourhat: Also, maybe just assert this for the 0th example?

PR #811, opened 2019-07-07
Add baselines

W4ngatang:
Adds commands for the STILTS-style SuperGLUE baselines (BERT++ in the tables) and CBOW baselines. I could see an argument for separating the CBOW commands into a different script.


src/tasks/tasks.py
```diff
@@ -2277,7 +2277,7 @@ def _process_preserving_word(sent, word):
             start_idx = len(sent_tok1[:-1])
             end_idx = start_idx + len(sent_mid[1:-1])
             assert end_idx > start_idx, "Invalid marked word indices. Something is wrong."
-            assert word.startswith(sent_tok[start_idx]), "Detected corrupted marked word."
+            # assert word.startswith(sent_tok[start_idx]), "Detected corrupted marked word."
```
sleepinyourhat: Cut from PR

PR #818, opened 2019-07-08
Adding winogender to downloading script

pruksmhc:
Self-explanatory


scripts/download_superglue_data.py
```diff
@@ -47,6 +48,13 @@ def download_diagnostic(data_dir):
         os.mkdir(os.path.join(data_dir, "RTE"))
```
sleepinyourhat: The print above will print twice now with the same message, even if you make each fn call only download one dataset.

scripts/download_superglue_data.py
```diff
@@ -84,7 +92,7 @@ def main(arguments):
     tasks = get_tasks(args.tasks)
 
     for task in tasks:
-        if task == "diagnostic":
+        if task in ["diagnostic", "winogender-diagnostic"]:
             download_diagnostic(args.data_dir)
```
sleepinyourhat: This will call the same method twice if you request all tasks.

scripts/download_superglue_data.py
```diff
@@ -47,6 +48,13 @@ def download_diagnostic(data_dir):
         os.mkdir(os.path.join(data_dir, "RTE"))
     data_file = os.path.join(data_dir, "RTE", "diagnostic-full.tsv")
     urllib.request.urlretrieve(TASK2PATH["diagnostic"], data_file)
+    winogender_dir = os.path.join(data_dir, "RTE", "diagnostics")
```
sleepinyourhat: Why is winogender in a subdir, but diagnostic-full not? We should be extremely consistent between these two.

pruksmhc: I'll change it so they're both in the diagnostics subdir

Issue #822, opened 2019-07-08
RTE BOW SuperGLUE baseline broken


sleepinyourhat:
Running the standard SuperGLUE RTE BOW baseline command from head, in a fresh directory with a fresh download, yields the following:

```
07/08 05:05:38 PM: Saved config to /Users/Bowman/Drive/JSALT/bow-rte-v2/rte/params.conf
07/08 05:05:38 PM: Using random seed 1234
07/08 05:05:38 PM: GPU access failed. You might be using a CPU-only installation of PyTorch. Falling back to CPU.
07/08 05:05:38 PM: Loading tasks...
07/08 05:05:38 PM: Writing pre-preprocessed tasks to /Users/Bowman/Drive/JSALT/bow-rte-v2/
07/08 05:05:38 PM: 	Creating task rte-superglue from scratch.
07/08 05:05:38 PM: 	Loading Tokenizer MosesTokenizer
07/08 05:05:41 PM: 	Finished loading RTE (from SuperGLUE formatted data).
07/08 05:05:41 PM: 	Task 'rte-superglue': |train|=2490 |val|=277 |test|=3000
07/08 05:05:41 PM: 	Finished loading tasks: rte-superglue.
07/08 05:05:41 PM: 	Building vocab from scratch.
07/08 05:05:41 PM: 	Counting words for task rte-superglue.
07/08 05:05:42 PM: 	Saved vocab to /Users/Bowman/Drive/JSALT/bow-rte-v2/vocab
07/08 05:05:42 PM: Loading token dictionary from /Users/Bowman/Drive/JSALT/bow-rte-v2/vocab.
07/08 05:05:42 PM: 	Loaded vocab from /Users/Bowman/Drive/JSALT/bow-rte-v2/vocab
07/08 05:05:42 PM: 	Vocab namespace chars: size 108
07/08 05:05:42 PM: 	Vocab namespace tokens: size 18927
07/08 05:05:42 PM: 	Finished building vocab.
07/08 05:05:42 PM: 	Building embeddings from scratch.
07/08 05:05:58 PM: 	Finished loading embeddings
07/08 05:05:58 PM: 	Saved embeddings to /Users/Bowman/Drive/JSALT/bow-rte-v2/embs.pkl
07/08 05:05:58 PM: Trimmed word embeddings: torch.Size([18927, 300])
07/08 05:05:58 PM: 	Task rte-superglue (train): Indexing from scratch.
[...]

PR #823, opened 2019-07-08
Some fixes

W4ngatang:
- In `rte-bow` move Winogender to a target task
- Revert breaking `rte-superglue` data loading change
- Return something in Winogender `process_split`
- Change data file from `winogender_filtered.jsonl` -> `Winogender.jsonl` to match server download

### DIFF
--- a/jiant/tasks/tasks.py
+++ b/jiant/tasks/tasks.py
@@ -1548,6 +1548,7 @@ class RTESuperGLUETask(RTETask):
     def load_data(self):
         """ Process the datasets located at path. """
         targ_map = {"not_entailment": 0, True: 1, False: 0, "entailment": 1}
+
         def _load_jsonl(data_file):
             data = [json.loads(d) for d in open(data_file, encoding="utf-8")]
             sent1s, sent2s, trgs, idxs = [], [], [], []
--- a/jiant/tasks/tasks.py
+++ b/jiant/tasks/tasks.py
@@ -1548,15 +1548,25 @@ class RTESuperGLUETask(RTETask):
     def load_data(self):
         """ Process the datasets located at path. """
         targ_map = {"not_entailment": 0, True: 1, False: 0, "entailment": 1}
-        self.train_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "train.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
-        self.val_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "val.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
-        self.test_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "test.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
+        def _load_jsonl(data_file):
+            data = [json.loads(d) for d in open(data_file, encoding="utf-8")]
+            sent1s, sent2s, trgs, idxs = [], [], [], []
+            for example in data:
+                sent1s.append(
+                    process_sentence(self._tokenizer_name, example["premise"], self.max_seq_len)
+                )
+                sent2s.append(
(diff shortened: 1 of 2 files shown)


jiant/tasks/tasks.py
```diff
@@ -1416,19 +1416,19 @@ def load_data(self):
         targ_map = {"not_entailment": 0, "entailment": 1}
 
         self.train_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "winogender_filtered.jsonl"),
+            os.path.join(self.path, "Winogender.jsonl"),
```
pruksmhc: I think it should be WInoGender no? (at least how Amanpreet had named the file).

jiant/tasks/tasks.py
```diff
@@ -1545,15 +1548,25 @@ class RTESuperGLUETask(RTETask):
...
-        self.train_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "train.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
-        self.val_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "val.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
-        self.test_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "test.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
+        def _load_jsonl(data_file):
```
pruksmhc: can you revert this change? This will fail in that winogender_metrics requires pair_id field.

jiant/tasks/tasks.py
W4ngatang: I told him to use `Winogender` (that's how the paper refers to itself)

jiant/tasks/tasks.py
W4ngatang: Why should winogender metrics fail on this RTE example?

jiant/tasks/tasks.py
pruksmhc: Alright - currently it isn't named that way

jiant/tasks/tasks.py
```diff
@@ -1545,15 +1548,25 @@ class RTESuperGLUETask(RTETask):
     def load_data(self):
         """ Process the datasets located at path. """
         targ_map = {"not_entailment": 0, True: 1, False: 0, "entailment": 1}
-        self.train_data_text = load_pair_nli_jsonl(
-            os.path.join(self.path, "train.jsonl"), self._tokenizer_name, self.max_seq_len, targ_map
-        )
```
pruksmhc: could you change load_pair_nli_json instead of re-inserting it back here? For example, instead of return [sent1s, sent2s, trgs, idxs, pair_ids], perhaps have another condition that only returns the first four items for RTE.

jiant/tasks/tasks.py
```diff
@@ -1545,15 +1548,25 @@ class RTESuperGLUETask(RTETask):
     def load_data(self):
         """ Process the datasets located at path. """
         targ_map = {"not_entailment": 0, True: 1, False: 0, "entailment": 1}
-        self.train_data_text = load_pair_nli_jsonl(
```
pruksmhc: Also, while you're at it, can you change the superglue-diagnosticheader so that the relative path is RTE/diagnostics?

W4ngatang: I'll address in the other diagnostics PR

jiant/tasks/tasks.py
W4ngatang: I'd rather not rely on a helper function in another file here. It's pretty straightforward to load a JSONL, and it only gets more complicated with all these different task-specific quirks in a single function.

W4ngatang:
Address #822

jiant/tasks/tasks.py
pruksmhc: Alright - the thinking was basically to reduce the amount of repeated code in the codebase, but up to you!

PR #826, opened 2019-07-09
Fix bugs

W4ngatang:
Fixes some bugs:

- In some task, accidentally wrote labels instead of predictions
- fix bug in write_preds test where we didn't have predictions...
- fix preprocessing bug in broad coverage diagnostics
- use `rte-superglue` classifier for Winogender
- fix bug in Winogender metrics that used logits as predictions
- fix bug in Winogender preprocessing where there were no labels when using BERT
- fix bug in download script where we weren't unzipping downloads
- add v2.0 to all download URLs and update broad coverage diagnostic

Some cleanup:

- remove mention of wmt in evaluation
- update baseline script comment

### DIFF
--- a/scripts/download_superglue_data.py
+++ b/scripts/download_superglue_data.py
@@ -33,15 +33,14 @@ TASK2PATH = {
 
 def download_and_extract(task, data_dir):
     print("Downloading and extracting %s..." % task)
-    if not os.path.isdir(os.path.join(data_dir, task)):
-        os.mkdir(os.path.join(data_dir, task))
-    data_file = os.path.join(data_dir, task, "%s.zip" % task)
-
+    if not os.path.isdir(data_dir):
+        os.mkdir(data_dir)
+    data_file = os.path.join(data_dir, "%s.zip" % task)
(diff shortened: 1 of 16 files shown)


W4ngatang:
Failing tests until Amanpreet moves the data to the v2.0 URLs.

W4ngatang:
It'd be really nice is someone could sanity check this and run the diagnostics from this branch.

scripts/download_superglue_data.py
```diff
@@ -51,12 +51,17 @@ def download_diagnostic(data_dir):
     diagnostic_dir = os.path.join(data_dir, "RTE", "diagnostics")
```
sleepinyourhat: Seems cleaner to make these two different tasks/methods, but not a big deal.

pruksmhc:
There's some path-related issues with the current download_scripts

pruksmhc:
Basically, the output of python download_superglue_scripts.py for RTE and diagnostics is AX-B and AX-g zip files in the same level as RTE.

PR #831, opened 2019-07-10
1.0 to readme.

sleepinyourhat:
Still need to make the 1.0 package and make a corresponding update to the site homepage (which probably also says WIP) and to the tutorial.


sleepinyourhat:
@W4ngatang: Bonus - cite the citation in the paper?

PR #832, opened 2019-07-10
Fix wsc labels

pruksmhc:
Currently, WSC labels are being loaded via the get_all_labels() function, which is mapping the labels to integers. However, this is causing complications when running a non-WSC baseline, followed by the WSC baseline, due to some naming errors. 
Here, we implement a more standard flow for preprocesisng abels.

### DIFF
--- a/jiant/tasks/tasks.py
+++ b/jiant/tasks/tasks.py
@@ -2337,15 +2337,8 @@ class SpanClassificationTask(Task):
             example["span" + str(i + 1) + "s"] = ListField(
                 [self._make_span_field(record["target"]["span" + str(i + 1)], text_field, 1)]
             )
-        example["labels"] = ListField(
-            [
-                MultiLabelField(
-                    [record["label"]],
-                    label_namespace=self._label_namespace,
-                    skip_indexing=True,
-                    num_labels=self.n_classes,
-                )
-            ]
+        example["labels"] = LabelField(
+            record["label"], label_namespace="labels", skip_indexing=True
         )
         return Instance(example)
 
@@ -2776,10 +2769,10 @@ class WinogradCoreferenceTask(SpanClassificationTask):
             return ones.index_select(0, batch)
 
         binary_preds = make_one_hot(logits, depth=2)
-        # Make label_ints a batch_size list of labels
-        label_ints = torch.argmax(labels, dim=1)
-        self.f1_scorer(binary_preds, label_ints)
-        self.acc_scorer(binary_preds.long(), labels.long())
+        labels_one_hot = make_one_hot(labels, depth=2)
+
+        self.f1_scorer(binary_preds, labels)
+        self.acc_scorer(binary_preds.long(), labels_one_hot.long())
 
     def get_metrics(self, reset=False):
         """Get metrics specific to the task"""
--- a/jiant/tasks/tasks.py
+++ b/jiant/tasks/tasks.py
@@ -2748,31 +2748,14 @@ class WinogradCoreferenceTask(SpanClassificationTask):
                 )
             else:
                 iters_by_split[split] = load_span_data(
-                    self.tokenizer_name, filename, label_fn=lambda x: 0 if x is False else 1
+                    self.tokenizer_name, filename, label_fn=lambda x: int(x)
                 )
         self._iters_by_split = iters_by_split
 
     def update_metrics(self, logits, labels, tagmask=None):
         logits, labels = logits.detach(), labels.detach()
-
-        def make_one_hot(batch, depth=2):
-            """
-            Creates a one-hot embedding of dimension 2.
-            Parameters:
-            batch: list of size batch_size of class predictions
-            Returns:
-            one hot encoding of size [batch_size, 2]
-            """
-            ones = torch.sparse.torch.eye(depth)
-            if torch.cuda.is_available():
-                ones = ones.cuda()
-            return ones.index_select(0, batch)
-
-        binary_preds = make_one_hot(logits, depth=2)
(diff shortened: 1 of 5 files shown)


jiant/utils/data_loaders.py
```diff
@@ -41,6 +41,9 @@ def load_span_data(tokenizer_name, file_name, label_fn=None, has_labels=True):
     rows = rows.apply(lambda x: realign_spans(x, tokenizer_name), axis=1)
     if has_labels is False:
         rows["label"] = False
+    else:
+        if label_fin is not None:
```
W4ngatang: this will break ?

jiant/tasks/tasks.py
```diff
@@ -2753,12 +2754,11 @@ def load_data(self):
...
                 )
             else:
-                iters_by_split[split] = load_span_data(self.tokenizer_name, filename)
+                iters_by_split[split] = load_span_data(
+                    self.tokenizer_name, filename, label_fn=lambda x: 0 if x is False else 1
+                )
         self._iters_by_split = iters_by_split
 
-    def get_all_labels(self):
-        return ["False", "True"]
```
W4ngatang: Make sure that the prediction files created in `evaluate.py` still have whatever the output labels should be

jiant/tasks/tasks.py
```diff
@@ -2340,9 +2340,10 @@ def make_instance(self, record, idx, indexers) -> Type[Instance]:
         example["labels"] = ListField(
             [
                 MultiLabelField(
-                    [str(record["label"])],
+                    [record["label"]],
```
W4ngatang: use just a LabelField w/o ListField or MultiLabelField

jiant/tasks/tasks.py
pruksmhc: yep checked!

jiant/utils/data_loaders.py
```diff
@@ -41,6 +41,9 @@ def load_span_data(tokenizer_name, file_name, label_fn=None, has_labels=True):
     rows = rows.apply(lambda x: realign_spans(x, tokenizer_name), axis=1)
     if has_labels is False:
         rows["label"] = False
+    else:
+        if label_fn is not None:
+            rows["label"] = rows["label"].apply(lambda x: label_fn(x))
```
W4ngatang: What about when `label_fn` is None? If you're doing nothing in that case, just collapse into elif

W4ngatang: Also, can we make labels 0 when there are no labels? More closely matches your usage with the label fn

jiant/tasks/tasks.py
```diff
@@ -2753,12 +2747,11 @@ def load_data(self):
                     self.tokenizer_name, filename, has_labels=False
                 )
             else:
-                iters_by_split[split] = load_span_data(self.tokenizer_name, filename)
+                iters_by_split[split] = load_span_data(
+                    self.tokenizer_name, filename, label_fn=lambda x: 0 if x is False else 1
```
W4ngatang: Just `int()` ?

sleepinyourhat: Equivalent to lambda x: int(x), if we know x is True or False.

jiant/utils/data_loaders.py
sleepinyourhat: `lambda x: f(x)` is identical to `f(x)`—replace.

I can't tell if this is Python apply (which you could replace with just `label_fn(rows ...` or Pandas apply, which does seem to do something.

W4ngatang: I think it's pandas, but you can still replace with just f(x)

jiant/tasks/tasks.py
```diff
@@ -2776,10 +2769,10 @@ def make_one_hot(batch, depth=2):
             return ones.index_select(0, batch)
 
         binary_preds = make_one_hot(logits, depth=2)
-        # Make label_ints a batch_size list of labels
-        label_ints = torch.argmax(labels, dim=1)
-        self.f1_scorer(binary_preds, label_ints)
-        self.acc_scorer(binary_preds.long(), labels.long())
+        labels_one_hot = make_one_hot(labels, depth=2)
```
W4ngatang: why do we need to make things one-hot?

W4ngatang: Two other things:
1) take `make_one_hot` out of `update_metrics`? Seems like a weird place to define it
2) The usage of cuda on line 2767 will break if the user has a GPU available but decides to run on CPU, I think. `ones` will be a cuda tensor but everything else will be on GPU.

pruksmhc: We need to make this one hot because acc_scorer expects the labels and the predictions to be the same dimension. Let me try without one hot though!

jiant/utils/data_loaders.py
pruksmhc: You need to have the lambda x since otherwise x is not defined in f(x)

jiant/tasks/tasks.py
```diff
@@ -2753,33 +2747,15 @@ def load_data(self):
...
+                )
         self._iters_by_split = iters_by_split
 
-    def get_all_labels(self):
-        return ["False", "True"]
-
     def update_metrics(self, logits, labels, tagmask=None):
         logits, labels = logits.detach(), labels.detach()
-
-        def make_one_hot(batch, depth=2):
```
pruksmhc: acc_scorer basically necessitates the preds and the labels to be the same dimension. Now that the labels are integers (due to the change from multilabel -> label), we can get rid of the one hot vectorization and pass in both as a list of integers of dimension [batch_size, 1].

PR #834, opened 2019-07-10
Fix get labels

pruksmhc:
WSC was crashing if it ran by itself (as opposed to running another experiment and then WSC right after). 
This also exposes a problem, in that all runs in an experiment directory share the same vocabulary.


W4ngatang:
What is this supposed to fix and how? It looks like we're not setting `task.all_labels` anywhere.

pruksmhc:
We don't need to use this for span classification because we don't need to index the labels (due to the previous PR's change). WSC was crashing before because we don't set self.all_labels anymore.

PR #835, opened 2019-07-10
AllenNLP F1_scorer update 

pruksmhc:
Allennlp F1Scorer expects a different input shape to 0.8.1 - updating that here for WSC.


jiant/modules/span_modules.py
```diff
@@ -133,7 +133,7 @@ def forward(
             out["loss"] = self.compute_loss(logits, batch["labels"], task)
             predictions = self.get_predictions(logits)
             tagmask = batch.get("tagmask", None)
-            task.update_metrics(predictions, batch["labels"], tagmask=tagmask)
+            task.update_metrics(logits, batch["labels"], tagmask=tagmask)
```
W4ngatang: delete `predictions` call above

PR #838, opened 2019-07-12
Make sure task-specific checkpoints are cleared when replaced

### LINKED ISSUE #837: Task-specific checkpoints preserved when keep_all_checkpoints = 0
This means that a single SuperGLUE fine-tuning run can take hundreds of GB. Investigating.

sleepinyourhat:
Fixes #837


sleepinyourhat:
Yep—that's the key change.

Issue #836, opened 2019-07-12
boolq task data preprocessing


desperadoola:
https://github.com/nyu-mll/jiant/blob/c713766fdbbb1dd50b4fc4afd558eaf9d5c6f948/jiant/tasks/tasks.py#L2821

Correct me if I'm wrong. Here we concatenate the sequence as [CLS] + context + [CLS] question + [SEP], instead of [CLS] + context + [SEP] question + [SEP], any reason doing that?

sleepinyourhat:
That does sound fishy, though I don't know this part of the code well. @W4ngatang, mind taking a look?

W4ngatang:
Oh yeah, that's a brainfart. I'll push a fix, though I expect numbers won't change that much (we were already in-line with the BERT baseline in the original work). Thanks for the catch !

PR #841, opened 2019-07-15
Fix bug in BERT preprocessing

W4ngatang:
Addresses #836 by swapping `sent1[:-1] + sent2` -> `sent1 + sent2[1:]`.


sleepinyourhat:
Worth making a 2.0.2 release? I can see arguments either way.

sleepinyourhat:
Er, 1.0.2.

PR #844, opened 2019-07-17
Update GCP set-up scripts

iftenney:
There's a nice pre-built image with PyTorch, Anaconda3, and most of the other things you need, so set-up is greatly simplified.

@sleepinyourhat I tested this with a fresh workstation on the new GCP environment.

Also delete old scripts that reference obsolete JSALT images and paths.


iftenney:
Good idea. I'm planning some more cleanup of the Dockerfile; will send a follow-up PR.

iftenney:
Thanks!

PR #851, opened 2019-07-17
Clean up edge probing tasks and configs

iftenney:
Remove a lot of old, unused task configs, and simplify the names of the remaining ones.

Also remove the `is_symmetric` option from `EdgeProbingTask`, because this was always set to False. Reintroduce as a hyperparameter, since it probably makes more sense there anyway.


jiant/modules/edge_probing.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, task, d_inp: int, task_params):
         self.loss_type = task_params["cls_loss_fn"]
         self.span_pooling = task_params["cls_span_pooling"]
         self.cnn_context = task_params["edgeprobe_cnn_context"]
-        self.is_symmetric = task.is_symmetric
+        self.is_symmetric = task_params.get("edgeprobe_symmetric", False)
```
sleepinyourhat: Why .get? It's in defaults, so it won't be unset.

sleepinyourhat: There are other mechanisms around for task-specific params in trainer setup, FWIW.

iftenney: Good point. Fixed.

What's the mechanism you're thinking of? (I haven't kept up with all the new features)

iftenney: NM, I thought there was a PR that changed this a while back, but looks like it's still the old codepath in `models.py`.

sleepinyourhat: Here! https://github.com/nyu-mll/jiant/blob/master/jiant/utils/config.py#L102

It's recent-ish.

PR #852, opened 2019-07-18
Clean up Kubernetes configs

iftenney:
- Use jsonnet for configuration instead of a nightmarish mess of shell interpolation.
- Move all cluster configuration to a single file that's easy to document.
- Update configuration for Sam's current GCP project


gcp/kubernetes/README.md
```diff
@@ -8,6 +8,23 @@ All the instructions below assume Ubuntu Linux running on a Google Compute Engin
 
 # Set-Up Instructions
 
+## jsonnet
+Several of the Kubernetes scripts use [jsonnet](https://jsonnet.org/) for configuration. To install the jsonnet command-line utility, you need to build it from source and add it to your system path:
+```
+git clone https://github.com/google/jsonnet.git jsonnet
```
sleepinyourhat: Not ideal. If conda/environment.yml handles this, say so? (I assume people will be installing the dependencies on the machine they use for GCP setup, right?) If not, say why not?

iftenney: You only need this on the workstation that's running Kubernetes jobs (e.g. with run_batch.sh). 

There are Python bindings available via `pip`, but I couldn't find an easier way to install the command-line version that we need here.

Issue #360, opened 2018-08-30
Demo runs not deterministic anymore


iftenney:
I'm now seeing variation on the order of +/- 1% between demo runs(demo.conf and edgeprobe_demo.conf) with identical configuration. When I last checked this a few weeks ago, I would get exactly the same numbers each time.

Did we make any recent changes that affect the random seeds? It's also possible that I've messed up the dev environment on my workstation. I'll follow up here with more tests.

pyeres:
Heads up: some run-to-run variation was smoothed out w/ #991. Is this still an issue for anyone?

pyeres:
As part of a recent PR I compared the logs for demo.conf before and after a change: [here](https://github.com/nyu-mll/jiant/pull/1080#issuecomment-620955946).

The results were the same, so I'm closing this issue as resolved.

PR #853, opened 2019-07-18
Use environment.yml in Dockerfile

iftenney:
Instead of installing individual packages, just use conda environments inside the Docker image to keep versions synchronized with non-Docker usage.

Also, remove hard-coded paths from the Dockerfile. For local runs you can pass these with -e, and for Kubernetes these are defined in `jiant_env.libsonnet` and handled appropriately in the job YAML files.

Kubernetes run scripts should behave identically as before, unless you were doing something weird with environment vars.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,81 +15,50 @@
 # Use CUDA base image.
 FROM nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04
 
-# Add Tini to handle init.
-# TODO: see if we still need this? More recent docker might have this built-in.
-ENV TINI_VERSION v0.18.0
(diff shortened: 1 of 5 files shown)


iftenney:
Don't merge this yet, I'm getting some inconsistencies between sts-b results on Kubernetes versus locally; trying to debug.

sleepinyourhat:
Any update? GCP is broken at head, since Dockerfile is missing at least one dependency (jsondiff).

iftenney:
Ok: as far as I can tell, it's just random variation (demo.conf on sts-b seems to vary a lot, between 0.160 and 0.280 or so) and I just had a couple bad ones in a row. I don't see any systematic differences between master and this branch, or between p100s and k80s.

Worth revisiting #360 so we can have a reliable integration test?

PR #858, opened 2019-07-21
Update edge probing runner and tokenization scripts

iftenney:
Fix some config and path issues. BERT experiments should work now both on workstation and Kubernetes.


gcp/kubernetes/run_batch.sh
```diff
@@ -8,7 +8,7 @@
 #    - the resources defined in templates/jiant_env.libsonnet are correct
```
sleepinyourhat: Is this just a copy of your other PR? If so, merge that first and confirm no conflicts.

iftenney: Done.

PR #860, opened 2019-07-22
Fix notebook support

iftenney:
Notebooks don't play nice with virtualenv unless you register a kernel. This adds instructions on how to do that, as well as the necessary ipykernel package.

Also add bokeh to the environment for plotting (in particular, to reproduce plots from the edge probing papers).


environment.yml
```diff
@@ -9,6 +9,8 @@ dependencies:
     - nltk=3.2.5
     - scikit-learn=0.19.1
     - pandas=0.23.0
+    - bokeh=1.2.0
```
sleepinyourhat: Mind adding comments here on where this is used? I want to make it easier for users to skip/ignore/delete items here in case conflicts appear later. (I'm doing the same in the XLNet PR.)

tutorials/setup_tutorial.md
```diff
@@ -38,11 +38,22 @@ before running any `jiant` code. (To deactivate run: `source deactivate`)
 
 Some requirements may only be needed for specific configurations. If you have trouble installing a specific dependency and suspect that it isn't needed for your use case, create an issue or a pull request, and we'll help you get by without it.
 
-You will also need to install dependencies for `nltk` if you do not already have them:
+You will also need to install dependencies for `nltk` and `spacy' if you do not already have them:
```
sleepinyourhat: Mention where SpaCy is needed. It's not part of the main codeflow for lots of models, so some users can skip this.

iftenney: Done.

environment.yml
iftenney: Done. Added comments for some older ones too.

Though, since this is a fully-specified environment, there really shouldn't be any package conflicts - that's on us to fix.

PR #864, opened 2019-07-22
Probe BERT by layer

iftenney:
Probe BERT by layer

- Modifications to the BERT wrapper to handle scalar mixing and layer selection appropriately
- Experiment configs for edge probing, as used in https://arxiv.org/abs/1905.05950
- Fix minor path bugs in edge probing analysis code
- Update edge probing sample analysis notebook showing how to do per-layer analysis and extract mixing scalars


jiant/bert/utils.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, args, cache_dir=None):
             args.input_module, cache_dir=cache_dir
         )
         self.embeddings_mode = args.bert_embeddings_mode
+        self.num_layers = self.model.config.num_hidden_layers
+        if args.bert_max_layer >= 0:
+            self.max_layer = args.bert_max_layer
+        else:
+            self.max_layer = self.num_layers
+        assert self.max_layer <= self.num_layers
```
sleepinyourhat: Move this inside the first conditional clause?

probing/get_scalar_mix.py
```diff
@@ -23,13 +23,13 @@
...
 import logging as log
 
-log.basicConfig(format="%(asctime)s: %(message)s", datefmt="%m/%d %I:%M:%S %p", level=log.INFO)
-
 import pandas as pd
 import torch
 
 from typing import List, Tuple, Iterable, Dict
 
+log.basicConfig(format="%(asctime)s: %(message)s", datefmt="%m/%d %I:%M:%S %p", level=log.INFO)
```
sleepinyourhat: Why this move?

jiant/bert/utils.py
iftenney: It's a little safer to enforce globally?

probing/get_scalar_mix.py
iftenney: Formatter was complaining.

iftenney:
Will update readme in a separate PR (there's a bunch of other documentation I need to add). Thanks for the review!

jiant/bert/utils.py
sleepinyourhat: Up to you, but seems more confusing then helpful. The pattern is `x = y; assert x >= y`.

Issue #862, opened 2019-07-22
Using multiple tokenizers in the same `exp_dir` can produce broken results without an error message


sleepinyourhat:
Currently, if you run a bert-base-cased run in some exp_dir, then a bert-base-uncased run, you'll get this odd behavior, where the model reads the raw data again, but doesn't index it again:

```
07/22 02:54:40 PM: Loading tasks...
07/22 02:54:40 PM: Writing pre-preprocessed tasks to /Users/srbowman/exp/jiant-demo/
07/22 02:54:40 PM: 	Creating task commitbank from scratch.
07/22 02:54:40 PM: 	Loading Tokenizer bert-base-uncased
07/22 02:54:41 PM: loading file https://s3.amazonaws.com/models.huggingface.co/bert/bert-base-uncased-vocab.txt from cache at /Users/srbowman/.cache/torch/pytorch_transformers/26bc1ad6c0ac742e9b52263248f6d0f00068293b33709fae12320c0e35ccfbbb.542ce4285a40d23a559526243235df47c5f75c197f04f37d1a0c124c32c9a084
07/22 02:54:41 PM: 	Finished loading CommitmentBank data.
07/22 02:54:41 PM: 	Task 'commitbank': |train|=250 |val|=56 |test|=250
07/22 02:54:41 PM: 	Creating task mrpc from scratch.
07/22 02:54:46 PM: 	Finished loading MRPC data.
07/22 02:54:46 PM: 	Task 'mrpc': |train|=3668 |val|=408 |test|=1725
07/22 02:54:46 PM: 	Creating task sst from scratch.
07/22 02:55:00 PM: 	Finished loading SST data.
07/22 02:55:00 PM: 	Task 'sst': |train|=67349 |val|=872 |test|=1821
07/22 02:55:00 PM: 	Creating task sts-b from scratch.
07/22 02:55:03 PM: 	Finished loading STS Benchmark data.
07/22 02:55:03 PM: 	Task 'sts-b': |train|=5749 |val|=1500 |test|=1379
07/22 02:55:03 PM: 	Finished loading tasks: commitbank mrpc sst sts-b.
07/22 02:55:03 PM: Loading token dictionary from /Users/srbowman/exp/jiant-demo/vocab.
07/22 02:55:03 PM: 	Loaded vocab from /Users/srbowman/exp/jiant-demo/vocab
07/22 02:55:03 PM: 	Vocab namespace chars: size 105
[...]

PR #866, opened 2019-07-24
Fixing multiple tokenizations in same experiment directory bug

pruksmhc:
https://github.com/nyu-mll/jiant/issues/862


jiant/preprocess.py
```diff
@@ -282,9 +284,9 @@ def build_tasks(args):
         setattr(task, "_classifier_name", task_classifier if task_classifier else task.name)
 
     tokenizer_names = {task.name: task.tokenizer_name for task in tasks}
-    assert len(set(tokenizer_names.values())) == 1, (
-        f"Error: mixing tasks with different tokenizers!" " Tokenizations: {tokenizer_names:s}"
-    )
```
sleepinyourhat: Out of curiosity, how could this ever be triggered? If we're allowing people to set task += {tokenizer_name ...}, we should just stop allowing that.

W4ngatang:
Should we just pass around the `Task` object rather than extract attributes? I'm fine either way on this.

jiant/preprocess.py
```diff
@@ -46,13 +46,15 @@
 ALL_SPLITS = ["train", "val", "test"]
 
 
-def _get_serialized_record_path(task_name, split, preproc_dir):
+def _get_serialized_record_path(task_name, tokenizer_name, split, preproc_dir):
     """Get the canonical path for a serialized task split."""
-    serialized_record_path = os.path.join(preproc_dir, "{:s}__{:s}_data".format(task_name, split))
+    serialized_record_path = os.path.join(
+        preproc_dir, "{:s}__{:s}_{:s}_data".format(task_name, tokenizer_name, split)
```
iftenney: Use double-underscores for the second and third separators as well? I don't think we rigorously enforce that names don't contain underscores elsewhere.

iftenney:
> Should we just pass around the `Task` object rather than extract attributes? I'm fine either way on this.

Passing attributes is safer (pass only what you need).

sleepinyourhat:
(See comment on #866 - If this does enforce one tokenization per task in the directory, then it doesn't solve the problem. I hadn't noticed that that was an expectation.)

W4ngatang:
Does this actually solve the issue? I think it will still break because if we find a cached vocabulary, we load it, even if it doesn't contain a vocabulary relevant to the tokenization we have now, and then when we index with this "vocabulary" (really just PAD and UNK), we get all UNKs.

pruksmhc:
The thinking behind this is that it would be a way for users to be able to structure experiment directories how they want (for example, having on experiment be made up of a task, with various runs representing various preprocessing). Ah, I do see the way vocab is currently handled must also be changed. 
In this case, looks like its a bit more involved to actually change the way we're currently handling experiments and tokenization, so I'll change this to throwing asserts if there are multiple tokenizations being used in an experiment dir.

jiant/preprocess.py
```diff
@@ -396,7 +396,19 @@ def _get_task(name, args, data_path, scratch_path):
...
+
+    task_pkls = glob.glob(os.path.join(scratch_path, "tasks", "%s*" % name))
     pkl_path = os.path.join(scratch_path, "tasks", f"{name:s}.{args.tokenizer:s}.pkl")
+
+    # There should only exist one task pickle per experiment directory with only one
+    # preprocessing, thus check that this is the case if the task pickle directory exists.
+    assert (
+        not os.path.isdir(os.path.join(scratch_path, "tasks"))
+        or len(task_pkls) == 0
+        or args.tokenizer in task_pkls[0]
```
sleepinyourhat: It seems plausible that for future models, one tokenizer name might be a substring of another, (maybe 'bert-base-cased' and 'multilingual-bert-base-cased'), in which case this check will pass. Mind extracting the exact substring? (Can you do this with pkl_path?)

iftenney:
re: experiment directories: I think it's only an abstraction worth supporting if it allows meaningful sharing of preprocessing, model weights, etc. Otherwise, the user can always just nest them inside their own directory structure, and jiant need not be aware of (nor try to avoid) the other experiments. (This is what I do for edge probing: https://github.com/nyu-mll/jiant/blob/master/scripts/edgeprobing/kubernetes_run_all.sh - there's a `project_dir` with many `exp_dir` inside it, each of which contains exactly one run.

jiant/preprocess.py
iftenney: +1. This kind of bug happens more than you think :)

sleepinyourhat:
Checking in: How close is this to ready?

sleepinyourhat:
Could you either finish or close this this week?

sleepinyourhat:
@pruksmhc - Mind taking a look at this sometime soon?

Issue #747, opened 2019-06-25
Add option to delete checkpoints when done


sleepinyourhat:
With BERT, it's increasingly common to run experiments that take minutes, but produce multiple GB of checkpoint files. When these experiments are quick preliminary investigations, rather than final runs for publication/analysis, it's rarely worth saving the checkpoints.

We can do this manually on the command line after each experiment, but it'd be nice to bake this in by adding an override—something like 'delete_checkpoints_when_done'—that identifies and removes the files automatically.

PR #867, opened 2019-07-24
Add option to automatically delete checkpoints when done.

sleepinyourhat:


sleepinyourhat:
#747

sleepinyourhat:
Yeah, oops. For your comment, it can be a real problem. Experiments involving fine-tuning BERT-Large models on several data-poor tasks can produce ~70GB of checkpoints in a few hours. Since you have size of bert * 2 (model params + optimizer params) * 2 (best + most recent) * num_tasks.

Issue #863, opened 2019-07-22
Check max_epochs stopping criterion at every step 


sleepinyourhat:
Currently, if we use the BertAdam optimizer and set max_epochs, there'll be a few steps in the last pass where we've gone over the max_epochs limit, because we only check the limit at the validation stops. At this point, the optimizer sets the LR to zero (harmless), and fills the logs with warnings (slightly harmful).

We should check that max_epochs hasn't been exceeded at every training step to avoid this.

PR #868, opened 2019-07-24
Update logic surrounding max_vals and max_epochs

sleepinyourhat:
Currently, bert_adam and other optimizers that need an upper bound on total training steps only get that upper bound when max_epochs is set, and the bound can be a bit off because we won't actually stop training until a validation is hit. With this change, that bound should be set correctly using whichever is more restrictive of max_epochs or max_vals. Addresses the main issue behind #863, and should be sufficient for pretty much all use cases.


config/defaults.conf
```diff
@@ -172,6 +172,13 @@ patience = 5  // Patience in early stopping. Training will stop if performance d
               // all in patience + 1 validations.
 keep_all_checkpoints = 0  // If set, keep checkpoint files from every validation. Otherwise, keep
                           // only the best and (if different) most recent.
+delete_checkpoints_when_done = 0  // If set, delete *all* saved checkpoints when the run is
```
W4ngatang: is it really space consuming to keep just the best (or last N) checkpoints around? Are they 1GB each?

W4ngatang:
Is the PR title accurate? Didn't see changes to `max_vals` or `max_epochs`

sleepinyourhat:
Eek. PTAL now.

W4ngatang:
Oh, I totally goofed and forgot to delete the task-specific optimization params when I deleted task-specific optimizers. Do you want to also blow those up in this PR?

sleepinyourhat:
Sure. Mind pushing to this branch?

sleepinyourhat:
Eh, mind just making an issue out of it?

Issue #446, opened 2018-12-21
Multi-GPU training support


sleepinyourhat:

sleepinyourhat:
cc @iftenney

sleepinyourhat:
It looks like the Allen fundamentals all support this, though with some caveats: https://github.com/allenai/allennlp/issues/2127

This is becoming increasingly urgent with BERT Large and XLNet—we can't use them at full sequence length on commodity GPUs.

HaokunLiu:
May I ask what’s the status of this thread? Do you need some extra hands?

sleepinyourhat:
Check with @pruksmhc, but I don't think anything is happening with this now. If you want to try it, please do!

pruksmhc:
@HaokunLiu I'm actually working on this and planning to release a PR early next week!

HaokunLiu:
@pruksmhc  Oh, that's great. Thanks.

PR #873, opened 2019-08-04
 Implementing Data Parallel

pruksmhc:
https://github.com/nyu-mll/jiant/issues/446

This is tested on BERT models and also GloVE embeddings, with all the major types of tasks. 

Below is the table for BERT models. 

| Task | Batch size with 1 GPU | Batch size with 2 GPUs | Time per 100 steps seconds (1 GPU) | Time per steps seconds (2 GPUs)|
| ------------- | ------------- | ------------- | ------------- | ------------- |
| ReCoRDl  | 11 | 22 | 167.55 | 187.90 |
| BoolQ  | 14 | 28 | 122.2 | 146.6 |

These results were on a machine with 2 P40s.


jiant/models.py
```diff
@@ -204,11 +201,10 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
             sep_embs_for_skip=args.sep_embs_for_skip,
             cove_layer=cove_layer,
         )
-        d_sent = 0
+        d_sent = 0  # skip connection added below
+        log.info("No shared encoder (just using word embeddings)!")
```
sleepinyourhat: It looks like you might be working from a stale version of master.

jiant/trainer.py
```diff
@@ -693,7 +686,11 @@ def clip_function(grad):
                 if self._TB_dir is not None:
                     self._metrics_to_tensorboard_val(n_step, all_val_metrics)
                 log.info(f"Global learning rate: {self._optimizer.param_groups[0]['lr']}")
-                elmo_params = self._model.get_elmo_mixing_weights(tasks)
+<<<<<<< HEAD
+                elmo_params = get_model_attribute("get_elmo_mixing_weights")(tasks)
+=======
+                elmo_params = get_model_attribute(self._model, "get_elmo_mixing_weights")(tasks)
+>>>>>>> 51701999fe9b1f4a6688f2babb900021c093b3eb
```
sleepinyourhat: :D

sleepinyourhat:
Exciting!

jiant/trainer.py
ypruksachatkun-asapp: Oh whoops

pruksmhc: Oh whoops

sleepinyourhat:
Just checking in: Do you have a timeline in mind for the PR?

pruksmhc:
I'm going to take it out of WIP today, and going to try to get it pushed it in this week/next week.

jiant/modules/span_modules.py
```diff
@@ -78,6 +80,7 @@ def forward(
         sent_mask: torch.Tensor,
         task: Task,
         predict: bool,
+        cuda_device: int,
```
sleepinyourhat: Is there a way to have this live with trainer or somewhere else external, to avoid having to pass it around?

jiant/utils/utils.py
```diff
@@ -33,6 +33,20 @@
...
 
 
+def get_output_attribute(out, attribute_name):
+    if torch.cuda.device_count() > 1:
+        return out[attribute_name].sum()
+    else:
+        return out[attribute_name]
+
+
+def get_model_attribute(model, attribute_name):
```
sleepinyourhat: This one isn't completely clear. Add a docstring explaining what this does/fixes.

jiant/utils/utils.py
```diff
@@ -355,9 +369,7 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
 
 def get_batch_size(batch):
     """ Given a batch with unknown text_fields, get an estimate of batch size """
-    batch_field = batch["inputs"] if "inputs" in batch else batch["input1"]
-    keys = [k for k in batch_field.keys()]
-    batch_size = batch_field[keys[0]].size()[0]
+    batch_size = batch["idx"].size()[0]
```
sleepinyourhat: This previously supported batches without "idx". I don't think that's a case that we'll actually encounter, but have you double checked that this is safe?

jiant/utils/utils.py
```diff
@@ -33,6 +33,20 @@
 _MOSES_DETOKENIZER = MosesDetokenizer()
 
 
+def get_output_attribute(out, attribute_name):
+    if torch.cuda.device_count() > 1:
+        return out[attribute_name].sum()
```
W4ngatang: Is it always the case that you can call sum (or want to call it) on the output? For loss that makes sense, but not sure it makes sense in general.

jiant/modules/span_modules.py
```diff
@@ -116,7 +118,8 @@ def forward(
             se_projs.append(se_proj)
 
         span_embs = torch.Tensor([]).cuda() if torch.cuda.is_available() else torch.Tensor([])
-        out["n_exs"] = batch_size
+        out["n_exs"] = torch.tensor(get_batch_size(batch))
```
W4ngatang: just make it a single line ?

jiant/modules/span_modules.py
W4ngatang: :+1:, or at least live with the top level model. Alternatively, is there a way to access the device from `self`, e.g. `self.device`?

I don't totally understand why you need to move `n_exs` to device

jiant/models.py
```diff
@@ -265,6 +266,9 @@ def build_model(args, vocab, pretrained_embs, tasks):
     model = MultiTaskModel(args, sent_encoder, vocab)
     build_task_modules(args, tasks, model, d_task_input, d_emb, embedder, vocab)
     model = model.cuda() if args.cuda >= 0 else model
+    if torch.cuda.device_count() > 1:
```
W4ngatang: worthwhile to have a flag like `use_distributed` ? Or is the expectation that users mask devices on the commandline ?

W4ngatang:
Have you compared run times on 1 vs 2 (or any N > 1) GPUs ? When I used DataParallel, I got no speed ups (though I could use larger batch size), which apparently is due to single threading of the data loader. The advice I've heard is to use DistributedDataParallel, even for single node multi-GPU, but that would be substantially more annoying to code.

jiant/modules/span_modules.py
pruksmhc: Yep, I'll move it up. 
As to your question, what data parallel does is that it makes copies of the model and then sends subsets of the batches to each of these copies (in each of the GPUs), then aggregates all the out dictionaries of all of these copies in one GPU. Thus, everything in the out dictionary needs to be a tensor in cuda.

jiant/utils/utils.py
pruksmhc: Yes, I checked that the current models being supported have "idx" but I'll double check that and also make sure it's in the tutorial.

pruksmhc: I changed it because the previous way was actually not safe because not all tasks have "inputs" (such as some of the QA models).

jiant/utils/utils.py
pruksmhc: Yes, that's true. For our current use cases (loss, n_exes) it's true, but for other people who use this they might have to tweak this function to suit their needs.

jiant/models.py
pruksmhc: Because we only check for cuda_device a few times in the code I don't think it's quite necessary to have a flag for this, and yes, they can use the commandline. However, if users are on GPU, they shouldn't need to change anything.  This should automatically detect if there's more than 1 GPU, and use it if so.

pruksmhc:
Alright, in that case I'll just leave this as DataParallel, and if we want to implement DistributedDataParallel in the future we can also add that in another PR.

jiant/__main__.py
```diff
@@ -1,7 +1,5 @@
 """Train a multi-task model using AllenNLP
```
sleepinyourhat: While you're editing this, mind replacing this line. It's obviously wrong...

jiant/__main__.py
```diff
@@ -1,7 +1,5 @@
...
-
```
sleepinyourhat: Leave in all these newlines here and below.

jiant/models.py
sleepinyourhat: This isn't usable as-is: We have an explicit flag that asks users to specify the GPU (cuda, see comment in defaults.conf), and it seems that we're now ignoring it.

There are reasonable/common use cases in which:
- People will want to use all the GPUs on a machine. (Obvious, I guess.)
- People will want to use a specific subset of GPUs on a machine, which might not be identified by the env variable. This is common on clusters managed with older batch managers that don't track GPUs—there's often some other automatic or manual system that informs you of what GPUs you can use.
- People will want to use none of the available GPUs.

I think the simplest option would be to leave several options for CUDA:
- 'auto': Default; use all available GPUs.
- 'none' or '-1': Disable CUDA.
- '0,2,3', etc.: Use GPUs 0, 2, and 3. List would be parsed the same way as in pretrain_tasks.

jiant/utils/utils.py
sleepinyourhat: This seems okay to me as is, but make sure to add a docstring that makes it clear that you use sum, and that this won't be appropriate for non-summed attributes like accuracy.

jiant/utils/utils.py
sleepinyourhat: I'm still not sure I understand. Add a short explanation for why this is necessary...

pruksmhc:
Tested this on no CPU, all GPUs, a subset of GPUs, and one GPU on demo.conf. Currently running experiments on the other major types of tasks.

jiant/__main__.py
```diff
@@ -100,7 +101,7 @@ def setup_target_task_training(args, target_tasks, model, strict):
     args: Params object
     target_tasks: list of target Task objects
     mdoel: a MultiTaskModel object
-
+
```
pruksmhc: I'll remove this addition.

sleepinyourhat:
How close is this to being ready? I'll likely need it soon myself...

pruksmhc:
Ready!

sleepinyourhat:
Cool. Back to you @W4ngatang. What do the speed results look like so far?

pruksmhc:
The current result on demo.conf
41.141s with no GPUs
46.387s with two GPUs

Right now, with no GPU, on BoolQ, it can accomodate a batch size of up to 32, but with two GPUs, it cannot accommodate that. It seems as though there's pretty significant imbalance in GPU usage right now. I'm debugging (right now waiting for a multi-GPU instance on HPC),but we may have to just switch to DistributedDataParallel which seems to be just "DataParallel but empirically better for running experiments"

sleepinyourhat:
Hrm—this doesn't really tell us what we want yet:

- Make sure to use a big model (superglue-bert?). For small models, data
transfer time will dominate, and we won't see a difference.
- Make sure the batch size is near the limit of each GPU.
- Compare 1 v 2 (or better 1 v 2 v 4) on the same machine. CPU optional.

pruksmhc:
@sleepinyourhat by 1 v 2 do you mean batch size of size 1 versus 2?

sleepinyourhat:
Oops—1 v 2 v 4 GPUs.

pruksmhc:
Alright, seems like with 2 GPUs, boolq runs out of memory with BERT with batch size 4 (as opposed to even 1 GPU, which can handle that batch size). I'm a bit confused as to why it can handle less batch size than 1 GPU, since it seems as though it is splitting the batch size in half over the two GPUs - right now it's running out of memory at loss.backward() 
Seems like there's some reasoning here about model parallel and how DP doesn't support it. 
https://pytorch.org/tutorials/intermediate/ddp_tutorial.html

sleepinyourhat:
Hrm, something sounds broken. Are you investigating?

- DataParallel looks like a good starting point for what we want: Single machine, sychronized multi-GPU training for models that can fit on one GPU.
- If implemented correctly, it should act almost exactly like normal training, but with support for roughly twice the batch size (with double the GPU memory/capacity), and a slowdown in per step time (due to the added cross-GPU communication) that is not significant for big/slow models.

sleepinyourhat:
One quick idea for a possible issue: Does the batch size parameter set the real/total batch size or the batch size per GPU? If it's setting batch size per GPU, that would explain that strange behavior.

pruksmhc:
Yep, I'm currently investigating. No it does not, each GPU is getting 2 examples (when bs=4).

HaokunLiu:
> Alright, seems like with 2 GPUs, boolq runs out of memory with BERT with batch size 4 (as opposed to even 1 GPU, which can handle that batch size). I'm a bit confused as to why it can handle less batch size than 1 GPU, since it seems as though it is splitting the batch size in half over the two GPUs - right now it's running out of memory at loss.backward()

That's really strange. Did you specify which type to gpu to use? Sometimes Prince just give you GTX1080 with tiny memories.

pruksmhc:
I indeed have GTX 1080, but regardless, with DataParallel we should be seeing an increase in memory between 1 and 2 GPUs, regardless of GPU size.

pruksmhc:
Actually, I had made a mistake, and now, when I try with 1 vs 2 GPUs with 2 GTX1080, it is able to bear the same amount of batch size. 
When I try with 2 P40s, the difference is much more marked. Specifically, I am able to fit batch_size=28 with two P40s, when one GPU alone cannot even fit a batch_size of 19.

sleepinyourhat:
That's encouraging. Make sure you're maxing out the GPUs in both the 1-GPU and 2-GPU setups. If the max batch sizes aren't very close to double, something's wrong...

pruksmhc:
I narrowed down the upper bound for 1 GPU to batch_size = 14, and actually cuda=auto can handle batch_size=30.

pruksmhc:
For Record, with 2 GPUs, we can train with batch_size = 22, and with 1 GPU, with batch_size = 11, maxing out the GPUs.

jiant/trainer.py
```diff
@@ -30,10 +29,12 @@
     assert_for_log,
     find_last_checkpoint_epoch,
     check_for_previous_checkpoints,
+    get_output_attribute,
+    get_model_attribute,
 )  # pylint: disable=import-error
 
 
-def build_trainer_params(args, task_names, phase="pretrain"):
+def build_trainer_params(args, use_cuda, cuda_device, task_names, phase="pretrain"):
```
W4ngatang: Can we merge these ? e.g. if `use_cuda = bool(cuda_device < 0)`

jiant/trainer.py
```diff
@@ -326,7 +333,7 @@ def _setup_training(
                 biggest_batch_first=True,
             )
             tr_generator = iterator(task.train_data, num_epochs=None)
-            tr_generator = move_to_device(tr_generator, self._cuda_device)
+            tr_generator = tr_generator
```
W4ngatang: ?

just delete?

jiant/trainer.py
```diff
@@ -843,15 +857,15 @@ def _calculate_validation_performance(
         val_generator = BasicIterator(batch_size, instances_per_epoch=max_data_points)(
             task.val_data, num_epochs=1, shuffle=False
         )
-        val_generator = move_to_device(val_generator, self._cuda_device)
+        val_generator = val_generator
```
W4ngatang: delete

jiant/preprocess.py
```diff
@@ -412,6 +412,29 @@ def parse_task_list_arg(task_list):
     return task_names
 
 
+def parse_cuda_list_arg(cuda_arg):
```
W4ngatang: `preprocess.py` seems like an odd place for this ? Why not in main ? Or in some `options.py` file (like how fairseq does it -- it's a bit unintuitive at first to look in another file for the option definitions, but I kind of like it)

jiant/preprocess.py
```diff
@@ -412,6 +412,29 @@ def parse_task_list_arg(task_list):
...
+        return result_cuda, use_cuda
+    if isinstance(cuda_arg, int):
+        result_cuda = [cuda_arg]
+        if result_cuda[0] < 0:
+            use_cuda = 0
+    elif len(cuda_arg.split(",")) > 1:
+        for device_id in cuda_arg.split(","):
+            result_cuda.append(int(device_id))
+    else:
+        raise EnvironmentError(
```
W4ngatang: me be super annoying: `EnvironmentError` seems weird, maybe `ValueError` ?

jiant/utils/utils.py
```diff
@@ -34,6 +34,46 @@
 _MOSES_DETOKENIZER = MosesDetokenizer()
 
 
+def get_output_attribute(out, attribute_name, use_cuda):
+    """
+    This function handles processing/reduction of output for both 
+    DataParallel or non-DataParallel situations. 
+    For the case of multiple GPUs, This functil will
```
W4ngatang: function

jiant/modules/span_modules.py
```diff
@@ -106,7 +106,6 @@ def forward(
         """
         out = {}
         batch_size = sent_embs.shape[0]
-        out["n_inputs"] = batch_size
```
W4ngatang: meant to delete this ?

jiant/__main__.py
```diff
@@ -596,11 +602,14 @@ def main(cl_arguments):
         # Evaluate on target_tasks.
         for task in target_tasks:
             # Find the task-specific best checkpoint to evaluate on.
-            task_to_use = model._get_task_params(task.name).get("use_classifier", task.name)
+            task_to_use = get_model_attribute(model, "_get_task_params", use_cuda)(task.name).get(
```
W4ngatang: unrelated to this PR, but this line is super unreadable haha

jiant/__main__.py
```diff
@@ -377,7 +378,8 @@ def initial_setup(args, cl_args):
     random.seed(seed)
     torch.manual_seed(seed)
     log.info("Using random seed %d", seed)
-    if args.cuda >= 0:
+    if isinstance(args.cuda, int) and args.cuda >= 0:
```
W4ngatang: shouldn't this cuda check also happen if you're trying to auto or use multiple GPUs ?

jiant/models.py
```diff
@@ -286,9 +286,12 @@ def build_model(args, vocab, pretrained_embs, tasks):
     d_task_input = d_sent_output + (args.skip_embs * d_emb)
 
     # Build model and classifiers
-    model = MultiTaskModel(args, sent_encoder, vocab)
+    model = MultiTaskModel(args, sent_encoder, vocab, cuda_devices, use_cuda)
     build_task_modules(args, tasks, model, d_task_input, d_emb, embedder, vocab)
-    model = model.cuda() if args.cuda >= 0 else model
+    model = model.cuda() if use_cuda else model
+    if use_cuda:
+        model = nn.DataParallel(model, device_ids=cuda_devices).cuda()
```
W4ngatang: so if we use GPU, we always use DataParallel? Does that incur a big cost in the single GPU case?

jiant/models.py
```diff
@@ -1094,7 +1113,11 @@ def _lm_forward(self, batch, task, predict):
         pad_idx = self.vocab.get_token_index(self.vocab._padding_token, "tokens")
         b_size, seq_len = batch["targs"]["words"].size()
         n_pad = batch["targs"]["words"].eq(pad_idx).sum().item()
-        out["n_exs"] = (b_size * seq_len - n_pad) * 2
+        out["n_exs"] = (
```
W4ngatang: slight preference to just split this in two lines rather than have the redundancy

jiant/models.py
```diff
@@ -881,7 +886,9 @@ def _single_sentence_forward(self, batch, task, predict):
         classifier = self._get_classifier(task)
         logits = classifier(word_embs_in_context, sent_mask)
         out["logits"] = logits
-        out["n_exs"] = get_batch_size(batch)
+        out["n_exs"] = (
```
W4ngatang: any reason not to have `get_batch_size` return a tensor (and maybe even a cuda tensor) rather than repeat this everywhere ?

W4ngatang:
fwiw @sleepinyourhat @pruksmhc when I used DataParallel, the gains were surprisingly below 2x batch size because the main GPU still needs to aggregate gradient information, so you have underutilization on every other GPU. DistributedDataParallel I think gets around it, but the interface is quite different.

pruksmhc:
Yeah, I think for now we're just sticking with DataParallel.

jiant/models.py
pruksmhc: I _think_ for one DataParallel, it doesn't do anything (because there's only one GPU). I'll change this though.

jiant/models.py
pruksmhc: what do you mean split into two lines? Even if we split the if else into two lines it'll still be redundant to those few lines. I thought of moving it out of these task specific forward functions but different tasks have different ways of calculating n_exs so that wouldn't work.

jiant/modules/span_modules.py
pruksmhc: Yep -it's not used anywhere.

jiant/preprocess.py
pruksmhc: It's where the rest of the config parsing code lives. I'll move it over to options.py

jiant/trainer.py
pruksmhc: Oh whoops

jiant/trainer.py
pruksmhc: The alternative is to have a function that checks whether to use_cuda based on the cuda_device.

jiant/__main__.py
pruksmhc: parse_cuda_args_list converts auto -> range(len(devices)), basically using all devices on the machine.

jiant/models.py
pruksmhc: Good point

sleepinyourhat:
Thanks! Leaving this to @W4ngatang to review (unless he'd rather delegate), but again: Please post a precise time comparison before we merge.

jiant/utils/utils.py
```diff
@@ -361,15 +401,40 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
...
+def format_output(obj, cuda_devices):
+    """
+    Format output based on whether model is using DataParallel or not. 
+    DataParallel necessitates objects to be gathered into GPU:0 to have 
+    dimension 0. 
+    This function will be used for scalar outputs of model forwards 
+    such as loss and n_exs.
+    """
+    if isinstance(cuda_devices, list):
+        if isinstance(obj, torch.Tensor) is False:
```
W4ngatang: `if not isinstance(obj, torch.Tensor)` is more pythonic

jiant/utils/options.py
```diff
@@ -0,0 +1,41 @@
...
+        elif task_name == "superglue":
+            task_names.extend(ALL_SUPERGLUE_TASKS)
+        elif task_name == "none" or task_name == "":
+            continue
+        else:
+            task_names.append(task_name)
+    return task_names
+
+
+def parse_cuda_list_arg(cuda_list):
```
W4ngatang: structure seems a little odd, though workable. Why not

`elif isinstance(cuda_list, int):
    return cuda_list
elif "," in cuda_list:
    return [int(d) for d in cuda_list.split(",")]
else:
    raise ValueError(...)`

jiant/trainer.py
```diff
@@ -1034,8 +1046,7 @@ def _check_stop(self, val_n, stop_metric, tasks):
         return should_stop
 
     def _forward(self, batch, task=None):
-        tensor_batch = move_to_device(batch, self._cuda_device)
-        model_out = self._model.forward(task, tensor_batch)
+        model_out = self._model.forward(task, batch)
```
W4ngatang: not related to this PR, but we should probably delete this at some point, it's a hold-over from when I was importing this from AllenNLP but we don't seem to have a good reason to keep this extra step around

jiant/preprocess.py
```diff
@@ -52,7 +52,8 @@
 from jiant.tasks import REGISTRY as TASKS_REGISTRY
 from jiant.tasks.seq2seq import Seq2SeqTask
 from jiant.tasks.tasks import SequenceGenerationTask
-from jiant.utils import config, serialize, utils
+from jiant.utils import config, serialize, utils, options
```
W4ngatang: delete the `options` import ?

jiant/utils/utils.py
W4ngatang: what type would `obj` by in this case ?

jiant/utils/utils.py
```diff
@@ -361,15 +401,40 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
...
+            obj = torch.tensor(obj).cuda()
+        return obj.unsqueeze(0)
+    else:
+        return obj
```
W4ngatang: why don't we need to call `.cuda()` if we're using a single GPU?

jiant/utils/utils.py
```diff
@@ -361,15 +401,40 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
...
+
+
+def uses_cuda(cuda_devices):
```
W4ngatang: combine into a single clause ?

`if isinstance(cuda_devices, list) or (isinstance(cuda_devices, int) and cuda_devices >= 0): ...`

Or even a single line if you want to live with long lines lol

jiant/utils/utils.py
```diff
@@ -361,15 +401,40 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
     return params
 
 
-def get_batch_size(batch, keyword="input"):
+def format_output(obj, cuda_devices):
```
W4ngatang: so this should only receive `torch.Tensor`s? Can you update the function name, docstring, or use a type check to avoid confusion ?

jiant/models.py
```diff
@@ -998,34 +997,6 @@ def _pair_sentence_forward(self, batch, task, predict):
                 _, out["preds"] = logits.max(dim=1)
         return out
 
-    def _seq_gen_forward(self, batch, task, predict):
```
W4ngatang: what's with the mass deletion ?

jiant/models.py
```diff
@@ -1143,51 +1116,18 @@ def _mc_forward(self, batch, task, predict):
...
-            out["loss"] = F.cross_entropy(logits, labels)
-            task.update_metrics(logits, labels)
+        out["n_exs"] = get_batch_size(batch, self._cuda_device, keyword="choice0")
 
-        if predict:
-            out["preds"] = logits.argmax(dim=-1)
-
-        return out
-
-    def _lm_only_lr_forward(self, batch, task):
```
W4ngatang: what's with the mass deletion ?

jiant/models.py
pruksmhc: Oh I think this was a merge error

jiant/preprocess.py
pruksmhc: You still need it for the build_tasks function

jiant/trainer.py
pruksmhc: agreed

jiant/utils/utils.py
pruksmhc: It could be a scalar for example and not a torch.Tensor. For example, out["loss"] is usually passed into format_output

pruksmhc: Because out["loss"] is already on cuda. We just explicitly put cuda here because we're creating a torch.Tensor.

jiant/utils/utils.py
pruksmhc: Because out["loss"] is already on cuda. We just explicitly put cuda here because we're creating a torch.Tensor.

jiant/utils/utils.py
pruksmhc: yeah not about that long lines life

jiant/utils/utils.py
pruksmhc: It shouldn't only receive torch.Tensor

jiant/utils/options.py
```diff
@@ -0,0 +1,40 @@
...
+        return cuda_list
+    elif isinstance(cuda_list, int):
+        return cuda_list
+    elif "," in cuda_list:
+        return [int(d) for d in cuda_list.split(",")]
+    else:
+        raise ValueError(
+            "Your cuda settings do not match any of the possibilities in defaults.conf"
+        )
+    if len(result_cuda) == 1:
```
W4ngatang: we shouldn't hit this ever , I think

jiant/utils/utils.py
```diff
@@ -361,15 +401,38 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
...
+            obj = torch.tensor(obj).cuda()
+        return obj.unsqueeze(0)
+    else:
+        return obj
+
+
+def uses_cuda(cuda_devices):
+    use_cuda = 1
+    if isinstance(cuda_devices, list) or isinstance(cuda_devices, int) and cuda_devices >= 0:
+        return use_cuda
```
W4ngatang: just do `return 1` (deleting `use_cuda`) , or just `return isinstance(cuda_devices, list) or ...` ?

jiant/utils/utils.py
```diff
@@ -361,15 +401,38 @@ def get_elmo_mixing_weights(text_field_embedder, task=None):
...
+    return 0
+
+
+def get_batch_size(batch, cuda_devices, keyword="input"):
```
W4ngatang: if we're doing distributed training, does it but the same batch size on all devices (so that we'd need to multiply by number of devices here), or is it splitting the batch across devices ?

jiant/utils/utils.py
W4ngatang: I guess my concern is making sure that we're not trying to call `torch.tensor` on something that shouldn't be a tensor ?

jiant/utils/utils.py
W4ngatang: Hrm, I guess I'm still confused as to what this function is doing. My rough understanding is that (1) we're only calling this on the direct outputs of the model (either tensors or lists of tensors?), and if it's the latter (or rather, if we're using Distributed), we want to aggregate that list of tensors into a single tensor. Is that right ? And if so, we can change the name of the function to be more descriptive ?

jiant/utils/utils.py
pruksmhc: It splits up the batch into equal parts to all devices. However, we don't have to multiply because the get_output_attribute sums the n_exes together from all batches.

jiant/utils/options.py
pruksmhc: We will hit this for if someone puts in "auto" and there's only 1 GPU in the device. I'll move this up.

jiant/utils/utils.py
pruksmhc: The only case where we make it a tensor is for nn.DataParallel (when there are multiple gpu devices being used).

jiant/evaluate.py
```diff
@@ -84,10 +85,10 @@ def evaluate(
         generator = iterator(dataset, num_epochs=1, shuffle=False)
         for batch_idx, batch in enumerate(generator):
             with torch.no_grad():
-                batch = move_to_device(batch, cuda_device)
+                if isinstance(cuda_device, int):
+                    batch = move_to_device(batch, cuda_device)
```
W4ngatang: will this fail if `cuda_device = -1` ? I think it might

W4ngatang: nvm, allennlp handles it

W4ngatang:
Cool! Your memory/time changes look like about what I'd expect and have seen. Adding on a few seconds to double batch size seems worth it.

PR #882, opened 2019-08-08
Fix config file name quirks

sleepinyourhat:


sleepinyourhat:
This PR features infinity commits-per-LOC.

PR #884, opened 2019-08-08
Simplify/fix several related issues on the MNLI/AX code path

sleepinyourhat:
A few related issues. They don't have to be one PR, but keeping them together simplifies testing quite a bit.
- `glue-diagnostic` and `mnli` both load and use the GLUE diagnostic set. I removed this from `mnli`, and modified code elsewhere to ensure that setting `target_tasks=glue` still behaves correctly. 
- `MultiNLITask` and `MultiNLISingleGenre` task were largely redundant. I merged these two, and refactored `MultiNLIHypothesisOnlyTask ` to also support a single-genre option.
- `check_configurations` in main was broken in a way that made it difficult to test `glue-diagnostic` and other tasks that rely on `use_classifier`. Fixed.
- All of this put me in contact with `load_tsv`, which is end-user-facing but quite confusing. I started to clean up the docstring, but don't follow enough of what's going on to finish the job. @pruksmhc  - Would you mind filling in the missing pieces and pushing to this branch?


jiant/utils/data_loaders.py
```diff
@@ -97,27 +97,36 @@ def load_tsv(
...
+        data_file (str): The path to the file to read.
+        max_seq_len (int): The maximum number of tokens to keep after tokenization, per text field.
+            Start and end symbols are introduced before tokenization, and are counted, so we will
+            keep max_seq_len - 2 tokens *of text*.
+        label_idx (int|None): The column index for the label field, if present.
+        s1_idx (int): The column index for the first text field.
+        s2_idx (int|None): The column index for the second text field, if present.
+        label_fn (fn: str -> int|None): A function to map items in column label_idx to int-valued
+            labels.
+        col_indices (list[int]): TODO(YADA): ???
```
pruksmhc: Right now, we do not red in the column names in pandas so we rely on indices. Tag_vocab I believe has to do with CCG-related tags where we want to evaluate on a subset of the dataset (@HaokunLiu)

jiant/evaluate.py
```diff
@@ -595,14 +595,9 @@ def _add_default_column(df, name: str, val):
             _get_pred_filename("mnli-m", pred_dir, split_name, strict_glue_format),
         )
         _write_preds_with_pd(
-            preds_df.iloc[9796:19643],
```
pruksmhc: Add a comment here as to why we slice here

jiant/utils/data_loaders.py
```diff
@@ -97,27 +97,36 @@ def load_tsv(
...
+        tag_vocab (TODO(YADA)): Vocab object contains the tags (TODO(YADA): What are these?)
+        tag2idx_dict (dict<string, int>): Map from coarse category name to column index
+            (TODO(Yada): Clarify.)
+
     Returns:
         List of first and second sentences, labels, and if applicable indices
     """
-    # TODO(Yada): Instead of index integers, adjust this to pass ins column names
+
+    # TODO(Yada): [Make this an issue] Instead of index integers, adjust this to pass in column names
```
sleepinyourhat: TODO

jiant/utils/data_loaders.py
HaokunLiu: Ah, sorry, I didn't notice this. There are too many github mails.

HaokunLiu: Tag_vocab is not CCG-related, it's related to cola-analysis etc., where we want to know the results on a subset of the task examples.

sleepinyourhat: Tx. If you know what it's for/what it requires, please push a comment to this branch. :D

HaokunLiu: The tag means, in some dataset, each example may have one or more tags, like mnli-diagnostic and cola-analysis. Tag_vocab is a vocab used to track what tags are there all together in different data splits of that task. tag2idx_dict means, the data tsv, some columns contains information about what tags the example have. And for both mnli-diagnostic and cola-analysis, the tags forms a two-level hierarchy, there are "big tags" beneath which there are "smaller tags", this dictionary is a mapping from the name of "big tags" to index of column contain the information about what "small tags" beneath that "big tag" the examples have.
Initially this mechanism was introduced when working on NPI task, as well as for refactoring cola-analysis and mnli-diagnostic, but the mnli-diagnostic refactoring thing get dropped in a consecutive jiant upgrades, superglue upgrades, etc.

sleepinyourhat: Would you mind updating the branch with this information? I'm not using this code myself now, I'm asking because there really needs to be documentation in the docstring. Thanks!

HaokunLiu: Ok, you mean update mnli-path-cleanup branch?

sleepinyourhat:
Yep! I know it's a slightly unusual way of doing this, but I think it's the
fastest way to finish this...

sleepinyourhat:
Thanks!

jiant/utils/data_loaders.py
sleepinyourhat: @pruksmhc Mind updating the TODO on this line?

jiant/utils/data_loaders.py
```diff
@@ -113,7 +113,8 @@ def load_tsv(
         s2_idx (int|None): The column index for the second text field, if present.
         label_fn (fn: str -> int|None): A function to map items in column label_idx to int-valued
             labels.
-        col_indices (list[int]): TODO(YADA): ???
+        col_indices (list[int]): column names, which should be 1-n, where n is the number of columns
```
sleepinyourhat: @pruksmhc Sorry to keep bugging you about this, but I need more help to figure out what to do here. I still don't understand what this is, or when it's used. Why is a list of indices called 'names', and why is it required when we already give the indices of the text and labels?

pruksmhc: Regardless of the column names, we currently index based on integers (since we specify indices as integers). Thus, we need to pass in col_indices. Another way to actually do this is to change the default value of skip_rows to 0, then we don't need to pass in col_indices. I'll make that change to make it clearer.

sleepinyourhat: Did you mean to push something? I still don't really understand from this comment. We already give column indices in the previous argument. What indices are these? Where are they used?

sleepinyourhat: Thanks!

PR #895, opened 2019-08-21
Make sure git branch/sha information appears even when worknig dir isn't repo dir

sleepinyourhat:
Currently, we only show git branch/sha information in the log if you run from within the git repo's directory. This assumption is often violated in cloud environments.

This fix ensures that the git commands that collect this information are run in the directory that contains main, regardless of the current working directory.


sleepinyourhat:
Thanks to @iftenney for a tip on what the underlying problem here was.

iftenney:
Yup, this looks correct (assuming this does the right thing and doesn't change the directory that the main binary runs from, but that would probably cause other bugs).

PR #898, opened 2019-08-25
Fix config file name and cb task name in comment

njjiang:
* Change the script to use the updated config file name.
* Update comments to reflect the correct way of running the CommitmentBank task.


sleepinyourhat:
Hrm—one of the changes seems to have disappeared...

njjiang:
It looks like this is fixed in a recent PR?

sleepinyourhat:
Thanks!

Issue #890, opened 2019-08-15
RoBERTa support


sleepinyourhat:
It's ready in pytorch_transformers, so we just need to add the interface code. Some discussion in #881.

https://www.google.com/url?q=https://urldefense.proofpoint.com/v2/url?u%3Dhttps-3A__github.com_huggingface_pytorch-2Dtransformers_releases_tag_1.1.0-3Femail-5Fsource-3Dnotifications-26email-5Ftoken-3DAAJZSWP6UCKXJH57B5VQENDQEVZH3A5CNFSM4IL7KGAKYY3PNVWWK3TUL52HS4DFU5JGK3DFMFZWLKTDN5WW2ZLOORPWSZGOAETMASQ%26d%3DDwMCaQ%26c%3DslrrB7dE8n7gBJbeO0g-IQ%26r%3DsCzLyHdE8zgQwk2-sKwA1w%26m%3Dcg2tUcgtNESh6Wn0rVEK5zfKhc5hCf4XfjlttyKszz0%26s%3DUoWZxG8KT3vLXb4vb3g-YyDn0rezu4tLUM9vlJQTA2A%26e%3D&source=gmail&ust=1565969616801000&usg=AFQjCNGzaMvGuBNc8DUth2g_9Cl8oZI8Rw

PR #903, opened 2019-08-27
Update retokenization tool to support RoBERTa; fix WSC

sleepinyourhat:
It looks like the retokenization code wasn't touched in #890, which means that WSC isn't supported with RoBERTa models. This should fix that.


jiant/tasks/qa.py
```diff
@@ -273,10 +273,12 @@ def tokenize_preserve_placeholder(sent):
...
             sent_parts = sent.split("@placeholder")
             assert len(sent_parts) == 2
-            sent_parts = [
-                tokenize_and_truncate(self.tokenizer_name, s, self.max_seq_len) for s in sent_parts
-            ]
-            return sent_parts[0] + ["@placeholder"] + sent_parts[1]
+            placeholder_loc = len(
+                tokenize_and_truncate(self.tokenizer_name, sent_parts[0], self.max_seq_len)
+            )
+            sent_tok = tokenize_and_truncate(self.tokenizer_name, sent, self.max_seq_len)
```
iftenney: This seems a little sketchy - prefer something like `return sent_tok[:i] + ["@placeholder"] + sent_tok[i:]` than mutating with insert?

Also, should this be truncated to `self.max_seq_len - ` to account for the placeholder?

jiant/utils/retokenize.py
```diff
@@ -328,12 +339,48 @@ def align_wpm(text: Text, tokenizer_name: str) -> Tuple[TokenAligner, List[Text]
...
+    sentencepiece_tokens = sentencepiece_tokenizer.tokenize(text)
+
+    modified_sentencepiece_tokens = list(
+        map(process_sentencepiece_for_alignment, sentencepiece_tokens)
+    )
+    ta = TokenAligner(bow_tokens, modified_sentencepiece_tokens)
+    return ta, sentencepiece_tokens
+
+
+def align_bpe(text: Text, tokenizer_name: str) -> Tuple[TokenAligner, List[Text]]:
```
iftenney: Add function docstrings giving examples and stating which models these apply to?

Since these are fairly dense and involve a number of edge cases, should we write some unit tests for this module?

jiant/tasks/qa.py
HaokunLiu: Thanks.

jiant/utils/retokenize.py
HaokunLiu: I am afraid it will not be very economical to do this.
The only occasions we need to modify this, is either adding new tokenizer or refactoring retokenize. When the first happens, the old test cases can always pass, and we always need to add new test cases (which has little reuse value). When the second happens, I don't expect we will keep the same interface.
I wonder what do you think about it?

iftenney: I don't think it'll be hard - this module has a very minimal API so it should be easy to write tests. We are refactoring and adding new tokenizers here, and we'd want tests both to ensure there aren't regressions and that the new functionality is doing what we expect.

(This is partly my fault for not having any on the original version, but since it's grown quite a bit in scope and tokenization bugs can be very hard to detect otherwise, I think it's warranted now.)

HaokunLiu: I see. I'll do it later today.

sleepinyourhat:
What's left to do here? We've been holding off on the RoBERTa release for a while now...

HaokunLiu:
> What's left to do here? We've been holding off on the RoBERTa release for a while now...

I have added docstrings and test cases, the only problem is using bypebpe tokenizer requires download spacy en, I don't know if there is anyway to do it in circleci. If we simply remove the test case for bytebpe, then this PR is ready to go. Otherwise I'm still waiting to see if @davidbenton  can tell me something.

sleepinyourhat:
Ah, that should be easy—I didn't realize you were waiting. CircleCI just runs a list of commands, so you can just add a shell command to download the data here: 

https://github.com/nyu-mll/jiant/blob/master/.circleci/config.yml#L50

If it takes more than a few minutes to download, though, it's okay to skip.

HaokunLiu:
This should be ready now.

sleepinyourhat:
Great! @iftenney, merge when ready.

jiant/utils/retokenize.py
```diff
@@ -294,28 +297,40 @@ def process_wordpiece_for_alignment(t):
...
+    if t.startswith("▁"):
+        return "<w>" + re.sub(r"^▁", "", t)
+    else:
+        return t
+
+
+def process_bytebpe_for_alignment(t):
+    """Add <w> markers to ensure word-boundary alignment."""
+    if t.startswith("▁"):
+        return "<w>" + re.sub(r"^Ġ", "", t)
```
iftenney: Is "^Ġ" an escape sequence for something, or is this the evil-underscore?

PR #908, opened 2019-09-05
Beamsearch 3.0

Kelina:
Beam search added to the seq2seq decoder.
Also, the decoder got different train/valid/test modes (see comment in the class).


jiant/models.py
```diff
@@ -1003,15 +1006,24 @@ def _seq_gen_forward(self, batch, task, predict):
...
             target = batch["targs"]["words"][:, 1:].contiguous()
             target_mask = out["target_mask"]
-            logits = out["logits"]
-            task.update_metrics(logits, target, target_mask[:, 1:].contiguous())
 
+            assert "logits" in out or "predictions" in out
+
+            if "logits" in out:
+                logits = out["logits"]
+                task.update_metrics(logits, target, target_mask[:, 1:].contiguous())
```
W4ngatang: slight preference for using keyword args to make sure the order of args is right, e.g. `(logits=logits, mask=target_mask...)` and `(target=target, ...)`

jiant/modules/beam_search.py
```diff
@@ -0,0 +1,297 @@
...
+    Parameters
+    ----------
+    end_index : ``int``
+        The index of the "stop" or "end" token in the target vocabulary.
+    max_steps : ``int``, optional (default = 50)
+        The maximum number of decoding steps to take, i.e. the maximum length
+        of the predicted sequences.
+    beam_size : ``int``, optional (default = 10)
+        The width of the beam used.
+    per_node_beam_size : ``int``, optional (default = beam_size)
```
W4ngatang: huh, interesting

jiant/modules/seq2seq_decoder.py
```diff
@@ -148,9 +208,16 @@ def forward(
         """
         batch_size, _, _ = encoder_outputs.size()
 
-        use_gold = target_tokens is not None and self.training
+        if self.training:
+            assert target_tokens
+            mode = "train"
+        elif target_tokens is not None:
+            assert target_tokens
```
W4ngatang: why would this assert ever fail given the above condition ?

jiant/modules/seq2seq_decoder.py
```diff
@@ -191,16 +273,16 @@ def forward(
...
-            output_dict["loss"] = loss
-        else:
-            # This is needed in case no gold target sequence is available.
-            output_dict["loss"] = torch.tensor([-1.0])
+        targets = target_tokens["words"]
+        target_mask = get_text_field_mask(target_tokens)
+        output_dict["target_mask"] = target_mask
+        relevant_logits = logits[:, : targets.shape[1] - 1, :].contiguous()
+        loss = self._get_loss(relevant_logits, targets, target_mask)
+        output_dict["loss"] = loss
```
W4ngatang: 1) Is there a way to return loss of gold tokens at test time ?

2) Do we never query `output_dict` for the loss at test time ? We're no longer hard-setting it to zero here.

jiant/tasks/seq2seq.py
```diff
@@ -118,10 +119,22 @@ def get_metrics(self, reset=False):
...
 
-    def update_metrics(self, logits, labels, tagmask=None):
-        # Cut logits, since generation is done until max seq length (usually >> target length).
-        relevant_logits = logits.max(dim=2)[1][:, : labels.shape[1]]
-        self.scorer2(relevant_logits, labels, tagmask)
+    def update_metrics(self, logits, labels, tagmask=None, predictions=None):
+        assert logits is not None or predictions is not None
+
+        if logits is not None:
+            # Cut logits, since generation is done until max seq length (usually >> target length).
```
W4ngatang: there's no way to stop generation if all sequences in the batch generate EOS ?

jiant/tasks/seq2seq.py
```diff
@@ -118,10 +119,22 @@ def get_metrics(self, reset=False):
...
+            relevant_logits = logits.max(dim=2)[1][:, : labels.shape[1]]
+            self.scorer2(relevant_logits, labels, tagmask)
+        else:
```
W4ngatang: quick comment: `# logits = None, predictions is not None`

jiant/trainer.py
```diff
@@ -851,7 +851,9 @@ def _calculate_validation_performance(
             batch_num += 1
             with torch.no_grad():
                 out = self._forward(batch, task=task)
-            loss = out["loss"]
+            if "loss" in out:
```
W4ngatang: when do we not have a loss ?

W4ngatang: I guess in some seq2seq example, but why can't/wouldn't we compute loss ?

jiant/trainer.py
```diff
@@ -862,16 +864,19 @@ def _calculate_validation_performance(
...
                         gold = batch["targs"]["words"][0][1:]
-                        logits = out["logits"]
-                        output = logits.max(2)[1][0]
-                        input_string, gold_string, output_string = task.get_prediction(
-                            voc_src, voc_trg, inputs, gold, output
-                        )
-                        log.info("\tInput:\t%s", input_string)
-                        log.info("\tGold:\t%s", gold_string)
-                        log.info("\tOutput:\t%s", output_string)
+                        # logits = out["logits"]
```
W4ngatang: delete

also, is `batch_num == 1` just so you don't print every batch ?

jiant/modules/seq2seq_decoder.py
```diff
@@ -148,9 +208,16 @@ def forward(
...
+            assert target_tokens
+            mode = "train"
+        elif target_tokens is not None:
+            assert target_tokens
+            mode = "valid"
+        else:
+            mode = "test"
 
-        if use_gold:
+        if mode == "train" or mode == "valid":
```
pruksmhc: Nit: valid -> validation

jiant/tasks/seq2seq.py
Kelina: There is a way: the beam search class currently does that. However, we have no way of knowing beforehand that the generated sequence(s) will not be longer than targets.

jiant/modules/seq2seq_decoder.py
Kelina: True, took this out. Thanks!

jiant/tasks/seq2seq.py
Kelina: Done.

jiant/trainer.py
Kelina: Done. Also: yes, I just wanted one example to see what's going on.

jiant/modules/seq2seq_decoder.py
Kelina: Done.

jiant/models.py
Kelina: Done.

jiant/modules/seq2seq_decoder.py
Kelina: Loss is initialized somewhere else in the code to 0.00; this is returned in case we return the beam_search_output earlier.

jiant/trainer.py
Kelina: Nono, never when getting validation performance; this is from before when I didn't compute a loss. Changed!

jiant/models.py
```diff
@@ -1003,15 +1006,26 @@ def _seq_gen_forward(self, batch, task, predict):
...
+            assert "logits" in out or "predictions" in out
+
+            if "logits" in out:
+                logits = out["logits"]
+                task.update_metrics(
+                    logits=logits, labels=target, tagmask=target_mask[:, 1:].contiguous()
+                )
+            else:
+                task.update_metrics(
+                    None, target, target_mask[:, 1:].contiguous(), out["predictions"]
```
W4ngatang: same here with keyword args :) ?

jiant/modules/seq2seq_decoder.py
```diff
@@ -129,6 +138,48 @@ def _initalize_hidden_context_states(self, encoder_outputs, encoder_outputs_mask
...
+            decoder_input, (state["decoder_hidden"], state["decoder_context"])
+        )
+
+        # output projection
+        proj_input = self._projection_bottleneck(state["decoder_hidden"])
+        # (batch_size, num_classes)
+        output_projections = self._output_projection_layer(proj_input)
+
+        # (batch_size, num_classes)
+        step_logit = output_projections
```
W4ngatang: what's the point of this line ?

jiant/modules/seq2seq_decoder.py
```diff
@@ -161,15 +227,30 @@ def forward(
...
         step_logits = []
 
         for timestep in range(num_decoding_steps):
-            if use_gold:
-                input_choices = targets[:, timestep]
-            elif timestep == 0:
-                input_choices = torch.ones((batch_size,)) * self._start_index
-            else:
-                input_choices = torch.max(step_logits[-1], dim=2)[1].squeeze(1)
+            input_choices = targets[:, timestep]
```
W4ngatang: what happened to the `timestep == 0` case ?

jiant/modules/seq2seq_decoder.py
```diff
@@ -138,7 +189,16 @@ def forward(
...
         # pylint: disable=arguments-differ
         """
-        Decoder logic for producing the entire target sequence at train time.
+        Forward pass for the decoder.
+
+        There are 3 possible modes:
+        1) Training: all metrics are computed given the last gold target tokens at each step
+        2) Validation [no training, but target tokens given]:
+               loss is computed given the gold target tokens;
+               accuracy/BLEU/etc. are computed using predicted tokens, no beam search
```
W4ngatang: isn't beam search being used (I guess with `k=1`, but the overall point being that we are also returning model generations in the validation setting) ?

W4ngatang:
Just to check my understanding, the seq2seq module has three modes: train, validation, test.

In train, we only run teacher forcing (use gold tokens as inputs, throw out model outputs)
In validation, we do both teacher forcing and beam search (greedy) to get a loss + sample generation
In test, we only do beam search

Is that right ?

jiant/modules/seq2seq_decoder.py
```diff
@@ -148,9 +208,15 @@ def forward(
...
+            mode = "train"
+        elif target_tokens is not None:
+            mode = "validation"
+        else:
+            mode = "test"
 
-        if use_gold:
+        if mode == "train" or mode == "validation":
             targets = target_tokens["words"]
             target_sequence_length = targets.size()[1]
```
W4ngatang: just combine these two lines? `target_sequence_length = targets.size(1) - 1`

jiant/modules/seq2seq_decoder.py
```diff
@@ -148,9 +208,15 @@ def forward(
...
+            assert target_tokens
+            mode = "train"
+        elif target_tokens is not None:
+            mode = "validation"
+        else:
+            mode = "test"
 
-        if use_gold:
+        if mode == "train" or mode == "validation":
             targets = target_tokens["words"]
```
W4ngatang: I would prefer to just remove this from the `if` condition, and then do something like `target_sequence_length = targets.size(1) - 1 if mode in ["train", "validation"] else self._max_decoding_steps`.

My mild concern is the presence of lots of code blocks that aren't totally obvious what they each do. This would get rid of a block.

jiant/modules/seq2seq_decoder.py
Kelina: timestep == 0 was needed for the case *without* gold targets only. This is handled earlier now.

jiant/modules/seq2seq_decoder.py
Kelina: Yes, beam search is used during validation; but with beam 1

jiant/models.py
Kelina: Done.

jiant/modules/seq2seq_decoder.py
Kelina: Absolutely none, I am not sure why it was there. :)

Kelina:
Alex, this is not totally right. (See comments in the forward method of the seq2seq_decoder.py).
However, I changed this slightly. Here both versions for comparison.

*Before your review*:
        There are 3 possible modes:
        1) Training: all metrics are computed given the last gold target tokens at each step
        2) Validation [no training, but target tokens given]:
               loss is computed given the gold target tokens;
               accuracy/BLEU/etc. are computed using predicted tokens, no beam search
        3) Test [no training, no target tokens given]:
               loss is NOT computed (the returned loss is 0.0);
               accuracy/BLEU/etc. are computed using predicted tokens, beam search is used

*Now*:
        1) Training/validation/test with gold data [if generate is False]:
               loss is computed given the gold target tokens;
               accuracy/BLEU/etc. are computed using predicted tokens, no beam search
        2) Generation [if generate is True]:
               loss is NOT computed (the returned loss is 0.0) if no target is given,
                   otherwise as above
               accuracy/BLEU/etc. are computed using predicted tokens, beam search is used

jiant/modules/seq2seq_decoder.py
Kelina: Done together with your comment below.

jiant/modules/seq2seq_decoder.py
Kelina: Done.

Issue #907, opened 2019-09-04
Update to pytorch_transformers 1.2.0


sleepinyourhat:
This will add support for the larger GPT-2 model, but may require some other depedency changes.

https://twitter.com/Thom_Wolf/status/1169227071698165760

Anyone have some bandwidth for this?

PR #881, opened 2019-08-08
[WIP] Complete pytorch transformers interface, deprecate old GPT implement

HaokunLiu:
- Replacing old GPT implement with the one from huggingface pytorch transformers
- Add GPT2, Transformer-XL, XLM to pytorch transformer interface 
- Refactor pytorch transformer interface a little bit to reduce duplicated / outdated code & comments

PR #909, opened 2019-09-05
update dependencies to support gpt2-large

HaokunLiu:
#907


HaokunLiu:
#881 already included gpt2-large, just huggingface haven't release 1.2.0 then. So, now it's done.

sleepinyourhat:
Sounds good. Confirm that you can install this *from scratch* in a fresh environment with both conda and pip to check for conflicts. Once that works, I'll merge.

HaokunLiu:
It works.

PR #921, opened 2019-10-01
replace correct_sent_indexing with non inplace version

HaokunLiu:
Previously, pytorch_transformers_interface uses inplace operation to adjust token index, this can cause problem when sent_enc is not "none".


pruksmhc:
Did you test this on the previous failing case (tagger with sent_enc=rnn)?

HaokunLiu:
Yes I did.

PR #922, opened 2019-10-01
Abductive NLI (aNLI)

zphang:
See: https://leaderboard.allenai.org/anli/submissions/get-started

No test data is provided for the task: instead the expectation is that the model is uploaded for evaluation.


jiant/tasks/tasks.py
```diff
@@ -3023,3 +3023,110 @@ def count_examples(self, splits=["train", "val", "test"]):
         for split in splits:
             st = self.get_split_text(split)
             self.example_counts[split] = len(st)
+
+
+@register_task("anli", rel_path="aNLI")
+class AlphaNLITask(MultipleChoiceTask):
+    """ Task class for Abductive Natural Language Inference. """
```
sleepinyourhat: Add a citation or link.

jiant/tasks/tasks.py
```diff
@@ -3023,3 +3023,110 @@ def count_examples(self, splits=["train", "val", "test"]):
...
+
+        self.train_data_text = _load_split(
+            inputs_file=os.path.join(self.path, "train.jsonl"),
+            labels_file=os.path.join(self.path, "train-labels.lst"),
+        )
+        self.val_data_text = _load_split(
+            inputs_file=os.path.join(self.path, "dev.jsonl"),
+            labels_file=os.path.join(self.path, "dev-labels.lst"),
+        )
+        self.test_data_text = [], [], [], [], []
```
sleepinyourhat: Double check that we don't assume the existence of a test set anywhere. Could this do bad things if we do a standard eval with multiple different target tasks?

Alternative would be to set test = dev, and just add a warning here.

zphang:
* bert-large-cased: 66.2 (dev)
* roberta-large: 83.8 (dev)

Compared to
* bert-large (case unclear): 68.9 (test)
* human: 91.4 (test)

PR #926, opened 2019-10-07
Adding Senteval Tasks

pruksmhc:
This adds tasks from https://arxiv.org/abs/1805.01070.
It also modifies and makes process_single_pair more flexible. 

| RoBERTa finetuned performance (accuracy)  | Previous best (accuracy) |
| ------------- | ------------- |
| 46.4  | 99.3 with autoencoder |  
42.4  | 23.3 on BiLSTM last, 59.6 using seq2tree . |  
42.4 t | 35.6 using BiLSTM last, 59.6 using seq2tree |  
89.0  | 78.2 using BiLSTM |  
87.1  | 62.0 using BiLSTM |  
96.4  | 84.3 using BiLSTM |  
96.0 | 84.7 using BiLSTM|  
96.6 | 82.1 using BiLSTM|  
78.9  | 49.9 using BiLSTM |  
89.1  | 65.1 using BiLSTM|


jiant/tasks/senteval.py
```diff
@@ -0,0 +1,530 @@
+""" 
+Set of tasks from Senteval
```
sleepinyourhat: Searching through that paper, it never describes these probing tasks as being part of SentEval.  Probably okay to use something like `senteval-probing` as a short name/filename, but we shouldn't use just `senteval`, and we should add a note here to clarify that we mean these probing tasks that were later added to the senteval repo.

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+import torch
+
+
+# Fields for instance processing
+from jiant.utils.data_loaders import tokenize_and_truncate
+from jiant.tasks.registry import register_task  # global task registry
+from jiant.tasks.tasks import SingleClassificationTask, process_single_pair_task_split
+
+
+@register_task("senteval-probing-sentence-length", rel_path="sentence_length/")
```
W4ngatang: everywhere: "SentEval"/"senteval" -> "SE" for brevity ?

scripts/preprocess_senteval_probing.py
```diff
@@ -0,0 +1,34 @@
...
+        files_val = file_pd[file_pd[0] == "va"]
+        task_name = file.split(".")[0]
+        files_val.to_csv(os.path.join(args.senteval_path, task_name, "val.tsv"))
+
+        files_test = file_pd[file_pd[0] == "te"]
+        task_name = file.split(".")[0]
+        files_test.to_csv(os.path.join(args.senteval_path, task_name, "test.tsv"))
+    import pdb
+
+    pdb.set_trace()
```
W4ngatang: delete pdb and trace ??

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+
+    def get_sentences(self):
+        return self.sentences
+
+    def load_data(self):
+        """ Load data """
+
+        def load_json(data_file):
+            rows = pd.read_csv(data_file, encoding="ISO-8859-1")
+            rows = rows.sample(frac=1, axis=0).reset_index(drop=True)
```
W4ngatang: is this just to shuffle ?

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+    def get_all_labels(self):
+        return [str(x) for x in list(range(6))]
+
+    def get_sentences(self):
+        return self.sentences
+
+    def load_data(self):
+        """ Load data """
+
+        def load_json(data_file):
```
W4ngatang: is it a json? Or csv?

W4ngatang: Or tsv ?

Whichever, update the name of the function

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+            rows = pd.read_csv(data_file, encoding="ISO-8859-1")
```
W4ngatang: random question: why do we need this encoding and not utf-8? Are there special characters in these tasks ?

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+        sentences = []
+        for split in ["train", "val", "test"]:
+            split_data = getattr(self, "%s_data_text" % split)
+            sentences.extend(split_data[0])
+        self.sentences = sentences
+
+
+@register_task("se-probing-subj-number", rel_path="subj_number")
+class SEProbingSubjNumberTask(SingleClassificationTask):
+    """ Subjective number task """
```
W4ngatang: "subject number task"?

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+        sentences = []
+        for split in ["train", "val", "test"]:
+            split_data = getattr(self, "%s_data_text" % split)
+            sentences.extend(split_data[0])
+        self.sentences = sentences
+
+
+@register_task("se-probing-obj-number", rel_path="obj_number")
+class SEProbingObjNumberTask(SingleClassificationTask):
+    """ Objective number task """
```
W4ngatang: "object number task"?

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+
+        def load_json(data_file):
+            rows = pd.read_csv(data_file, encoding="ISO-8859-1")
+            rows = rows.sample(frac=1, axis=0).reset_index(drop=True)
+            rows["s1"] = rows["2"].apply(
+                lambda x: tokenize_and_truncate(self._tokenizer_name, x, self.max_seq_len)
+            )
+            return rows["s1"].tolist(), [], rows["1"].tolist(), list(range(len(rows)))
+
+        self.train_data_text = load_json(os.path.join(self.path, "train.tsv"))
```
W4ngatang: if these are `*.tsv* should `read_csv` have `delimiter='\t'` ?

jiant/tasks/senteval_probing.py
```diff
@@ -0,0 +1,507 @@
...
+
+    def get_sentences(self):
+        return self.sentences
+
+    def load_data(self):
+        """ Load data """
+
+        def load_json(data_file):
+            rows = pd.read_csv(data_file, encoding="ISO-8859-1")
+            labels = rows["1"].apply(lambda x: str(x.split("\t")[0]))
```
W4ngatang: something seems off... Aren't these already .tsv's ? Why do we need to split on tabs again ?

jiant/tasks/senteval_probing.py
pruksmhc: Yep, this is because the dataset authors did not shuffle the dataset (the data is ordered from shortest to longest sentence length).

jiant/tasks/senteval_probing.py
pruksmhc: There are some datasets where the input is of the form  10\tWho knew who would be there ?, but is otherwise a valid csv (quite strange, but I just got around it with this method).

PR #928, opened 2019-10-08
Fixing bug with restoring checkpoint with two gpus + cleaning CUDA parsing related code

pruksmhc:
There was a stray map_location that doesn't currently work with 2 GPUs. I also did some minor cleanup for the cuda parsing code.


jiant/utils/utils.py
```diff
@@ -418,10 +418,7 @@ def format_output(obj, cuda_devices):
 
 
 def uses_cuda(cuda_devices):
-    use_cuda = 1
-    if isinstance(cuda_devices, list) or isinstance(cuda_devices, int) and cuda_devices >= 0:
-        return use_cuda
-    return 0
+    return isinstance(cuda_devices, list) or isinstance(cuda_devices, int) and cuda_devices >= 0
```
W4ngatang: me being paranoid: add parens to the latter two statements ?

but yaaaay for succinctness

jiant/utils/options.py
```diff
@@ -19,22 +20,25 @@ def parse_task_list_arg(task_list):
...
     else:
         raise ValueError(
             "Your cuda settings do not match any of the possibilities in defaults.conf"
         )
-    if len(result_cuda) == 1:
-        result_cuda = result_cuda[0]
+    if torch.cuda.device_count() == 0:
+        # If there are no cuda devices available, default to -1.
+        log.info("CUDA devices not found, defaulting to cuda=-1")
+        result_cuda = -1
```
W4ngatang: how do we feel about failing / louder warning if user thinks they should be using CUDA but devices aren't found ?

pruksmhc: we can do that.

PR #932, opened 2019-10-14
QAMR + QA-SRL Update

zphang:
* Addition of QAMR task
* Switching out QA-SRL evaluation
* Moved some span metrics over to `allennlp_mods/span_metrics.py`

Notes on QAMR:
* QAMR's data is PTB tokenize, so we first detokenize it using sacremoses. All subsequent processing, including evaluation is based on the detokenized data.
* QAMR has both a given answer (provided by the first annotator, who writes both the question and annotates the answer) as well as two validator answers (from two other annotators). We're currently using only the first-annotator answer, on the filtered data (the filtering is done based on the other two validation annotator feedback, but we're not using their provided answers). 
* Also, the QAMR answers are given as individual tokens, but the paper states that the answer spans are usually but not always contiguous. We're simply taking the min/max to get the full span currently.
* Conversion from PTB token-spans is based on `get_aligner_fn`. We also use this to map back from predicted (e.g. BERT) tokens to the original string for the purposes of SQuAD-style scoring. The downside of relying on `get_aligner_fn` is that the SQuAD-scored predictions can only go up to the granularity of space-tokenized tokens. This is usually not an issue because SQuAD scoring normalizes for spacing and punctuation.

Notes on QA-SRL:
* Similar to QAMR, we move over to full SQuAD-style scoring with space-token spans.

I also wrote some additional functions that are not currently used (e.g. for detokenizing BERT tokens to raw strings, if we do move away from `get_aligner_fn`'s reliance on space-tokenization).


jiant/models.py
```diff
@@ -964,32 +964,59 @@ def _span_prediction_forward(self, batch, task, predict):
...
-            task.update_metrics(
-                logits=logits_dict,
-                labels={"span_start": batch["span_start"], "span_end": batch["span_end"]},
+            pred_span_end_i = min(
+                max(pred_span_end_i, pred_span_start_i + 1),
+                len(batch["space_processed_token_map"][i]) - 1,
+            )
+
+            # space_processed_token_map is a list of tuples
+            #   (space_token, processed_token (e.g. BERT), space_token_index)
```
pruksmhc: period after ).

pruksmhc: Also put periods after sentences below.

jiant/allennlp_mods/span_metrics.py
```diff
@@ -1,47 +1,97 @@
-import numpy as np
-import torch
+import collections
```
pruksmhc: Can you move this to jiant/metrics?

jiant/tasks/qa.py
```diff
@@ -608,3 +574,243 @@ def preprocess_qasrl_datum(cls, datum):
...
+
+
+@register_task("qamr", rel_path="QAMR/")
+class QAMRTask(SpanPredictionTask):
+    """ Question-Answer Meaning Representation (QAMR)
+        https://github.com/uwnlp/qamr
+    """
+
+    def __init__(self, path, max_seq_len, name="qamr", **kw):
+        """ There are 1363 supertags in CCGBank without introduced token. """
```
pruksmhc: This doesn't belong here.

pruksmhc:
What's the status of this?

pruksmhc:
@zphang  is this ready for a review?

PR #942, opened 2019-10-23
CommonsenseQA+hellaswag

HaokunLiu:


jiant/tasks/qa.py
```diff
@@ -608,3 +607,102 @@ def preprocess_qasrl_datum(cls, datum):
                 for verb_idx, verb_entry in datum["verbEntries"].items()
             ],
         }
+
+
+@register_task("commonsenseqa", rel_path="CommonsenseQA/")
+@register_task("commonsenseqa-easy", rel_path="CommonsenseQA/", easy=True)
```
pruksmhc: Why is this called "easy"?

jiant/tasks/qa.py
```diff
@@ -608,3 +607,102 @@ def preprocess_qasrl_datum(cls, datum):
...
+class CommonsenseQATask(MultipleChoiceTask):
+    """ Task class for CommonsenseQA Task.  """
+
+    def __init__(self, path, max_seq_len, name, easy=False, **kw):
+        """ """
```
pruksmhc: Delete

jiant/tasks/qa.py
```diff
@@ -608,3 +607,102 @@ def preprocess_qasrl_datum(cls, datum):
...
+        self.choice_idx2label = ["A", "B", "C", "D", "E"]
+
+    def load_data(self):
+        """ Process the dataset located at path.  """
+
+        def _load_split(data_file):
+            questions, choices, targs, id_str = [], [], [], []
+            data = [json.loads(l) for l in open(data_file, encoding="utf-8")]
+            for example in data:
+                this_question = tokenize_and_truncate(
```
pruksmhc: Please rename these "this" variables :S

jiant/tasks/qa.py
```diff
@@ -608,3 +607,102 @@ def preprocess_qasrl_datum(cls, datum):
...
+            + self.val_data_text[0]
+            + [choice for choices in self.train_data_text[1] for choice in choices]
+            + [choice for choices in self.val_data_text[1] for choice in choices]
+        )
+        log.info("\tFinished loading CommonsenseQA data.")
+
+    def process_split(
+        self, split, indexers, model_preprocessing_interface
+    ) -> Iterable[Type[Instance]]:
+        """ Process split text into a list of AlleNNLP Instances. """
```
pruksmhc: "AllenNLP" typo

jiant/tasks/tasks.py
```diff
@@ -2994,6 +2994,95 @@ def get_metrics(self, reset=False):
...
+            + self.val_data_text[0]
+            + [choice for choices in self.train_data_text[1] for choice in choices]
+            + [choice for choices in self.val_data_text[1] for choice in choices]
+        )
+        log.info("\tFinished loading HellaSwag data.")
+
+    def process_split(
+        self, split, indexers, model_preprocessing_interface
+    ) -> Iterable[Type[Instance]]:
+        """ Process split text into a list of AlleNNLP Instances. """
```
pruksmhc: Another typo

jiant/tasks/qa.py
HaokunLiu: The commensenQA comes with a sanity check setting, that's the easy task.

jiant/tasks/tasks.py
HaokunLiu: Yeah, typo proliferates through copy and paste. Sorry, I'll fix it.

jiant/tasks/qa.py
HaokunLiu: what would be a better name?

pruksmhc: just "question" instead of this_question, and same for the others.

HaokunLiu: I add "this" because multichoice task will have multiple answers. 
If we use "answer" for answers in an instance, and "answers" for all the answers in the dataset, that's kind of misleading.
If we use "answers" for answers in an instance, then we need a different name for all the answers.

jiant/tasks/qa.py
HaokunLiu: done

jiant/tasks/qa.py
HaokunLiu: And if we use names like "choicess", that's opening the doorway to endless typo-caused bugs.

pruksmhc: I'm pretty highly against using definite articles in variable names (since to a code reviewer, what does "the" or "this" exactly entail?

pruksmhc: You could also convert this_choices to something like multiple_choices, and then the rest of the variables you can just not use the "this_" in front of them.

HaokunLiu: better now?

pruksmhc:
Thanks for making the edits!

Issue #934, opened 2019-10-16
ValueError on aligner_fun for RoBERTa tokenizer


zphang:
```python
from jiant.utils.retokenize import get_aligner_fn
aligner_fn = get_aligner_fn("roberta-large")
aligner, processed_sentence_tokens = aligner_fn('ξ word')
aligner.project_span(0, 1)
```

Leads to `ValueError: min() arg is an empty sequence`.

No error if we replace the input text with `'a word'`.

zphang:
Follow-up: Another RoBERTa alignment bug, not related to odd characters (but still pertaining to first tokens?)

We're going to match each space-tokenized length-1 span to the corresponding the corresponding token spans.

Setup code:
```python
from jiant.utils.retokenize import get_aligner_fn
roberta_aligner_fn = get_aligner_fn("roberta-large")
bert_aligner_fn = get_aligner_fn("bert-large-cased")
bad_text = 'Larger than life.'
good_text = 'This was larger than life.'

def print_alignment(aligner_fn, input_text):
    aligner, processed_sentence_tokens = aligner_fn(input_text)
    for i, token in enumerate(input_text.split()):
        print(token, processed_sentence_tokens[slice(*aligner.project_span(i, i+1))])
```

RoBERTa will never catch the starting "L":

```
>>> print_alignment(roberta_aligner_fn, bad_text)
Larger ['arger']
than ['Ġthan']
life. ['Ġlife', '.']
```

Compare to these other good examples:

```
>>> print_alignment(roberta_aligner_fn, good_text)
This ['ĠThis']
was ['Ġwas']
larger ['Ġlarger']
than ['Ġthan']
life. ['Ġlife', '.']
```

```
>>> print_alignment(bert_aligner_fn, good_text)
This ['This']
was ['was']
larger ['larger']
than ['than']
life. ['life', '.']
```

```
>>> print_alignment(bert_aligner_fn, bad_text)
Larger ['Large', '##r']
than ['than']
life. ['life', '.']
```

@phu-pmh @pruksmhc

pruksmhc:
@iftenney

iftenney:
I wasn't involved with implementing RoBERTa tokenization, but maybe @sleepinyourhat or @HaokunLiu knows more?

iftenney:
#903 #890

iftenney:
[...]

PR #953, opened 2019-11-06
Adding tokenizer alignment function

pruksmhc:
https://github.com/nyu-mll/jiant/issues/934
Adding a function that does tokenization alignment for any sequence of tokens. 
Note: This is a temporary fix to get QAMR, QA-SRL, and other tasks that is running into the error in the issue up and running. 
It works for the bad cases that Jason had pointed out
Screenshot 2019-11-06 14 00 27

The newest version of huggingface changed their tokenization to match the original tokenization, in which token-by-token tokenization does not work (see Thomas wolf's comment here https://github.com/huggingface/transformers/issues/1196). We will probably have to get rid of the TokenAligner class and adopt and/or generalize fairseq's alignment_utils functionality to have the tokenizers properly working in another bigger PR.


jiant/utils/retokenize.py
```diff
@@ -95,6 +95,29 @@ def _mat_from_spans_sparse(spans: Sequence[Tuple[int, int]], n_chars: int) -> Ma
     return sparse.csr_matrix((data, (ridxs, cidxs)), shape=(len(spans), n_chars))
 
 
+def create_tokenization_alignment(
+    tokens: Sequence[str], tokenizer_name: str
+) -> Sequence[Tuple[str, str]]:
+    """
+    Builds alignment mapping between space tokenization andtokenization of
```
sleepinyourhat: 'andtokenization'

jiant/utils/retokenize.py
```diff
@@ -95,6 +95,29 @@ def _mat_from_spans_sparse(spans: Sequence[Tuple[int, int]], n_chars: int) -> Ma
     return sparse.csr_matrix((data, (ridxs, cidxs)), shape=(len(spans), n_chars))
 
 
+def create_tokenization_alignment(
+    tokens: Sequence[str], tokenizer_name: str
+) -> Sequence[Tuple[str, str]]:
+    """
+    Builds alignment mapping between space tokenization andtokenization of 
+    choice.
```
sleepinyourhat: Add an example of what this does (example input, example output).

sleepinyourhat: And add a blank line before each header (Parameters, etc.)

PR #955, opened 2019-11-08
CCG update

HaokunLiu:
update ccg with lazy iterator
note: when running ccg in an existing experiment directory, reload_vocab is required.

fix a small bug in tagger forward, and some clean up


jiant/tasks/tasks.py
```diff
@@ -2223,7 +2223,7 @@ class TaggingTask(Task):
     def __init__(self, name, num_tags, **kw):
         super().__init__(name, **kw)
         assert num_tags > 0
-        self.num_tags = num_tags + 2  # add tags for unknown and padding
```
HaokunLiu: since we use "*_tags" as label namespace,  we don't need to +2 now.

jiant/models.py
```diff
@@ -1116,30 +1116,22 @@ def _tagger_forward(self, batch: dict, task: TaggingTask, predict: bool) -> dict
...
-            batch_mask = [mask[i][:seq_len] for i in range(b_size)]
-            batch_mask = torch.stack(batch_mask)
-            keep_idxs = torch.nonzero(batch_mask.view(-1).data).squeeze()
+            batch_mask = batch["mask"][:, :seq_len]
+            keep_idxs = torch.nonzero(batch_mask.contiguous().view(-1).data).squeeze()
             logits = logits.index_select(0, keep_idxs)
             targs = targs.index_select(0, keep_idxs)
-        pad_idx = self.vocab.get_token_index(self.vocab._padding_token)
-        out["loss"] = format_output(
-            F.cross_entropy(logits, targs, ignore_index=pad_idx), self._cuda_device
```
HaokunLiu: this was the bug

sleepinyourhat: Glad we caught it!

PR #959, opened 2019-11-14
Set _unk_id in Roberta module

njjiang:
Currently the `_unk_id` for Roberta is not set correctly, which triggers the assertion error on line 118.


sleepinyourhat:
Good catch—thanks!

PR #960, opened 2019-11-15
Fixing load_target_train_checkpoint with mixing setting

pruksmhc:
In the current version of jiant, load_target_train_checkpoint was not working in the "mix" setting.  This is a fix for that.


pruksmhc:
This won't break ELMo setup because checkpointing operates in the same way for all models (and the tests for loading checkpoints are passing here). The only change is that if we _do_ specify load_target_task_checkpoint, we will be able to load those model weights for both finetuning and frozen.

PR #961, opened 2019-11-17
Auto-determine probing config file

zphang:


zphang:
Are we intending to redo the hyperparam sweeps?

pruksmhc:
The hyperparameter sweeps would only effect the non-edge probing tasks (so cola, senteval, and the funcation words probing) , but yes, for the final runs we'll need to have the config numbers correpsonding to the base_roberta.conf

pruksmhc:
Merging for now and I'll add to hyperparam later

PR #970, opened 2019-11-25
Squad

phu-pmh:
- SQuAD v 2.0
- The performance is still very low. (F1: 0.372)


sleepinyourhat:
Are you actively working on this. If so, great! If not, close?

PR #974, opened 2019-12-04
Adding gradient accumulation

pruksmhc:
This PR is to address the push towards integrating optimizations around fitting larger batch sizes in smaller GPUs.
The introduction to the method of gradient accumulation is beautifully explained here: 
https://medium.com/huggingface/training-larger-batches-practical-tips-on-1-gpu-multi-gpu-distributed-setups-ec88c3e51255
Phil and I are currently running experiments to make sure this works, however, the default is that accumulation_gradient = 1 (which is the original behavior of jiant without this PR).

### DIFF
--- a/jiant/__main__.py
+++ b/jiant/__main__.py
@@ -149,6 +149,8 @@ def check_configurations(args, pretrain_tasks, target_tasks):
     ):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
 
+    assert_for_log(args.accumulation_steps >= 1, "accumulation_steps must be a positive int")
+
     if args.load_target_train_checkpoint != "none":
         assert_for_log(
             not args.do_pretrain,
--- a/jiant/__main__.py
+++ b/jiant/__main__.py
@@ -149,7 +149,8 @@ def check_configurations(args, pretrain_tasks, target_tasks):
     ):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
 
-    assert_for_log(args.accumulation_steps >= 1, "accumulation_steps must be a positive int")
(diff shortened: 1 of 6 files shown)


pyeres:
OK, we have some results on BoolQ and ReCoRD:

 **BoolQ** configs:
1. batch size 4, no gradient accumulation (from [superglue_baselines.sh](https://github.com/nyu-mll/jiant/blob/18ca100925c7ccc823c0e3da1d3097a80e1eba10/scripts/superglue-baselines.sh#L14))
2. batch size 2, with gradient accumulation (accumulation_steps=2).
3. batch size 2, no gradient accumulation

Validation performance (F1 and accuracy) for batch size 2 _with gradient accumulation_ was quite close to the baseline config (batch size 4) after adjusting for the number of examples trained on. And it also performed better than the config with batch size 2 _and no gradient accumulation_.

**ReCoRD** configs:
1. batch size 8, no gradient accumulation
2. batch size 4, with gradient accumulation (accumulation_steps=2).
3. batch size 4, no gradient accumulation

Performance with batch size 4 and gradient accumulation_steps = 2 was very similar to performance with batch size 8 and no gradient accumulation (after adjusting for the number of examples trained on). However, on this task there was a small performance disadvantage to using gradient accumulation when we compared batch_size = 4 and accumulation_steps = 2 vs. batch_size = 4 and no gradient accumulation.

So although the performance effects were mixed on **ReCoRD**, it's helpful to add gradient accumulation as a training option on **BoolQ**, and generally we think it'll be helpful wherever we want to get larger effective batch sizes.

pruksmhc:
After meeting with Phil, closing this PR so that Phil can open up another one with the latest implementation of the accumulation.

PR #979, opened 2019-12-19
Updated MT branch

Kelina:
Introducing a recurrent sequence-to-sequence model for MT.

### DIFF
--- a/jiant/config/defaults.conf
+++ b/jiant/config/defaults.conf
@@ -309,7 +309,7 @@ force_include_wsj_vocabulary = 0  // Set if using PTB parsing (grammar induction
 
 // Sentence Encoders on top of Input/Word Encoders //
 
-sent_enc = rnn  // The type of sentence encoder we should use. This is part of the core jiant model,
+sent_enc = ''  // The type of sentence encoder we should use. This is part of the core jiant model,
                 // while encoders like ELMo, GPT, and BERT are input handlers, so it's safe to set
                 // this to 'none' when using ELMo and standard/expected to set this to 'none' when
                 // using GPT or BERT.
--- a/jiant/config/defaults.conf
+++ b/jiant/config/defaults.conf
@@ -446,7 +446,7 @@ s2s {
     attention = "Bahdanau"
     output_proj_input_dim = 100
     beam_size = 10
-}
+  }
 
 // rte = ${glue-small-tasks-tmpl-1}
 // wnli = ${glue-small-tasks-tmpl-1}
--- a/jiant/config/wmt14_en_de.conf
+++ b/jiant/config/wmt14_en_de.conf
@@ -18,7 +18,7 @@ reload_vocab = 0
 do_pretrain = 1
 do_target_task_training = 0
 do_full_eval = 1
-
+sent_enc="rnn"
 pretrain_tasks = "wmt14_en_de"
 target_tasks = ""
 max_seq_len = 30
--- a/jiant/config/wmt14_en_de.conf
+++ b/jiant/config/wmt14_en_de.conf
@@ -1,4 +1,9 @@
-// Template for translation as a seq2seq problem.
+// Template for translation as a seq2seq problem with a recurrent model.
+// Hyperparameters follow very roughly early MT approaches such as Luong et al. (2015), cf. https://arxiv.org/pdf/1508.04025.pdf,
+// but we recommend tuning them for your specific problem.
+
+// This has been verified to run with RoBERTa input by substituting the respective options.
+// However, using RoBERTa for MT with jiant hasn't been tested properly, yet.
 
 // This imports the defaults, which can be overridden below.
 include "defaults.conf"  // relative path to this file
@@ -16,24 +21,23 @@ do_full_eval = 1
 
 pretrain_tasks = "wmt14_en_de"
 target_tasks = ""
-max_seq_len = 64
+max_seq_len = 30
 max_word_v_size = 40000
 max_targ_word_v_size = 40000
 pair_attn = 0
 
-input_module = "roberta-large"
-sent_enc = ""
 d_word = 1000
 d_hid = 1000
 n_layers_enc = 1
 d_proj = 1000
 d_hid_attn = 1000
 skip_embs = 0
-batch_size = 8
+batch_size = 32
 
 optimizer = adam
 lr = 0.001
-//tokenizer = "MosesTokenizer"
+tokenizer = "MosesTokenizer"
+input_module = "scratch"
 val_interval = 1000
 lr_patience = 100
 patience = 200
--- a/jiant/preprocess.py
+++ b/jiant/preprocess.py
@@ -259,8 +259,7 @@ def _build_vocab(args: config.Params, tasks: List[Task], vocab_path: str):
     word2freq, char2freq = get_words(tasks)
     vocab = get_vocab(word2freq, char2freq, max_v_sizes)
     for task in tasks:  # add custom label namespaces
-        # TODO: surface more docs for add_task_label_vocab:
-        add_task_label_vocab(vocab, task)
+        add_task_label_vocab(vocab, task, args.max_targ_word_v_size)
     if args.force_include_wsj_vocabulary:
         # Add WSJ full vocabulary for PTB F1 parsing tasks.
         add_wsj_vocab(vocab, args.data_dir)
@@ -637,7 +636,7 @@ def get_vocab(
     return vocab
 
 
-def add_task_label_vocab(vocab, task):
+def add_task_label_vocab(vocab, task, max_targ_v_size):
     """Add custom task labels to a separate namespace.
 
     If task has a 'get_all_labels' method, call that to get a list of labels
@@ -665,6 +664,7 @@ def add_task_label_vocab(vocab, task):
     log.info("\tTask '%s': adding vocab namespace '%s'", task.name, namespace)
 
     if isinstance(task, SequenceGenerationTask):
+        task.max_targ_v_size = max_targ_v_size
         for special in SPECIALS:
             vocab.add_token_to_namespace(special, namespace)
 
--- a/jiant/tasks/__init__.py
+++ b/jiant/tasks/__init__.py
@@ -99,7 +99,7 @@ ALL_COLA_NPI_TASKS = [
 ]
 
 # Seq2seq tasks
-ALL_SEQ2SEQ_TASKS = ["seg_wix"]
+ALL_SEQ2SEQ_TASKS = ["seg_wix", "wmt14_en_de"]
 
 # people are mostly using nli-prob for now, but we will change to
 # using individual tasks later, so better to have as a list
--- a/jiant/tasks/seq2seq.py
+++ b/jiant/tasks/seq2seq.py
@@ -127,7 +127,7 @@ class Seq2SeqTask(SequenceGenerationTask):
             tagmask = tagmask[:, : predictions.shape[1]]
         self.scorer2(predictions, labels, tagmask)
         return
-        
+
     def get_prediction(self, voc_src, voc_trg, inputs, gold, output):
         return self._get_char_prediction(voc_src, voc_trg, inputs, gold, output)
 
@@ -160,16 +160,17 @@ class Seq2SeqTask(SequenceGenerationTask):
 
         return input_string, gold_string, output_string
 
+
 @register_task("wmt14_en_de", rel_path="wmt14/en_de/", max_targ_v_size=40000)
 class MTTask(Seq2SeqTask):
     def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
         super().__init__(name, max_seq_len, max_targ_v_size, name, **kw)
         self.tokenizer = get_tokenizer(self._tokenizer_name)
         exclusion_index_set = [
-            self.tokenizer.pad_token,
-            self.tokenizer.bos_token,
-            self.tokenizer.eos_token,
-            self.tokenizer.unk_token,
+            self.tokenizer.convert_tokens_to_ids(self.tokenizer.pad_token),
+            self.tokenizer.convert_tokens_to_ids(self.tokenizer.bos_token),
+            self.tokenizer.convert_tokens_to_ids(self.tokenizer.eos_token),
+            self.tokenizer.convert_tokens_to_ids(self.tokenizer.unk_token),
         ]
         self.scorer2 = BLEU(exclude_indices=set(exclusion_index_set))
         self.scorers.append(self.scorer2)
@@ -198,7 +199,7 @@ class MTTask(Seq2SeqTask):
         """Get metrics specific to the task"""
         avg_nll = self.scorer1.get_metric(reset)
         val_metric = self.scorer2.get_metric(reset)
-        return {"perplexity": math.exp(avg_nll), "blue": val_metric["BLEU"]}
+        return {"perplexity": math.exp(avg_nll), "bleu": val_metric["BLEU"]}
 
     def update_metrics(self, logits, labels, tagmask=None, predictions=None):
         # This doesn't require logits for now, since loss is updated in another part.
@@ -215,4 +216,3 @@ class MTTask(Seq2SeqTask):
 
     def get_prediction(self, voc_src, voc_trg, inputs, gold, output):
         return self._get_mt_prediction(voc_src, voc_trg, inputs, gold, output)
-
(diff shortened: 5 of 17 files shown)


sleepinyourhat:
Thanks! What results/sanity checks do you have so far?

sleepinyourhat:
And more generally, what's the state of MT/seq2seq support right now? What's verified to work well, what should work, and are there any possible configurations to avoid?

Kelina:
> Thanks! What results/sanity checks do you have so far?

I trained one model with similar hyperparameters to [https://arxiv.org/pdf/1508.04025.pdf](https://arxiv.org/pdf/1508.04025.pdf). The new model obtained 21.12 BLEU on the validation set, so it's roughly comparable.

Kelina:
> And more generally, what's the state of MT/seq2seq support right now? What's verified to work well, what should work, and are there any possible configurations to avoid?

For now, morphological segmentation as in the respective config file works. MT works as in the respective config file as well; no other configurations have been tried.

sleepinyourhat:
Could you say more about the exact comparison you're making for WMT. Is it test vs. test? What's the target number you're trying to match?

jiant/config/wmt14_en_de.conf
```diff
@@ -0,0 +1,51 @@
+// Template for translation as a seq2seq problem.
```
sleepinyourhat: Say a bit more about where these values come from. Are they supposed to match a paper? Which ones are necessary parts of the setup, and which are just things that you tuned?

sleepinyourhat: Mention that this is a non-Transformers setup, and say what it'd take to use a Transformers model.

jiant/tasks/seq2seq.py
```diff
@@ -22,14 +23,18 @@
...
 
 @register_task("seg_wix", rel_path="seg/wix/", max_targ_v_size=200)
+@register_task("wmt14_en_de", rel_path="wmt14/en_de/", max_targ_v_size=40000)
 class Seq2SeqTask(SequenceGenerationTask):
     """Sequence-to-sequence Task"""
 
     def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
         super().__init__(name, **kw)
-        self.scorer2 = BooleanAccuracy()
+        if name == "seg_wix":
```
sleepinyourhat: Cleaner to make this a keyword argument (like max_targ_v_size) rather than a switch on name.

jiant/tasks/seq2seq.py
```diff
@@ -71,7 +72,10 @@ def get_data_iter(self, path):
                 if len(row) < 2 or not row[0] or not row[1]:
                     continue
                 src_sent = tokenize_and_truncate(self._tokenizer_name, row[0], self.max_seq_len)
-                tgt_sent = tokenize_and_truncate(self._tokenizer_name, row[2], self.max_seq_len)
+                if self.name == "seg_wix":
```
sleepinyourhat: Use a specific property of the task object, rather than the name. (It's fairly common to override name if you want to do MTL on multiple datasets for the same task.)

sleepinyourhat: Same for similar if statements below.

jiant/tasks/seq2seq.py
```diff
@@ -128,10 +137,33 @@ def update_metrics(self, logits, labels, tagmask=None, predictions=None):
             # Cut labels if predictions (without gold target) are shorter.
             labels = labels[:, : predictions.shape[1]]
             tagmask = tagmask[:, : predictions.shape[1]]
```
sleepinyourhat: Are we now creating this even in cases where we won't use it?

jiant/trainer.py
```diff
@@ -860,7 +860,8 @@ def _calculate_validation_performance(
 
             all_val_metrics["%s_loss" % task.name] += loss.data.cpu().numpy()
             n_examples += get_output_attribute(out, "n_exs", self._cuda_device)
-            if print_output:
+            # The following doesn't work for other inputs, yet.
+            if print_output and "inputs" in batch and "words" in batch["inputs"]:
```
sleepinyourhat: This seems like a place where an assert would help: Right now, won't this cause jiant to just ignore a config option?

scripts/wmt14/prepare_MT_data.py
```diff
@@ -0,0 +1,6 @@
+import sys
```
sleepinyourhat: Use lowercase for filenames, as elsewhere.

scripts/wmt14/README
```diff
@@ -0,0 +1,20 @@
+For machine translation (MT) tasks, all examples should be in the format
```
sleepinyourhat: Say where the example data comes from.

Kelina: Done.

scripts/wmt14/prepare_MT_data.py
Kelina: Done.

jiant/trainer.py
Kelina: I made it a NotImplementedError. Does that seem okay?

jiant/config/wmt14_en_de.conf
Kelina: Done.

jiant/tasks/seq2seq.py
Kelina: Done.

jiant/tasks/seq2seq.py
Kelina: Done.

jiant/tasks/seq2seq.py
Kelina: Changed. Thanks!

jiant/trainer.py
sleepinyourhat: Yep!

jiant/tasks/__init__.py
```diff
@@ -99,7 +99,7 @@
 ]
 
 # Seq2seq tasks
-ALL_SEQ2SEQ_TASKS = ["seg_wix"]
+ALL_SEQ2SEQ_TASKS = ["seg_wix", "wmt14_en_de"]
```
sleepinyourhat: Delete. (See above.)

jiant/tasks/seq2seq.py
```diff
@@ -21,15 +22,19 @@
...
+@register_task("seg_wix", rel_path="seg/wix/", max_targ_v_size=200, valid_metric="accuracy")
+@register_task("wmt14_en_de", rel_path="wmt14/en_de/", max_targ_v_size=40000, valid_metric="BLEU")
 class Seq2SeqTask(SequenceGenerationTask):
     """Sequence-to-sequence Task"""
 
-    def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
+    def __init__(self, path, max_seq_len, max_targ_v_size, name, valid_metric, **kw):
         super().__init__(name, **kw)
-        self.scorer2 = BooleanAccuracy()
+        if valid_metric == "accuracy":
```
sleepinyourhat: Make it clear in the header comment for this task class what the options are for this argument. (And generally spell out what all the arguments are.)

sleepinyourhat: (And what this task class is and isn't meant to support.)

jiant/tasks/seq2seq.py
```diff
@@ -21,15 +22,19 @@
...
+            self.scorer2 = BooleanAccuracy()
+            self.val_metric = "%s_accuracy" % self.name
+        else:  # for MT
+            self.scorer2 = BLEU(exclude_indices=set([0, 1, 2, 3]))  # PAD, SOS, EOS, UNK
```
sleepinyourhat: How/where do we verify that this index order is right? Won't this break if we use something like RoBERTa (which we'll have to for our paper) that has its own vocab?

jiant/tasks/seq2seq.py
```diff
@@ -71,7 +72,10 @@ def get_data_iter(self, path):
                 if len(row) < 2 or not row[0] or not row[1]:
                     continue
                 src_sent = tokenize_and_truncate(self._tokenizer_name, row[0], self.max_seq_len)
-                tgt_sent = tokenize_and_truncate(self._tokenizer_name, row[2], self.max_seq_len)
+                if self.val_metric == "%s_accuracy" % self.name:
+                    tgt_sent = tokenize_and_truncate(self._tokenizer_name, row[2], self.max_seq_len)
+                else:  # for MT
```
sleepinyourhat: This looks dangerous, especially since val_metric is user-settable. Find another way to distinguish the two task types.

jiant/tasks/seq2seq.py
```diff
@@ -114,24 +118,52 @@ def _make_instance(input_, target):
...
+            val_metric = val_metric["BLEU"]
+        return {"perplexity": math.exp(avg_nll), metric_name: val_metric}
 
     def update_metrics(self, logits, labels, tagmask=None, predictions=None):
         # This doesn't require logits for now, since loss is updated in another part.
         assert logits is None and predictions is not None
 
-        if labels.shape[1] < predictions.shape[2]:
-            predictions = predictions[:, 0, : labels.shape[1]]
+        if self.val_metric != "%s_accuracy" % self.name:  # for MT
```
sleepinyourhat: See above.

jiant/tasks/seq2seq.py
```diff
@@ -114,24 +118,52 @@ def _make_instance(input_, target):
...
+            else:
+                predictions = predictions[:, 0, :]
+                # Cut labels if predictions (without gold target) are shorter.
+                labels = labels[:, : predictions.shape[1]]
+                tagmask = tagmask[:, : predictions.shape[1]]
+            self.scorer2(predictions, labels, tagmask)
         return
 
     def get_prediction(self, voc_src, voc_trg, inputs, gold, output):
+        if self.val_metric == "%s_accuracy" % self.name:
```
sleepinyourhat: See above.

jiant/config/defaults.conf
```diff
@@ -183,6 +183,8 @@ delete_checkpoints_when_done = 0  // If set, delete *all* saved checkpoints when
                                   // tests on any specific trained model. Will not apply if
                                   // keep_all_checkpoints is set.
 
+valid_metric = "BLEU"  // The validation metric used when a task has multiple options.
```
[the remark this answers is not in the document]
Kelina: Yes, I think this should rather be a property of the task object. Thanks! 
Options are "BLEU" (to be changed to lowercase, see below) and "accuracy".

jiant/tasks/seq2seq.py
Kelina: This part of the code for MT needs to receive the EOS index (related to the above comment regarding BLEU).

jiant/tasks/seq2seq.py
Kelina: @HaokunLiu @pruksmhc Could you take a look at this?

jiant/tasks/seq2seq.py
Kelina: @pruksmhc Could you please have a look at this, too? What would be the best way to distinguish between tasks? (I think Sam was against using task names.)

jiant/tasks/seq2seq.py
Kelina: @pruksmhc

jiant/tasks/seq2seq.py
Kelina: @pruksmhc

jiant/preprocess.py
```diff
@@ -417,6 +417,7 @@ def _get_task(name, args, data_path, scratch_path):
             task_kw["probe_path"] = args["nli-prob"].probe_path
         if name in ALL_SEQ2SEQ_TASKS:
```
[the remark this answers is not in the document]
Kelina: Will do - but just a quick note that I copied this from how it was done for other task (groups). Maybe this should be revised as well?

jiant/config/defaults.conf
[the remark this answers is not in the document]
Kelina: Done.

jiant/preprocess.py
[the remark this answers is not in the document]
Kelina: Update: I don't think this can be done here a you suggested, since it's before the task object's creation. So will leave it as is for now.

jiant/tasks/__init__.py
Kelina: Will stay for now (see above).

jiant/tasks/seq2seq.py
Kelina: I added a comment that this comes from the config file. The options are listed there.

sleepinyourhat:
Still work in progress, right?

Kelina:
> Still pending from before:
> 
> * Could you say more about the exact comparison you're making for WMT. Test set vs test set? Dev set vs dev set? What's the exact target number you're trying to match, how close are you, and how does our setup differ?
> * If transformers (like RoBERTa) are supported, give an example config (or at least some instructions), if not, add an assert/error, and open an issue saying what we'd need to do to add support. We can't start our experiments without this.

I compared the development set results to reported (test set) results. We are slightly better, but our hyperparameters also differ slightly. From above: "I trained one model with similar hyperparameters to [https://arxiv.org/pdf/1508.04025.pdf](https://arxiv.org/pdf/1508.04025.pdf). The new model obtained 21.12 BLEU on the validation set, so it's roughly comparable."

The code runs with RoBERTa, but the results in my one-run experiment were worse (12.4 BLEU on dev); I am unsure if that's because of the hyperparameters. I added the config file I used for that.

Kelina:
> Still work in progress, right?

The open remaining points are probably resolved better and faster by either @pruksmhc or @HaokunLiu. I tagged them above in the respective comments.

jiant/preprocess.py
```diff
@@ -417,6 +417,7 @@ def _get_task(name, args, data_path, scratch_path):
...
             task_kw["max_targ_v_size"] = args.max_targ_word_v_size
+            task_kw["valid_metric"] = args.s2s.valid_metric
```
pruksmhc: Why call this variable valid_metric as opposed to metric?

pruksmhc: Ah validation metric. Let's rename to val_metric.

jiant/tasks/seq2seq.py
```diff
@@ -21,15 +22,20 @@
...
 
-    def __init__(self, path, max_seq_len, max_targ_v_size, name, **kw):
+    def __init__(self, path, max_seq_len, max_targ_v_size, name, valid_metric, **kw):
         super().__init__(name, **kw)
-        self.scorer2 = BooleanAccuracy()
+        # Set scorer2 to compute the validation metric determined in the config file (see 's2s.valid_metric').
+        if valid_metric == "accuracy":
+            self.scorer2 = BooleanAccuracy()
+            self.val_metric = "%s_accuracy" % self.name
+        else:  # for MT
```
pruksmhc: Also why don't we set self.val_metric for BLEU case?

jiant/tasks/seq2seq.py
pruksmhc: There might be a way to use the embedder https://github.com/nyu-mll/jiant/blob/9a47eea0f71021d8c714b4a455f6ad28793baa3c/jiant/pytorch_transformers_interface/modules.py and the _sep_id, _cls_id, _pad_id, and _unk_id variables.

jiant/tasks/seq2seq.py
pruksmhc: Honestly, what I would do is create another class that inherits from Seq2SeqTask (MTTask for example), which uses BLEU. Then, the flow for everything in that class would be for BLEU, and avoids too many if-elses in the code.

jiant/tasks/seq2seq.py
pruksmhc: Same comment

jiant/tasks/seq2seq.py
pruksmhc: Same comment

scripts/wmt14/README
```diff
@@ -0,0 +1,20 @@
...
+
```
pruksmhc: Make sure you indicate in the jiant.info documentation that the documentation for MT is in scripts/wmt14/README.md

pruksmhc:
@Kelina  where did you get the data?

pruksmhc:
A number we should aim to _at_least_ have for RoBERTa is  27.14 BLEU (from https://openreview.net/pdf?id=Hyl7ygStwB),which is the number reported for using BERT to initialize the encoder of NMT).

jeswan:

PR #980, opened 2019-12-20
Add gradient accumulation as an option

pyeres:
This PR adds gradient accumulation as a training option. Gradient accumulation allows us to increase the effective batch size (and use a wider range of GPUs for experiments).

We validated these changes on ReCoRD and BoolQ tasks where we confirmed that we can use gradient accumulation and smaller batch sizes to match the performance of larger batch sizes.

### DIFF
--- a/jiant/__main__.py
+++ b/jiant/__main__.py
@@ -150,9 +150,8 @@ def check_configurations(args, pretrain_tasks, target_tasks):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
 
     assert (
-        hasattr(args, "accumulation_steps") and args.accumulation_steps >= 1,
-        "accumulation_steps must be a positive int.",
-    )
+        hasattr(args, "accumulation_steps") and args.accumulation_steps >= 1
+    ), "accumulation_steps must be a positive int."
 
     if args.load_target_train_checkpoint != "none":
         assert_for_log(
--- a/jiant/__main__.py
+++ b/jiant/__main__.py
@@ -1,4 +1,4 @@
-"""Main flow for jiant. 
+"""Main flow for jiant.
 
 To debug this, run with -m ipdb:
 
@@ -148,6 +148,11 @@ def check_configurations(args, pretrain_tasks, target_tasks):
         [not t.val_metric_decreases for t in pretrain_tasks]
     ):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
+
+    assert_for_log(
+        hasattr(args, "accumulation_steps") and args.accumulation_steps >= 1,
+        "accumulation_steps must be a positive int",
+    )
 
     if args.load_target_train_checkpoint != "none":
         assert_for_log(
--- a/jiant/__main__.py
+++ b/jiant/__main__.py
@@ -149,9 +149,9 @@ def check_configurations(args, pretrain_tasks, target_tasks):
     ):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
 
-    assert_for_log(
+    assert (
         hasattr(args, "accumulation_steps") and args.accumulation_steps >= 1,
-        "accumulation_steps must be a positive int",
+        "accumulation_steps must be a positive int.",
     )
 
     if args.load_target_train_checkpoint != "none":
--- a/jiant/config/defaults.conf
+++ b/jiant/config/defaults.conf
@@ -161,12 +161,13 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
                        // by do_full_eval.
 val_interval = 1000  // Number of gradient steps to take between validation checks. Note that
                      // the stopping criteria are only checked at these intervals.
-accumulation_steps = 1 // The number of batches for which to accumulate gradients before taking
-                       // an optimizer step. If accumulation_steps = 2, then two batches/updates
-                       // are processed before taking one optimizer step. Training metrics (e.g.,
-                       // loss) are reported at step intervals, i.e., training with batch_size = n
-                       // and accumulation_steps = 2 approximates the training process with
-                       // double that batch size (batch_size = 2n) and accumulation_steps = 1.
+accumulation_steps = 1  // The number of batches for which to accumulate gradients before taking
+                        // an optimizer step. If accumulation_steps = 2, then two batches/updates
+                        // are processed before averaging the gradients. Training metrics (e.g.,
+                        // loss) are reported at step intervals, i.e., training with batch_size
+                        // = n and accumulation_steps = 2 simulates the training process with
+                        // batch_size = 2n and accumulation_steps = 1, with lower peak memory
+                        // usage and longer step time.
 max_vals = 1000  // Maximum number of validation checks. Will stop once this limit has been
                  // reached. This cannot be disabled, but if you plan to rely on max_epochs or
                  // min_lr instead for stopping training, simply set it to a very high number.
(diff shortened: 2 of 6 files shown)


jiant/trainer.py
```diff
@@ -591,31 +605,29 @@ def clip_function(grad):
...
-                assert_for_log(
-                    "loss" in output_dict, "Model must return a dict containing a 'loss' key"
-                )
-                loss = get_output_attribute(
-                    output_dict, "loss", self._cuda_device
-                )  # optionally scale loss
+                assert_for_log("loss" in output_dict, "Model must return a dict with 'loss' key")
+                loss = get_output_attribute(output_dict, "loss", self._cuda_device)
+                if self._accumulation_steps > 1:
+                    loss = loss / self._accumulation_steps
```
sleepinyourhat: Double check this. I could be wrong, but I thought that the loss functions we use all _sum_ the loss over entries in a batch. In that case, if we're trying to simulate a larger batch size, we should sum over these sub-batches, not average.

jiant/trainer.py
```diff
@@ -591,31 +605,29 @@ def clip_function(grad):
...
                 loss *= scaling_weights[task.name]
                 loss.backward()
                 assert_for_log(not torch.isnan(loss).any(), "NaNs in loss.")
                 tr_loss += loss.data.cpu().numpy()
+            n_batches_since_val += 1
+            total_batches_trained += 1
+            n_step += 1
```
sleepinyourhat: This doesn't seem to match what we'd talked about. I thought you'd said that we increment the step counter every time we run the model, not every time we take a gradient step. Either way is fine, but make sure we're consistent within the PR.

tests/test_gradient_accumulation.py
```diff
@@ -0,0 +1,65 @@
+import tempfile
```
sleepinyourhat: How hard would it be to add a test that verifies that the gradient for a toy model is the same with two different (but effectively equivalent) batch size/accumulation combinations? Given our current testing standards, I'm okay merging without this, but it'd definitely be good to have.

jiant/__main__.py
```diff
@@ -149,6 +149,11 @@ def check_configurations(args, pretrain_tasks, target_tasks):
     ):
         log.warn("\tMixing training tasks with increasing and decreasing val metrics!")
 
+    assert_for_log(
```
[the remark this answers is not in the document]
pyeres: Done in 97620bb

jiant/__main__.py
```diff
@@ -149,6 +149,11 @@ def check_configurations(args, pretrain_tasks, target_tasks):
...
+        hasattr(args, "accumulation_steps") and args.accumulation_steps >= 1,
+        "accumulation_steps must be a positive int",
```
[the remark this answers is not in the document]
pyeres: Done in 97620bb

jiant/config/defaults.conf
```diff
@@ -161,7 +161,8 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
                        // by do_full_eval.
 val_interval = 1000  // Number of gradient steps to take between validation checks. Note that
                      // the stopping criteria are only checked at these intervals.
-
+accumulation_steps = 1 // the number of steps for which to accumulate the gradient before
+                       // an optimizer step. See https://medium.com/p/ec88c3e51255#e519 .
```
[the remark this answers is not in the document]
pyeres: Done in db5381c

jiant/trainer.py
```diff
@@ -579,6 +587,12 @@ def clip_function(grad):
         offset = 0
         all_tr_metrics = {}
         log.info("Beginning training with stopping criteria based on metric: %s", stop_metric)
+        log.info(
```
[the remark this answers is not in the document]
pyeres: Removed in 602daff

jiant/trainer.py
pyeres: With 9995fe5 we count `batches` and `steps` separately (incrementing `steps` following each optimizer step, and `batches` on each update). I think this should resolve the batch/step counting confusion.

jiant/trainer.py
pyeres: The loss functions I tracked down are using the default mean for reduction, so I think this is OK.

jiant/trainer.py
```diff
@@ -336,10 +340,14 @@ def _setup_training(
                 #  example is included or excluded independently using a hashing function.
                 # Fortunately, it doesn't need to be.
                 task_info["n_tr_batches"] = math.ceil(
-                    task.n_train_examples * self._training_data_fraction / batch_size
+                    task.n_train_examples
+                    * self._training_data_fraction
+                    / (batch_size * self._accumulation_steps)
```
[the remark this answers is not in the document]
pyeres: This is made a clearer w/ 9995fe5f3ba6c38d6851d876b4bcfdc2a7fa550a

tests/test_gradient_accumulation.py
pyeres: Right now this would be hard. But I agree that this is important. I'll open an issue on it.

jiant/config/defaults.conf
```diff
@@ -161,7 +161,12 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
                        // by do_full_eval.
 val_interval = 1000  // Number of gradient steps to take between validation checks. Note that
                      // the stopping criteria are only checked at these intervals.
-
+accumulation_steps = 1 // The number of batches for which to accumulate gradients before taking
```
sleepinyourhat: Nit: Two spaces before comment.

jiant/config/defaults.conf
```diff
@@ -161,7 +161,12 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
...
+                       // an optimizer step. If accumulation_steps = 2, then two batches/updates
+                       // are processed before taking one optimizer step. Training metrics (e.g.,
```
sleepinyourhat: ... summing the gradients

jiant/config/defaults.conf
```diff
@@ -161,7 +161,12 @@ val_data_limit = 5000  // Maximum number of examples to be used during mid-train
...
+                       // loss) are reported at step intervals, i.e., training with batch_size = n
+                       // and accumulation_steps = 2 approximates the training process with
```
sleepinyourhat: approximates => simulates ... with a lower peak memory usage and a longer step time

jiant/config/defaults.conf
sleepinyourhat: (or averaging)

tests/test_gradient_accumulation.py
sleepinyourhat: Open when ready.

jiant/trainer.py
```diff
@@ -328,23 +332,27 @@ def _setup_training(
...
+                n_training_examples *= self._training_data_fraction
+            task_info["n_tr_batches"] = math.ceil(n_training_examples / batch_size)
+            task_info["n_tr_steps"] = math.ceil(
+                task_info["n_tr_batches"] / self._accumulation_steps
+            )
 
-            task_info["tr_generator"] = tr_generator
             task_info["loss"] = 0.0
             task_info["total_batches_trained"] = 0
             task_info["n_batches_since_val"] = 0
```
sleepinyourhat: Make sure you've double-checked all existing uses of these two _batches_ variables. I'd expect that most of them should transition over to the new _steps_ variables.

pyeres: Confirmed — looks like they're accounted for.

pyeres:
Thanks for the review, @sleepinyourhat — merging.

PR #981, opened 2019-12-22
Specify version of black for CircleCI

sleepinyourhat:
Currently, our pre-commit hook targets a specific version of the `black` Python style checker/fixer, but our CircleCI test tool automatically downloads the latest version. As of now, the latest version is substantially different from the version we target, so there can be cases where the automatic pre-commit style fix actually causes the code to fail the CircleCI style check. 

See, for example: https://circleci.com/workflow-run/502bba6a-1ffc-4845-83df-58094be11d4a?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification 

This change tells CircleCI to use the same fixed version we use locally for pre-commit. (This seems pretty clearly better than using the latest version in both places, as that could easily introduce confusing transient failures after new versions of `black` come out.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@
         - run:
             name: check style
             command: |
-              pip install --user black
+              pip install --user black==19.3b0
               black . --check --exclude "/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|_build|buck-out|build|dist|jiant/modules/cove|venv)/"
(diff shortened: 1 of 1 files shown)


pyeres:
Not an official reviewer on this PR, but 👍: making CircleCI's version of `black` consistent with the version of `black` installed/run with `pre-commit` ([.pre-commit-config.yaml](https://github.com/nyu-mll/jiant/blob/master/.pre-commit-config.yaml)) is a good idea.

Issue #954, opened 2019-11-06
Updating Tokenization


pruksmhc:
We will most likely need an overhaul of how we're doing tokenization. We're currently relying on the aligner_fn, which depends on TokenAligner to keep an alignment mapping. However, the TokenAligner fails for cases where strings contain non-ASCII characters (see https://github.com/nyu-mll/jiant/issues/934). Thus, we should probably do away with that class. We also have a subset of tokenization for tasks like winograd and CCG that rely on token-by-token tokenization. 
In addition, the newest version of huggingface changed their tokenization to match the original fairseq tokenization scheme, in which token-by-token tokenization does not work (see Thomas Wolf's comment here huggingface/transformers#1196). Thus, in order to do any sort of alignment, we probably should follow how fairseq is doing their alignment for RoBERTa ( https://github.com/pytorch/fairseq/blob/master/fairseq/models/roberta/alignment_utils.py).

PR #982, opened 2019-12-28
fix roberta retokenization error

HaokunLiu:
#934  #903 #904 #954

I'm very sorry about this bug. I thought I have fixed retokenization last time. But actually there was a typo left. I should have found this if I checked the test cases carefully.

This bug will cause the roberta retokenization to be less accurate, possibly negatively affect the performance on tasks that use realign index.


tests/test_retokenize.py
```diff
@@ -337,13 +337,13 @@ def test_bytebpe(self):
         ]
         self.token_index_tgt = [
             [[0], [1], [2], [3], [4, 5], [6], [7]],
-            [[0], [1], [2], [3, 4], [5], [6, 7], [8], [9, 10, 11]],
-            [[0], [1, 2, 3], [4], [5], [6], [7], [8], [9, 10, 11], [12], [13], [14, 15]],
+            [[0], [1], [2], [3, 4], [5, 6], [7], [8], [9, 10, 11]],
+            [[0, 1], [2, 3], [4], [5], [6], [7], [8], [9, 10, 11], [12], [13], [14, 15]],
```
HaokunLiu: I wasn't computing this manually. I just printed out the output and skim over it.
I know this isn't how test cases are suppose to be made.
Me trying to save 5 minutes in the wrong place has wasted everyone else hours, there can't be enough apology for this.

sleepinyourhat:
@pruksmhc: Is there a reason this hasn't been merged?

Issue #984, opened 2020-01-07
Find a home for MaskedMultiHeadSelfAttention


pyeres:
`MaskedMultiHeadSelfAttention` is implemented in [jiant/utils/utils.py](https://github.com/nyu-mll/jiant/blame/master/jiant/utils/utils.py). Suggestion: extract class to [jiant/modules](https://github.com/nyu-mll/jiant/tree/master/jiant/modules).

pyeres:
Addressed by #987

PR #985, opened 2020-01-08
Extract MaskedMultiHeadSelfAttention from /utils to /modules

pyeres:
This is a minor cleanup PR (for #984) — currently the `MaskedMultiHeadSelfAttention` class is located in `jiant/utils/utils.py` (where it's the only class among utility fns). This PR would extract this class to `masked_multi_head_self_attention.py` under `jiant/modules/`.

### DIFF
--- a/jiant/modules/masked_multi_head_self_attention.py
+++ b/jiant/modules/masked_multi_head_self_attention.py
@@ -1,0 +1,183 @@
+import torch
+from torch.autograd import Variable
+from torch.nn import Dropout, Linear, Parameter, init
+
+from allennlp.nn.util import masked_softmax
+from allennlp.common.params import Params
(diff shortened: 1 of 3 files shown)


sleepinyourhat:
This module and MaskedStackedSelfAttentionEncoder (which seems to be the only module that uses it) look likely redundant with Transformers and/or AllenNLP, and a quick search didn't show any cases where the latter is used.

Look around a bit more at the history of these two classes, but I think it'd be safe to delete both of these.

pyeres:
Thanks @sleepinyourhat. Confirmed from git history and talking with @zphang and @HaokunLiu that this looks safe to delete.

Closing this PR (and opening a new one to delete these classes).

PR #988, opened 2020-01-09
Cleanup/readability changes in trainer.py

pyeres:
This PR cleans up a couple items in `jiant/trainer.py`:
1. a6fef61 changes the name of `task_info`'s "loss" field (which really tracks loss _since last val_ because it is reset as a side effect of calling `_validate`) to "loss_since_val".
2. There are a number of places in `SamplingMultiTaskTrainer` where a local var in a method is set to reference a field of the class. These extra variables / points of access can get confusing. 1cc7f6e and 37d6b32 make changes to access the member fields directly.
3. 7f72e96 makes changes to access dict values directly (instead of through extra local vars).

### DIFF
--- a/jiant/trainer.py
+++ b/jiant/trainer.py
@@ -244,6 +244,8 @@ class SamplingMultiTaskTrainer:
         self._training_data_fraction = training_data_fraction
         self._task_infos = None
         self._metric_infos = None
+        self._scheduler = None
+        self._optimizer = None
         self._accumulation_steps = accumulation_steps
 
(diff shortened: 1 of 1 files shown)


pyeres:
@reviewers — these changes are pretty straightforward, but they're more straightforward if you look at them by individual commit.

sleepinyourhat:
I'm procrastinating, so will review now...

pyeres:
Thanks @sleepinyourhat — merging.

PR #989, opened 2020-01-10
Use mean as reduction method for loss with DataParallel training

pyeres:
This change makes loss metrics under different GPU configs more consistent by computing the average loss (instead of sum) across GPUs:
* 4b920c6 adds 'mean' as a reduction type option to `get_output_attribute` (the fn used to reduce scalars gathered in the output dict during DataParallel training). 
* 07a3b15 switches the `loss` calculation under DataParallel training to use 'mean' (instead of 'sum') as its reduction fn.

### DIFF
--- a/jiant/trainer.py
+++ b/jiant/trainer.py
@@ -580,7 +580,7 @@ class SamplingMultiTaskTrainer:
             for batch in itertools.islice(task_info["tr_generator"], self._accumulation_steps):
                 output_dict = self._forward(batch, task=task)
                 assert_for_log("loss" in output_dict, "Model must return a dict with 'loss' key")
-                loss = get_output_attribute(output_dict, "loss", self._cuda_device)
+                loss = get_output_attribute(output_dict, "loss", self._cuda_device, "mean")
                 # Losses are reduced as means. When aggregating loss for accumulation steps,
                 # losses are divided to calculate mean loss over batches within a step.
                 if self._accumulation_steps > 1:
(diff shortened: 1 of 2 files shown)


sleepinyourhat:
Mind pinging others for a review? Otherwise can take a look by Mon.

pyeres:
@pruksmhc and @zphang — you're probably the most familiar with this code. Can you review?

pruksmhc:
Are you sure that without DataParallel we're getting the average loss per batch? Later on in trainer.py we divide the loss by the batch_num as well, so worth double checking.

pyeres:
Thanks for the comments, @pruksmhc — here's why I think we want to use mean:

> Are you sure that without DataParallel we're getting the average loss per batch?

My read is that [pytorch loss fns](https://pytorch.org/docs/stable/nn.html#torch.nn.CrossEntropyLoss) use 'mean' as their default reduction method and I don't see jiant code overriding that. So without DataParallel I'd expect [this loss](https://github.com/nyu-mll/jiant/blob/07a3b15f078f82607dd746e29f4c0b69a16f1063/jiant/trainer.py#L583-L587) to be avg loss per batch.

> Later on in trainer.py we divide the loss by the batch_num as well, so worth double checking.

The division you're referring to might be [this cumulative loss](https://github.com/nyu-mll/jiant/blob/f148485acf438378f430ae1dae92c4d56363a243/jiant/trainer.py#L591) in `task_info` [later divided by steps](https://github.com/nyu-mll/jiant/blob/07a3b15f078f82607dd746e29f4c0b69a16f1063/jiant/trainer.py#L613) (before a6fef61 that cumulative loss in `task_info` was just called `loss` which was confusing).

I also did a quick check that training losses were consistent using `mean` reduction on training runs with different numbers of GPUs (and different settings of `accumulation_steps`).

pyeres:
Thanks for the review, @pruksmhc — merging.

PR #991, opened 2020-01-14
Minor change to maintain sorted order of tasks to reduce variation in logs

pyeres:
Conversion of the [sorted tasks list](https://github.com/nyu-mll/jiant/blob/3ecea5584cd179588314451aaf515dd7922232ad/jiant/__main__.py#L516) to an [unordered set in build_task_modules](https://github.com/nyu-mll/jiant/blob/3ecea5584cd179588314451aaf515dd7922232ad/jiant/models.py#L500) is a source of variation in outputs on runs with the same configs (for example, see logs for "running demo.sh" on builds [1772](https://circleci.com/gh/nyu-mll/jiant/1772?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and [1776](https://circleci.com/gh/nyu-mll/jiant/1776?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)). This change maintains a sorted order for `tasks` making it easier to reproduce runs.

### DIFF
--- a/jiant/models.py
+++ b/jiant/models.py
@@ -1001,8 +1001,8 @@ class MultiTaskModel(nn.Module):
             # The assumption is that each space_token corresponds to multiple processed_tokens.
             # After we get the corresponding start/end space_token_indices, we can do " ".join
             #   to get the corresponding string that is definitely within the original input.
-            # One constraint here is that our predictions can only go up to a the granularity of space_tokens.
-            # This is not so bad because SQuAD-style scripts also remove punctuation.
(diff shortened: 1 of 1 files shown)


sleepinyourhat:
I _think_ this is safe, but try it out with overrides like `pretrain_tasks = \"commitbank,rte\", rte += {use_classifier = commitbank}`. We should make sure that we only build one classifier module, and that we only do target task training once. It seems plausible that we'd accidentally build two different modules whose names both reference commitbank.

Issue #947, opened 2019-11-03
Potential issue with metrics calculation when using multi-GPU training


zphang:
When using multi-GPU training, occasionally I encounter in accurate computation of metrics. For example, training RTE on RoBERTa-large, which should get close to 100% training accuracy after a few epochs, it is possible to see >100% training accuracy.

`11/02 09:02:46 PM: Update 3171: task rte, batch 46 (3171): accuracy: 1.0167, rte_loss: 0.0107`

I suspect the cause of this is that to support multi-GPU, we're calling `DataParallel`, which should be well-behaved for computing the model in parallel across multiple GPUs, but may cause issues because we are also computing/accumulating metrics in our forward step, which may not be multi-processing safe.

sleepinyourhat:
@pruksmhc - Would you mind taking a look at this ASAP or else removing multi-GPU support for now. We're blocked on releases until we're sure multi-GPU is bulletproof.

sleepinyourhat:
If this is really just a small error in validation metric calculation, then it's not a big deal. But this looks like the kind of thing that could very easily be a sign of a deeper problem.

pruksmhc:
I just verified that it indeed is a multi-GPU problem, and that some of the variables in the metrics are being updated at different times (which means depending on which GPU get_metrics() is called on, accuracy can be more than 1.0. 
I'm working on a fix, which will basically be calling task.update_metrics() in _forward as opposed to the various forward functions in jiant/models.py. 
For future reference, for multi-GPU case, any function or operation executed in the scope of the model (self._model) is going to be executed with multi parallelism. However, anything outside that scope (for example trainer scope) is executed on one GPU.

PR #990, opened 2020-01-11
Update to transformers 2.3.0 & Add ALBERT

HaokunLiu:
### Important notice
You will need to change PYTORCH_PRETRAINED_BERT_CACHE to HUGGINGFACE_TRANSFORMERS_CACHE in your own env setting.

### Updates
This PR update from pytorch_transformers 1.0 to transformers 2.3. The major changes are
1. Add albert to input modules
2. Rename mentions of "pytorch_pretrained_bert", "pytorch_transformers" to "transformers". Sometimes the name "huggingface transformers" is used to avoid ambiguity. I only wish they do not transform their name anymore.
3. As #953 pointed out, due to huggingface reverted their bytebpe tokenizer, our temporary fix does not work. This PR changes the alignment function in qa.py back to using TokenAligner.

### Related Issues
#972  #920  #730

### Other
Transformers 2.3 introduced a new feature: AutoModel, AutoTokenizer etc, this may be used to simplify some code. But it is not very necessary right now, so I didn't do it.

PR #992, opened 2020-01-16
Fix Dataparallel Metric Calculation

pruksmhc:
Addressing https://github.com/nyu-mll/jiant/issues/947
The problem is there is a race condition with dataparallelism to do with scorers and variable updating where metrics can exceed 1.0.  To fix this, this PR moves task.update_metrics() in trainer.py as opposed to the various forward functions in jiant/models.py.


sleepinyourhat:
@pyeres Mind taking the lead on this?

pyeres:
Hey @pruksmhc, fyi, running ReCoRD like this:

`python main.py --config jiant/config/superglue_bert.conf --overrides "run_name = rte-batch-8-GPU-2, pretrain_tasks = \"record\", target_tasks = \"record\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, batch_size = 8, val_interval = 1000, val_data_limit = -1"`

It looks like there's an exception related to DataParallel:

```
01/21 07:06:21 PM: Beginning training with stopping criteria based on metric: record_avg
01/21 07:06:30 PM: Fatal error in main():
Traceback (most recent call last):
  File "main.py", line 16, in <module>
    main(sys.argv[1:])
  File "/home/pcy214/projects/jiant/jiant/__main__.py", line 553, in main
    phase="pretrain",
  File "/home/pcy214/projects/jiant/jiant/trainer.py", line 579, in train
    output_dict = self._forward(batch, task=task)
  File "/home/pcy214/projects/jiant/jiant/trainer.py", line 1034, in _forward
    model_out = self._model.forward(task, batch)
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 153, in forward
    return self.gather(outputs, self.output_device)
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 165, in gather
    return gather(outputs, output_device, dim=self.dim)
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 67, in gather
    return gather_map(outputs)
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 61, in gather_map
    for k in out))
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 61, in <genexpr>
    for k in out))
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 62, in gather_map
    return type(out)(map(gather_map, zip(*outputs)))
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 62, in gather_map
    return type(out)(map(gather_map, zip(*outputs)))
  File "/home/pcy214/.conda/envs/jiant/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 62, in gather_map
    return type(out)(map(gather_map, zip(*outputs)))
TypeError: zip argument #1 must support iteration
```

pruksmhc:
Hm, let me debug this, thanks for catching!

pruksmhc:
Alright, I tested it on ReCoRD and STS-B as well as RTE now. @pyeres ready for another look

pyeres:
If I understand these changes correctly, in `model.py` data from the model's output dict (`out`) and data from the model's input dict (`batch`) is stored under special keys with prefix `update_metrics_*` in the model output dict. Then in `trainer.py` and `evaluate.py`, a new dictionary `update_metrics_kwargs` is created to collect the special keys in the model's `out` dict and pass that data as keyword arguments to `task`'s `update_metrics` method. This also involves introducing a task class import into `trainer.py` to handle updating metrics for `ReCoRDTask` which has some special requirements.

Instead of adding special keys to the `out` dict in `models.py` and handling the special case of the ReCoRD task in our `trainer.py`, could you instead just pass the `out` and `batch` dicts directly as arguments to `task`'s `update_metrics` method and handle extracting the required data for the task's `update_metrics` method in the tasks themselves? If you can do that, you can have common update_metrics interface for all tasks (`update_metrics(out, batch)`), and this PR could be one-line changes in `trainer.py` and `evaluate.py`, there would also be no need to write special keys into the output dict in `models.py`.

pyeres:
Separate from the implementation details above, because multi-GPU support touches many code paths (and address a bug), with your final PR can you please show a performance comparison of multi-GPU vs single-GPU on some task(s). For PR #990 @Haokun ran a performance comparison across a a set of common tasks — running these should make for a good comparison (what Haokun already reported in #990 vs. including your changes and using a single-gpu vs. including your changes and using multi-gpu). If you have trouble getting cluster time, just pass me a script and I can run some of these experiments.

pruksmhc:
Makes sense!

pruksmhc:
Hm, for some reason, I'm getting accuracy of 0.56 using 1 GPU in jiant master, but 0.527 with 1 GPU in this branch. Debugging

pruksmhc:
Alright, figured it out. For RTE, it was a matter of increasing batch_size with multi-GPU, I'm able to get 0.56 accuracy. ReCoRD exps are still running.

jiant/models.py
```diff
@@ -1129,7 +1088,6 @@ def _tagger_forward(self, batch: dict, task: TaggingTask, predict: bool) -> dict
         hid2tag = self._get_classifier(task)
         logits = hid2tag(sent[:, 1:-1, :]).view(b_size * seq_len, -1)
         out["logits"] = logits
-        targs = batch["targs"]["words"][:, :seq_len].contiguous().view(-1)
```
pyeres: Looks like there's a code path below that can throw an exception if this is deleted.

jiant/tasks/tasks.py
```diff
@@ -330,6 +334,14 @@ def process_split(
             split, indexers, model_preprocessing_interface, is_pair=False
         )
 
+    def update_metrics(self, out, batch):
+        logits = out["logits"]
+        labels = out["labels"]
+        tagmask = batch.get("tagmask", None)
```
pyeres: Looks like `tagmask` is defined but never used (same in a few more locations in this file). Delete?

jiant/models.py
```diff
@@ -1050,22 +1013,23 @@ def _pair_sentence_forward(self, batch, task, predict):
                 logits = classifier(sent1, sent2, mask1, mask2, [batch["idx1"]], [batch["idx2"]])
             else:
                 logits = classifier(sent1, sent2, mask1, mask2)
-        out["logits"] = logits
+        logits_np = logits
```
pyeres: Is this var a numpy array at this point in the code? If it is, let's rename the `logits` var used above this line to consistently show it's a numpy array, if it isn't let's remove the `*_np`.

jiant/tasks/tasks.py
```diff
@@ -288,10 +288,14 @@ def get_metrics(self, reset: bool = False) -> Dict:
...
         return self.scorers
 
-    def update_metrics(self, logits, labels, tagmask=None):
-        assert len(self.get_scorers()) > 0, "Please specify a score metric"
-        for scorer in self.get_scorers():
-            scorer(logits, labels)
+    def update_metrics(self, out, batch):
+        raise NotImplementedError
+
+    def handle_preds(self, preds, batch):
```
pyeres: Should `handle_preds` be a public method of the generic `Task` class? I think this may be better as a private helper called during `update_metrics` in `SocialIQA` (which I think is the only task using this functionality).

pyeres: Actually, please ignore this comment for now. I'm looking more closely at this.

pruksmhc: Actually I think I'm going to move it to models.py since that's where "preds" is calculated, and this is only a problem for SpanPrediction class.

jiant/preprocess.py
```diff
@@ -552,6 +552,15 @@ def get_tasks(args: config.Params) -> (List[Task], List[str], List[str]):
         # Count examples, store in example_counts.
         if task.example_counts is None:
             task.count_examples()
+        utils.assert_for_log(
+            (
+                task.example_counts["train"] % args.batch_size > len(args.cuda) - 1
```
pyeres: Readability (minor): please use `... >= len(args.cuda)` instead of `... > len(args.cuda) - 1`

tests/test_write_preds.py
```diff
@@ -23,15 +23,15 @@ def model_forward(task, batch, predict=True):
     if task.name == "sts-b":
         logits = torch.Tensor([0.6, 0.4])
         labels = torch.Tensor([0.875, 0.6])
-        out = {"n_exs": 2, "preds": [1.0, 0.8]}
+        out = {"logits": logits, "abels": labels, "n_exs": 2, "preds": [1.0, 0.8]}
```
pyeres: Minor (because these tests are not inspecting fields in `out`): spelling of `labels` key in `out` dict.

tests/tasks/test_tasks.py
```diff
@@ -1,21 +1,59 @@
...
+
+    def load_data(self):
+        return
+
+    def count_examples(self):
+        return 0
+
+
+class TestTasks(unittest.TestCase):
+    def setUp(self):
```
pyeres: The fields initialized in `setUp` aren't necessarily relevant to all tests in `TestTasks`. Can you move the test-specific initializations into the relevant test case?

jiant/trainer.py
```diff
@@ -34,6 +34,7 @@
     get_model_attribute,
     uses_cuda,
 )  # pylint: disable=import-error
+from jiant.tasks.qa import ReCoRDTask
```
pyeres: We don't need to import this task class into trainer.

jiant/utils/utils.py
```diff
@@ -343,7 +343,7 @@ def load_model_state(model, state_path, gpu_id, skip_task_models=[], strict=True
         not (skip_task_models and strict),
         "Can't skip task models while also strictly loading task models. Something is wrong.",
     )
-
+    strict = False
```
pyeres: I'm not familiar with this setting. Can you tell me more about why we're hard-coding this here?

jiant/tasks/tasks.py
```diff
@@ -2380,6 +2428,12 @@ def __init__(self, path, max_seq_len, name, tokenizer_name, **kw):
         self.val_data_text = None
         self.test_data_text = None
 
+    def update_metrics(self, out, batch):
+        logits = out["logits"]
+        labels = batch["targs"]["words"][:, :seq_len].contiguous().view(-1)
```
pyeres: Looks like it'll break with unresolved reference ('seq_len').

pruksmhc: Ah, good catch!

jiant/utils/utils.py
pruksmhc: Oh whoops this was for debugging one of the performance runs. I'll remove this

pruksmhc:
Ready! @pyeres

jiant/tasks/tasks.py
```diff
@@ -2946,6 +3005,66 @@ class SpanPredictionTask(Task):
 
     n_classes = 2
 
+    def update_metrics(self, out, batch):
+        batch_size = sum(out["n_exs"]).item()
```
pyeres: This line breaks `SpanPredictionTask`s (i.e., `QASRL` and `QAMR`) in single-GPU mode:

```
Traceback (most recent call last):
  File "/home/pcy214/jiant/main.py", line 27, in <module>
    raise e  # re-raise exception, in case debugger is attached.
  File "/home/pcy214/jiant/main.py", line 16, in <module>
    main(sys.argv[1:])
  File "/home/pcy214/jiant/jiant/__main__.py", line 638, in main
    phase="target_train",
  File "/home/pcy214/jiant/jiant/trainer.py", line 579, in train
    output_dict = self._forward(batch, task=task)
  File "/home/pcy214/jiant/jiant/trainer.py", line 1035, in _forward
    task.update_metrics(model_out, batch)
  File "/home/pcy214/jiant/jiant/tasks/tasks.py", line 3009, in update_metrics
    batch_size = sum(out["n_exs"]).item()
TypeError: 'int' object is not iterable
```

Looks like a minimal fix could just check the type here and handle `int`s (for single-GPU mode).

pruksmhc: cuda_device isn't passed in here, so I think it might be a bit more involved. Instead, I'll check for batch size from somewhere else in out.

jiant/tasks/tasks.py
```diff
@@ -2946,6 +3005,66 @@ class SpanPredictionTask(Task):
 
     n_classes = 2
 
+    def update_metrics(self, out, batch):
+        batch_size = len(out["logits"]["span_start"])
```
pruksmhc: Decided to do it this way since get_output_attirubte necessitates cuda_device, which we don't pass into here.

jiant/evaluate.py
```diff
@@ -101,11 +101,14 @@ def evaluate(
             with torch.no_grad():
                 if isinstance(cuda_device, int):
                     batch = move_to_device(batch, cuda_device)
-                out = model.forward(task, batch, predict=True)
+                out = model.forward(task=task, batch=batch, predict=True)
+            if task is not None:
+                task.update_metrics(out, batch)
```
pyeres: One more subtle issue: in `/models.py` we missed a call to an [update metrics method](https://github.com/nyu-mll/jiant/blob/41edb47250d37c278f87c58cd53e7a31378ba905/jiant/models.py#L955). It's easy to overlook because it has an unusual name, "`update_diagnostic_metrics`".

pyeres:
Heads up: for CCG it looks like there's an issue with the calculation of training and validation accuracy (accuracy is stuck at zero while training and validation loss is decreasing). 

If it would be helpful I'm happy to debug this further, please let me know.

pyeres:
FYI, with the exception of tasks that have open issues noted in the thread above (diagnostics and CCG), these changes have been successfully benchmarked on tasks from GLUE, SuperGLUE and the benchmarks provided with PR #990 in both single-GPU and multi-GPU mode:

Task | albert-xxlarge-v1 ([#990](https://github.com/nyu-mll/jiant/pull/990#issuecomment-578271369)) | albert-xxlarge-v2 ([#990](https://github.com/nyu-mll/jiant/pull/990#issuecomment-578271369)) | roberta-large ([#990](https://github.com/nyu-mll/jiant/pull/990#issuecomment-578271369)) | roberta-large DataParallel 2-GPUs | roberta-large DataParallel 1GPU
---  |  ---  |  ---  |  ---  |  ---  | ---
commitbank | 0.943 | 0.932 | 1 | 0.946 | 0.982
copa | 0.93 | 0.95 | 0.94 | 0.95 | 0.94
wsc | 0.721 | 0.76 | 0.692 | 0.74 | 0.721
rte | 0.895 | 0.888 | 0.866 | 0.863 | 0.866
multirc | 0.658 | 0.666 | 0.628 | 0.633 | 0.628
wic | 0.752 | 0.74 | 0.724 | 0.724 | 0.74
boolq | 0.886 | 0.89 | 0.879 | 0.873 | 0.879
commonsenseqa | 0.762 | 0.771 | 0.735 | 0.737 | 0.735
cosmosqa | 0.832 | 0.847 | 0.814 | 0.812 | 0.814
record | 0.843 | 0.858 | 0.863 | 0.859 | 0.863
cola | 0.671 | 0.708 | 0.664 | 0.673 | 0.648
 |  |  |  |  | 
qamr | 0.717 | 0.723 | 0.726 | 0.725 | 0.726
scitail | 0.977 | 0.978 | 0.976 | 0.977 | 0.976
social-iqa | 0.783 | 0.778 | 0.766 | 0.774 | 0.766
ccg | 0.954 | 0.956 | 0.96 | **Pending** | **Pending**
hellaswag | 0.885 | 0.896 | 0.851 | 0.851 | 0.851
qasrl | 0.643 | 0.641 | 0.664 | 0.664 | 0.66
sst | 0.95 | 0.955 | 0.961 | 0.963 | 0.961
qqp | 0.893 | 0.898 | 0.903 | 0.898 | 0.903
mnli | 0.899 | 0.898 | 0.896 | 0.897 | 0.896
 |  |  |  |  | 
sts-b |  |  |  | 0.923 | 0.92
mrpc |  |  |  | 0.913 | 0.911
wnli |  |  |  | 0.676 | 0.775
glue-diagnostic |  |  |  | **Pending** | **Pending**
winogender-diagnostic |  |  |  | **Pending** | **Pending**
broadcoverage-diagnostic |  |  |  | **Pending** | **Pending**
qnli |  |  |  | 0.944 | 0.943

pruksmhc:
Thanks for running the extensive benchmark tests!

PR #998, opened 2020-01-31
Masked Language Modeling

phu-pmh:
- Code for MLM task


sleepinyourhat:
Thanks!
- Make sure all tests pass.
- Run a few sanity check tasks. If it's not clear what makes sense, check in with me/Haokun/Phil.

sleepinyourhat:
Comments from a quick look:

- Mark as WIP if it's not ready to review.
- Make sure to only include relevant changes—I can't actually read the code, as there are 128 files changed right now...
- You mentioned that the MLM ppl. is going down during training. That seems like a big red flag, as we're loading an MLM head that has already trained for weeks. What is MLM ppl. like at the start of training right now?

PR #1003, opened 2020-02-04
Pass config to tasks and drop task-specific logic from task builders

pyeres:
Currently [the task building code in preprocessing.py](https://github.com/nyu-mll/jiant/blob/4666ed55572409e894921f6b082a899c3debed10/jiant/preprocess.py#L474-L480) contains logic to determine if the task being built is a special case that requires extracting task-specific parameters from the run config and, if so, it passes these params (among other keyword args) to the task's constructor as `**kwargs`. 

This PR would move responsibility for extracting task-specific params into the tasks themselves (by passing a copy of the run config directly to task constructors).

Advantages:
- makes it easier for task contributors to access config params from inside their tasks.
- addresses [todo in preprocessing.py: "remove special case, replace with something general"](https://github.com/nyu-mll/jiant/blob/master/jiant/preprocess.py#L476-L477).

### DIFF
--- a/jiant/preprocess.py
+++ b/jiant/preprocess.py
@@ -55,6 +55,7 @@ from jiant.tasks import REGISTRY as TASKS_REGISTRY
 from jiant.tasks.seq2seq import Seq2SeqTask
 from jiant.tasks.tasks import SequenceGenerationTask, Task
 from jiant.utils import config, serialize, utils, options
+from jiant.utils.config import Params
 from jiant.utils.options import parse_task_list_arg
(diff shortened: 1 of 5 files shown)


pyeres:
@HaokunLiu, tagging you for visibility because we discussed something like this earlier.

sleepinyourhat:
@pruksmhc would probably review faster than I can this week/month, but the basic idea sounds good to me, FWIW!

jeswan:

Issue #1000, opened 2020-02-01
wrong error description when both pretrain and target tasks are empty


HaokunLiu:
instead of telling the user there is no task. error arises in preprocess line 345.
since the number of tokenizer names is 0, it will tell the user "Error: mixing tasks with different tokenizers!" " Tokenizations"

pyeres:
Resolved in https://github.com/nyu-mll/jiant/pull/1005.

PR #1005, opened 2020-02-07
Fix "wrong error description when both pretrain and target tasks are empty"

pyeres:
Simple fix for https://github.com/nyu-mll/jiant/issues/1000: "wrong error description when both pretrain and target tasks are empty".

Previously, error read: `Error: mixing tasks with different tokenizers! Tokenizations`.

Error now reads: `Error: Must specify at least one pretraining task`, or `Error: Must specify at least one target task`. 

These changes continue to handle the case where task args are set to string "none", as demonstrated in [these configs](https://github.com/search?q=%22pretrain_tasks+%3D+none%22&type=Code).

### DIFF
--- a/jiant/__main__.py
+++ b/jiant/__main__.py
@@ -181,14 +181,14 @@ def check_configurations(args, pretrain_tasks, target_tasks):
 
     if args.do_pretrain:
         assert_for_log(
-            args.pretrain_tasks != "none",
+            args.pretrain_tasks not in ("none", "", None),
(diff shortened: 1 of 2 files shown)


pyeres:
@HaokunLiu, can you please pull this branch and confirm if it resolves the issue you were seeing?

HaokunLiu:
Yes, now it errors correctly. Thank you, for making the world a better place.

pyeres:
Thanks for the review, @pruksmhc. Merging.

Issue #1002, opened 2020-02-04
max_seq_len Issue


zphang:
There appears to be some issue with our tokenization/truncation that doesn't abide by the max_seq_len.

```
AssertionError: input length exceeds position embedding capacity, reduce max_seq_len
```

Here is a reproducible case:
```
python main.py --config jiant/config/superglue_bert.conf \
    --overrides "random_seed = 1234, cuda = 0, exp_name = \"testing\", run_name = qnli, batch_size = 2, input_module = albert-xxlarge-v2, pretrain_tasks = \"qnli\", target_tasks = \"\", do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1, val_interval = 10000, max_seq_len=512, accumulation_steps=2"
```

W4ngatang:
Yea, this is the issue I was discussing on Slack. The issue is that we
apply `max_seq_len` when loading the data, but since we might be
concatenating inputs later, we can easily be over `max_seq_len`. Right now
I'm hacking the relevant model interfaces so that when we concatenate, we
also apply some `max_seq_len`.

pyeres:
Hey @W4ngatang, in your workaround where are you modifying `model_preprocessing_interface`? (`boundary_token_fn`?)

W4ngatang:
In `apply_boundary_tokens`. I'll try and submit a WIP PR later today

mahsash:
Hello.
I am using your library and I faced the same issue. Has it been solved now?

pyeres:
> More precisely, I am using edge probing tasks. I added a new task but I get this error. Can you help me to solve it?

@mahsash, a fix that should cover your use case was merged into master today (61211f3).

HaokunLiu:
When we do this, remember to add max_seq_len for egde probing tasks as well.

naveenjafer:
Hi @pyeres @HaokunLiu  I seem to be facing this issue for edge probind task, could you please help me with this?

naveenjafer:
The exact error that I receive is

zphang:
*This is an automatically generated comment.*

As we update `jiant` to v2.x, `jiant` v1.x has been migrated to https://github.com/nyu-mll/jiant-v1-legacy. As such, we are closing all issues relating to `jiant` v1.x in this repository.

If this issue is still affecting you in `jiant` v1.x, please follow up at https://github.com/nyu-mll/jiant-v1-legacy/issues/1002.

[...]

PR #1006, opened 2020-02-10
Address max sequence length issue (#1002)

pyeres:
This PR proposes a change to addresses the max_seq_len issue: https://github.com/nyu-mll/jiant/issues/1002. 

These changes were validated on [the example](https://github.com/nyu-mll/jiant/issues/1002#issue-559833991) provided by @zphang. @W4ngatang for visibility.

### DIFF
--- a/jiant/preprocess.py
+++ b/jiant/preprocess.py
@@ -830,6 +830,31 @@ class ModelPreprocessingInterface(object):
                     args = list(args)
                     kwargs["s2"] = s2_truncated
                     return apply_boundary_tokens(*args, **kwargs)
+                elif trunc_strategy == "trunc_both":
+                    s1 = args[0]
+                    s2 = kwargs["s2"]
+                    if trunc_side == "right":
+                        s1_truncated = s1[: -num_excess_tokens // 2]
+                        s2_truncated = s2[: -num_excess_tokens // 2]
+                    elif trunc_side == "left":
+                        s1_truncated = s1[num_excess_tokens // 2 :]
+                        s2_truncated = s2[num_excess_tokens // 2 :]
+                    log.info(
+                        "Before truncation s1 length = "
+                        + str(len(s1))
+                        + " and s2 length = "
+                        + str(len(s2))
+                        + ". After truncation s1 length = "
+                        + str(len(s1_truncated))
+                        + " and s2 length = "
+                        + str(len(s2_truncated))
+                    )
+                    assert len(s1_truncated) > 0, "After truncation, s1 length would be 0."
+                    assert len(s2_truncated) > 0, "After truncation, s2 length would be 0."
+                    args = list(args)
+                    args[0] = s1_truncated
+                    kwargs["s2"] = s2_truncated
+                    return apply_boundary_tokens(*args, **kwargs)
                 elif trunc_strategy == "trunc_s1":
                     s1 = args[0]
                     if trunc_side == "right":
--- a/jiant/preprocess.py
+++ b/jiant/preprocess.py
@@ -831,47 +831,57 @@ class ModelPreprocessingInterface(object):
     def __init__(self, args):
         boundary_token_fn = None
         lm_boundary_token_fn = None
+        max_pos: int = None
 
         if args.input_module.startswith("bert-"):
             from jiant.huggingface_transformers_interface.modules import BertEmbedderModule
 
             boundary_token_fn = BertEmbedderModule.apply_boundary_tokens
+            max_pos = BertTokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("roberta-"):
             from jiant.huggingface_transformers_interface.modules import RobertaEmbedderModule
 
             boundary_token_fn = RobertaEmbedderModule.apply_boundary_tokens
+            max_pos = RobertaTokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("albert-"):
             from jiant.huggingface_transformers_interface.modules import AlbertEmbedderModule
 
             boundary_token_fn = AlbertEmbedderModule.apply_boundary_tokens
+            max_pos = AlbertTokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("xlnet-"):
             from jiant.huggingface_transformers_interface.modules import XLNetEmbedderModule
 
             boundary_token_fn = XLNetEmbedderModule.apply_boundary_tokens
+            max_pos = XLNetTokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("openai-gpt"):
             from jiant.huggingface_transformers_interface.modules import OpenAIGPTEmbedderModule
 
             boundary_token_fn = OpenAIGPTEmbedderModule.apply_boundary_tokens
             lm_boundary_token_fn = OpenAIGPTEmbedderModule.apply_lm_boundary_tokens
+            max_pos = OpenAIGPTTokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("gpt2"):
             from jiant.huggingface_transformers_interface.modules import GPT2EmbedderModule
 
             boundary_token_fn = GPT2EmbedderModule.apply_boundary_tokens
             lm_boundary_token_fn = GPT2EmbedderModule.apply_lm_boundary_tokens
+            max_pos = GPT2Tokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("transfo-xl-"):
             from jiant.huggingface_transformers_interface.modules import TransfoXLEmbedderModule
 
             boundary_token_fn = TransfoXLEmbedderModule.apply_boundary_tokens
             lm_boundary_token_fn = TransfoXLEmbedderModule.apply_lm_boundary_tokens
+            max_pos = TransfoXLTokenizer.max_model_input_sizes.get(args.input_module, None)
         elif args.input_module.startswith("xlm-"):
             from jiant.huggingface_transformers_interface.modules import XLMEmbedderModule
 
             boundary_token_fn = XLMEmbedderModule.apply_boundary_tokens
+            max_pos = XLMTokenizer.max_model_input_sizes.get(args.input_module, None)
         else:
             boundary_token_fn = utils.apply_standard_boundary_tokens
 
-        # TODO: consider the max_tokens from a model-specific config instead of args.max_seq_len?
-        self.boundary_token_fn = self._apply_boundary_tokens(boundary_token_fn, args.max_seq_len)
+        self.max_tokens = min(x for x in [max_pos, args.max_seq_len] if x is not None)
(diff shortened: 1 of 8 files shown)


jiant/preprocess.py
```diff
@@ -785,11 +870,15 @@ def __init__(self, args):
         else:
             boundary_token_fn = utils.apply_standard_boundary_tokens
 
-        self.boundary_token_fn = boundary_token_fn
+        # TODO: consider the max_tokens from a model-specific config instead of args.max_seq_len?
```
W4ngatang: probably should be something like `min(args.max_seq_len, model.max_pos)` (assuming `args.max_seq_len` is set, and defaulting to `model.max_pos` if not)

pyeres: 👍, done w/ 65912cc

pyeres:
Thanks for your notes, @W4ngatang —

> This makes sense to me! Scattered thoughts:
> 
> * AFAICT, there's no way to pass `trunc_strategy` right now

Not directly, but the outer method `_apply_boundary_tokens` returns the inner function `_apply_boundary_tokens_with_trunc_strategy`, exposing it to be called with keyword args.

> * will this break if there's a single sentence task that's really long w/ `trunc_strategy = 'trunc_s2'` as the default? I think so b/c it will try to access `args[1]`?
> * I can imagine some pretty subtle bugs arising based on fields being truncated unintentionally based on how the example fields are ordering (e.g. if I've reversed my question and my paragraph, I may cut off a lot of may question). Do you think there's a clean way to do this with named fields, or just rely on lots of loud warnings in the docs?

1) To simplify things I'm updating all the existing `apply_boundary_tokens` functions to require that `s2` be passed as a keyword arg (this is the existing signature, but the existing signature also allowed passing `s2` as a positional arg which made the wrapper logic prone to bugs).
2) I'm updating all the callers of `apply_boundary_tokens` to explicitly provide `s2` as a keyword.
3) To avoid truncating anything implicitly, these changes will now raise an exception when `apply_boundary_tokens` is called IF the sequence is longer than the model/max_seq_len allows AND the truncation strategy was left unspecified. This will push task-related decisions about how to truncate back to the task.

> * We'll probably need to update the docs to point out to people implementing new tasks how to write this and what the strategy should be set to for different tasks, though I think can just truncate s2.

Agreed, happy to turn to this (and finalize the docstrings) if we're aligned on design.

jiant/preprocess.py
```diff
@@ -743,53 +743,171 @@ class ModelPreprocessingInterface(object):
...
+            seq_w_boundry_tokens = apply_boundary_tokens(*args, **kwargs)
+            # if after calling the apply_boundary_tokens function the number of tokens is greater
+            # than the model's max, a truncation strategy can reduce the number of tokens:
+            num_excess_tokens = len(seq_w_boundry_tokens) - max_tokens
+            if num_excess_tokens > 0:
+                if not (trunc_strategy and trunc_side):
+                    raise ValueError(
+                        "Input(s) will exceed model capacity or max_seq_len. "
+                        + "Adjust settings as necessary, or to automatically truncate "
+                        + "inputs in apply_boundary_tokens, call apply_boundary_tokens"
```
W4ngatang: + space

jiant/preprocess.py
```diff
@@ -743,53 +743,171 @@ class ModelPreprocessingInterface(object):
...
+                    + str(max_tokens)
+                    + ". Truncation strategy "
+                    + trunc_strategy
+                    + " will be applied, removing elements from the "
+                    + trunc_side
+                    + " side of the input."
+                )
+                if trunc_strategy == "trunc_s2":
+                    s2 = kwargs["s2"]
+                    log.warning("Before truncation, s2 length = " + str(len(s2)))
```
W4ngatang: incredibly inconsequential: I'd prefer to just print the before/after length on the same line. Some datasets might have lots of lines to be truncated.

pyeres: Sounds like a good idea to me — done in 09cb503.

pyeres:
At this point the PR implements a mechanism for applying truncation strategies which are specifiable when tasks call the `model_preprocessing_interface`. But actually specifying default strategies for the tasks isn't done yet. Planning to request info on reasonable defaults from the user group.

pyeres:
@W4ngatang, @zphang, @sleepinyourhat — request for comment:

Should tasks' strategies for handling model inputs be specified in the task's code, or in the config?

If in the task's code, the implementation in this PR requires specifying a strategy in any task that need a strategy (and some might not need a strategy because their inputs are small).

If in the config, we need a way to pass that option to the task. One way to do that would be to use the route I suggested in #1003.

What do you think?

sleepinyourhat:
I think that truncation strategy will rarely vary within a task—the only case where that might occur is for a very generic task that is @register-ed to point to very different datasets. I think it will basically never vary across runs in a single experiment.

So, I think it's safe/reasonable to leave it as a hardcoded property of task objects if that's expedient. It'd be nice to expose this as an optional task-specific config override, but I think that'll be useful rarely enough that it's safe to merge in this change without that exposure.

W4ngatang:
I agree with Sam. Leaving the truncation strategy hardcoded in the task class will cover almost all of the cases. Also, we discussed this in person, but I'd advocate for an implementation of an `equal` strategy, where `s1` and `s2` are equally truncated.

I am now realizing there's some weirdness in that we are now truncating in two places: once when we load the raw text data and again when using models that concatenate all inputs. Is it worth restructuring so that we only truncate once (probably not, as this would almost never be significant)?

pyeres:
> I am now realizing there's some weirdness in that we are now truncating in two places: once when we load the raw text data and again when using models that concatenate all inputs. Is it worth restructuring so that we only truncate once (probably not, as this would almost never be significant)?

From a performance/correctness point of view, it doesn't seem like a problem to sometimes truncate more than once. However, I agree that it would be good to reduce unnecessary complexity.

IINM, the first instance of truncation that you're referring to is in the `tokenize_and_truncate` function (`/utils/data_loaders.py`), and this fn has a naive approach to account for boundary tokens (`max_seq_len -= 2`), and at this point in the code which tokenizer to apply is known (so perhaps we can use a less naive method). 

But later on, when the tasks' `process_split` method is called, a good deal of conditional logic is used to decide if/how to concatenate inputs to form examples for the model. If this logic forms examples longer than model limits, there will be an error. So, it seems like it's possible to concat two inputs that were individually under the length limit at data loading time, but put together at the point of `process_split` could exceed the limit.

So, if my understanding of the above is correct, it seems like a good reason to have the truncation happen at/after the model-specific decisions on if/how to concatenate inputs (which is how the PR would arrange it). It's also possible that we can a) move this logic upstream to data loading time, or b) eliminate the truncation from `tokenize_and_truncate` and let it happen when the model interface is in play, but I think these would be more substantial refactors.

pyeres:
> I agree with Sam. Leaving the truncation strategy hardcoded in the task class will cover almost all of the cases.

Commit 4cc77fc adds default truncation strategies for most common tasks.

A request was raised yesterday to add handling for any number of inputs (s_n) to `apply_boundary_tokens`. This will change the signature of the call to `apply_boundary_tokens`, but it shouldn't change the default truncation strategies. So, please take a look at the default strategies specified in 4cc77fc and let me know if they seem reasonable to you.

tests/test_preprocess.py
```diff
@@ -75,3 +81,92 @@ def test_build_vocab(self):
...
+        seq = mpi.boundary_token_fn(
+            ["Apple", "buy", "call"],
+            s2=["Xray", "you", "zoo"],
+            trunc_strategy="trunc_s1",
+            trunc_side="left",
+        )
+        self.assertEqual(len(seq), MAX_SEQ_LEN)
+        self.assertEqual(seq, ["[CLS]", "call", "[SEP]", "Xray", "you", "zoo", "[SEP]"])
+
+    def test_boundary_token_fn_trunc_s1_left_side(self):
```
W4ngatang: duplicate test?

tests/test_preprocess.py
```diff
@@ -75,3 +81,92 @@ def test_build_vocab(self):
...
+        seq = mpi.boundary_token_fn(
+            ["Apple", "buy", "call"],
+            s2=["Xray", "you", "zoo"],
+            trunc_strategy="trunc_both",
+            trunc_side="left",
+        )
+        self.assertEqual(len(seq), MAX_SEQ_LEN)
+        self.assertEqual(seq, ["[CLS]", "buy", "call", "[SEP]", "you", "zoo", "[SEP]"])
+
+    def test_boundary_token_fn_trunc_with_short_sequence_an_no_max_seq_len(self):
```
W4ngatang: nit: "an" -> "and"

tests/test_preprocess.py
W4ngatang: I'm guessing this should left -> right?

jiant/preprocess.py
```diff
@@ -743,53 +743,202 @@ class ModelPreprocessingInterface(object):
...
         elif args.input_module.startswith("xlm-"):
             from jiant.huggingface_transformers_interface.modules import XLMEmbedderModule
 
             boundary_token_fn = XLMEmbedderModule.apply_boundary_tokens
+            max_pos = XLMTokenizer.max_model_input_sizes.get(args.input_module, None)
         else:
             boundary_token_fn = utils.apply_standard_boundary_tokens
 
-        self.boundary_token_fn = boundary_token_fn
+        self.max_tokens = min(x for x in [max_pos, args.max_seq_len] if x is not None)
```
W4ngatang: probably would never happen, but is it possible to get both as `None`? (though I think models always have `max_pos != None`)

W4ngatang:
I'd prefer (b) and not (a), but I agree it'd be a larger refactor beyond this PR. In any case, seems low priority.

jiant/preprocess.py
pyeres: If `args.max_seq_len` isn't set, earlier calls to `tokenize_and_truncate` will raise a TypeError. So we shouldn't encounter a scenario where both are `None` here.

tests/test_preprocess.py
pyeres: Thanks — yeah, should have been `right`. Fixing with next commit.

pyeres:
> LGTM modulo some minor cleanup (see comments). I think the defaults for all tasks is reasonable/correct, though I'm not familiar with the format of Winogrande and Abductive NLI. I think `trunc_s1` and `trunc_right` are both reasonable.

Thanks @W4ngatang. For Abductive NLI I'll be conservative and check w/ author @zphang.

pyeres:
> Thanks @W4ngatang. For Abductive NLI I'll be conservative and check w/ author @zphang.

Connected with @zphang — confirmed that to properly apply truncation strategy to `AlphaNLITask` (`(obs1 + hyp, s2=obs2)`), `obs1` should be truncated. This isn't handled by current logic, but has been raised as a feature request and is captured in #1009.

pyeres:
After discussion with @sleepinyourhat, merging these changes to handle the most common cases. Leaving original issue (#1002) open to motivate followup work described in #1009.

PR #1007, opened 2020-02-11
Updated Kubernetes setup

zphang:


gcp/config/jiant_paths.sh
```diff
@@ -14,6 +14,7 @@ export JIANT_PROJECT_PREFIX="$HOME/exp"
 export ELMO_SRC_DIR="/nfs/jiant/share/elmo"
 # cache for BERT etc. models
 export HUGGINGFACE_TRANSFORMERS_CACHE="/nfs/jiant/share/transformers_cache"
+export PYTORCH_TRANSFORMERS_CACHE="/nfs/jiant/share/transformers_cache"
```
sleepinyourhat: Why not just use one? I think the latter replaces the former.

gcp/kubernetes/README.md
```diff
@@ -1,29 +1,43 @@
...
 
-All the instructions below assume Ubuntu Linux running on a Google Compute Engine (GCE) instance, but should also work on any Ubuntu machine with the Google Cloud SDK installed.
+======
+
+# Prerequisites
+
+### Workstation
 
-# Set-Up Instructions
+All setup instructions assume COS (Container OS) running on a Google Compute Engine (GCE) instance, but should also work on any Ubuntu machine with the Google Cloud SDK installed.
```
sleepinyourhat: Is this about the workstation or the container image? Clarify.

gcp/kubernetes/README.md
```diff
@@ -121,3 +179,22 @@ export JIANT_PATH="/nfs/jiant/home/$USER/jiant"
...
+
+There are also additional options, such as for sending jobs to different GPU types, and deleting jobs (pods in Kubernetes terminology).
+
+======
+
+### Using an existing GKE cluster
+
+If someone has already set up a GKE cluster for jiant, and you have a different set of things to do:
+
+1. Create a lightweight GCP instance and SSH into it. git clone jiant.
```
sleepinyourhat: => install jiant. (In case anyone's working from a release package or pip.)

gcp/kubernetes/README.md
```diff
@@ -1,29 +1,43 @@
-# Kubernetes Configuration
+# jiant on GKE
```
sleepinyourhat: Insane nit (you're allowed to ignore): jiant => `jiant` in all running text.

gcp/kubernetes/templates/run_batch.jsonnet
```diff
@@ -38,6 +38,10 @@ function(job_name, command, project_dir, uid, fsgroup,
           name: "HUGGINGFACE_TRANSFORMERS_CACHE",
           value: jiant_env.transformers_cache_path
         },
+        {
+          name: "PYTORCH_TRANSFORMERS_CACHE",
+          value: jiant_env.transformers_cache_path
+        },
```
sleepinyourhat: Redundant, no?

gcp/set_up_workstation.sh
```diff
@@ -4,32 +4,20 @@
...
 source /etc/profile.d/jiant_paths.sh
 if [ ! -d "${JIANT_PROJECT_PREFIX}" ]; then
   mkdir "${JIANT_PROJECT_PREFIX}"
 fi
 if [ ! -d "${HUGGINGFACE_TRANSFORMERS_CACHE}" ]; then
-  sudo mkdir -m 0777 "${HUGGINGFACE_TRANSFORMERS_CACHE}"
+  sudo mkdir -m 0777 -p "${HUGGINGFACE_TRANSFORMERS_CACHE}"
+fi
+if [ ! -d "${PYTORCH_TRANSFORMERS_CACHE}" ]; then
+  sudo mkdir -m 0777 -p "${PYTORCH_TRANSFORMERS_CACHE}"
```
sleepinyourhat: See above.

sleepinyourhat:
The previous setup obviously didn't work out of the box, so I'm fine merging this in as an improvement. The ideal test for this, though, would be to have someone else on the team set up a new GKE cluster using only these instructions.

gcp/kubernetes/README.md
zphang: Updated.

gcp/kubernetes/README.md
zphang: I don't believe jiant needs to be installed at this point. Really this cloning step is just to get the `mount_nfs.sh` script and related config files.
(Or are you referring to using an installation of jiant as opposed to a git cloned copy?)

gcp/kubernetes/README.md
zphang: Updated.

gcp/kubernetes/templates/run_batch.jsonnet
zphang: See discussion above.

gcp/set_up_workstation.sh
zphang: See discussion above.

zphang:
I walked Yada through the latter part of the set up to ensure that she could connect to a cluster I created. Potentially she can go through the full setup from scratch as well?

gcp/config/jiant_paths.sh
zphang: The latter is used by Transformers, while the former is used by jiant.
(In theory we should supply the former directly to Transformers which means the latter should not be necessary, but I ran into errors until I supplied it, so I'll look into that.)

sleepinyourhat: Dang.

gcp/kubernetes/README.md
sleepinyourhat: Yeah. Maybe change this to something like 'clone or download a copy of jiant'?

PR #1008, opened 2020-02-20
Remove unused method of Task classes: tokenizer_is_supported

pyeres:
Project-wide search turned up no uses of the `tokenizer_is_supported` method.

This is a bite-size PR to remove this unused method from the tasks.

Possible review strategy: verify that these methods aren't called.


sleepinyourhat:
It looks like this was added with a pretty specific use case in mind here: https://github.com/nyu-mll/jiant/pull/456/files 

If tests are passing and the method name doesn't appear anywhere in `jiant`, I'm comfortable merging.

PR #1016, opened 2020-02-28
adversarial_nli -> adversarial-nli, SocialIQA -> socialiqa

HaokunLiu:


pyeres:
Thanks @HaokunLiu — are there reasons not to include `wiki103_classif` and `seg_wix`?

HaokunLiu:
Oops. I forgot to change that, sorry. Could you update those names? I’m on
the train.

pyeres:
> Oops. I forgot to change that, sorry. Could you update those names? I’m on the train.

Sure, `seg-wix` and `wiki103-classif` are addressed by ad1dfb7.

HaokunLiu:
> When I searched for `seg_wix` (pre- [ad1dfb7](https://github.com/nyu-mll/jiant/commit/ad1dfb7a473c7c6a962ddc029fccdb0acac872ba)), I noticed that that name string appears outside the `tasks.py` file (in a config and var). Can you confirm that the names you're changing with this PR have been replaced project-wide (beyond the task decorators)?

Good catch. Yes, there are other places. I have updated those ones.

PR #1030, opened 2020-03-08
Adding Masked Language Modelling

pruksmhc:
Adding Masked Language Modeling Task for RoBERTa and ALBERT
What this version of MLM supports: RoBERTa and ALBERT embedders.
Additionally, we fix the get pretrained_lm head function for Transformer-based embedders

**Performance Tests:**
We tested this on CCG and QQP, on both making sure that RoBERTa MLM + QQP -> target task and RoBERTa MLM + CCG -> target were getting reasonable numbers (meaning that MLM perplexity was decreasing and QQP/CCG performance should be close to RoBERTa without MLM training. 

| Task      | Task performance (best RoBERTa single finetuned on task)  | Task performance (multitask trained with MLM with RoBERTa)    | Perplexity after training
| ------------- |:-------------:| :-------------:| :-------------:| 
| CCG   | 0.96 |  0.953 | 2.90563|
| QQP     | 0.898     | 0.853 | 2.45982|

| Target task | Task Performance | CCG w/o MLM performance 
| ------------- |:-------------:| :-------------:|
WIC | 0.760 | 0.716| 
COPA | 0.860 | 0.55| 
CB | 0.820 |0.791 |
RTE | 0.834 | 0.794 | 
CSenseQA | 0.739 | |
BoolQ | 0.837 | 0.84 |
MultiRC | 0.655 | 0.42| 
ReCoRD | 0.82976 | 0.838|
Cosmos | 0.81 | 0.774|


sleepinyourhat:
- Add a description of what was added and how it was tested/verified.
- Make sure tests pass, including style.
- Make sure that you only include the intended changes. It looks like the very first change is a stale/inappropriate README change.

This is general good practice. Please do all of the above for every non-trivial PR in the future, before contacting potential reviewers or turning off [WIP].

jiant/__main__.py
```diff
@@ -567,7 +567,20 @@ def main(cl_arguments):
     if args.do_pretrain:
         # Train on pretrain tasks
         log.info("Training...")
-        stop_metric = pretrain_tasks[0].val_metric if len(pretrain_tasks) == 1 else "macro_avg"
+        if args.early_stopping_method != "auto":
```
HaokunLiu: this is the right direction, but you may want move it somewhere else, otherwise 
```
elif args.early_stopping_method in target_names:
                index = target_names.index(args.early_stopping_method)
                stop_metric = target_tasks[index].val_metric
```
won't have any effect.

jiant/models.py
```diff
@@ -746,6 +761,12 @@ def build_lm(task, d_inp, args):
     return hid2voc
 
 
+def build_mlm(task, d_inp, task_params, embedder):
+    " Build MLM components " ""
+    lm_head = embedder.get_pretrained_lm_head()
```
HaokunLiu: if build mlm is simply getting pretrained head, why does it need task, d_inp, task_params as argument as well?

jiant/huggingface_transformers_interface/modules.py
```diff
@@ -358,8 +361,8 @@ def get_pretrained_lm_head(self):
             self.input_module, cache_dir=self.cache_dir
         )
         lm_head = model_with_lm_head.lm_head
-        lm_head.predictions.decoder.weight = self.model.embeddings.word_embeddings.weight
-        return nn.Sequential(lm_head, nn.LogSoftmax(dim=-1))
+        lm_head.decoder.weight = self.model.embeddings.word_embeddings.weight
```
HaokunLiu: thank you!

jiant/models.py
```diff
@@ -1154,6 +1176,63 @@ def _lm_forward(self, batch, task, predict):
             pass
         return out
 
+    def _masked_lm_forward(self, batch, task, predict):
+        mlm_probability = 0.15
+        out = {}
+        sent_encoder = self.sent_encoder
+        tokenizer = self.sent_encoder._text_field_embedder.tokenizer
+        # mask_idx = self.sent_encoder._text_field_embedder._mask_id #
+        mask_idx = tokenizer.convert_tokens_to_ids("<mask>")
```
HaokunLiu: better use _mask_id and _pad_id from the self.sent_encoder._text_field_embedder, not every transfromer uses  <mask> as the mask token.

HaokunLiu: in addition, I don't think this is using the right index.
when these indexes get into the transformer, they will be subtracted by 2. the transformer will see them as something other than <mask>.
I'm also not sure about pad_idx. Can you try printing out padding_mask, and see if it is fully zero?

jiant/preprocess.py
```diff
@@ -650,6 +652,14 @@ def add_task_label_vocab(vocab, task):
     This can then be accessed when generating Instances, either via a custom
     Indexer or by invoking the namespace when creating a LabelField.
     """
+    if isinstance(task, MLMTask):
+        tokenizer =  RobertaTokenizer.from_pretrained("roberta-large")
```
HaokunLiu: can you try to generalize this for other input modules? you should be able to get corresponding tokenizer by calling "get_tokenizer"

jiant/tasks/lm.py
```diff
@@ -172,3 +175,102 @@ def __init__(self, path, *args, **kw):
...
+        def _make_instance(sent_):
+            """ Forward targs adds <s> as a target for input </s>
+            and bwd targs adds </s> as a target for input <s>
+            to avoid issues with needing to strip extra tokens
+            in the input for each direction """
+            sent_ = model_preprocessing_interface.boundary_token_fn(sent_)  # Add <s> and </s>
+            input_sent = sentence_to_text_field(sent_, indexers)
+            d = {
+                "input": input_sent,
+                "targs": SequenceLabelField(
```
HaokunLiu: are input and targs  the same sequence? do we need two separate domains?

jiant/tasks/lm.py
```diff
@@ -172,3 +175,102 @@ def __init__(self, path, *args, **kw):
...
+        return
+
+    def get_data_iter(self, path):
+        """Loading data file and tokenizing the text
+        Args:
+            path: (str) data file path
+        """
+        import csv
+
+        f = open(path, "r")
```
HaokunLiu: I'm not very sure about this. When I worked on blimp, I remember I saw wikitext103 is already tokenized by Moses. If that is the case, you may need to detokenize it first before tokenize using self._tokenizer_name.

jiant/models.py
```diff
@@ -158,8 +158,10 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
         )
         d_sent = args.d_word
         log.info("Using PRPN sentence encoder!")
-    elif any(isinstance(task, LanguageModelingTask) for task in tasks) or args.sent_enc == "bilm":
-        assert_for_log(args.sent_enc in ["rnn", "bilm"], "Only RNNLM supported!")
+    elif any(isinstance(task, LanguageModelingTask) for task in tasks):
```
HaokunLiu: wait, how do we deal with language modeling task when we have a input_module like bert? do we have that supported now?

jiant/models.py
HaokunLiu: I guess padding_mask should be correct, since the targs uses "something_tag" as its namespace. But mask_idx probably isn't.

HaokunLiu: To verify, you can try directly evaluate pretrained model, with no finetuning at all. If it is correctly set to be mask_idx, you will see a fairly low ppl.

jiant/models.py
pruksmhc: Good point, I'll delete those.

pruksmhc:
Thanks for the comments Sam, I think I was a bit rushed for asking for reviews (although Haokun, thank you for bringing up some good points!).  I think we should expand on the contributing guidelines if we haven't in the google docs on what is needed before asking for a review (such as tests, what type of verification is needed to prove a PR is working, etc).

jiant/models.py
pruksmhc: @HaokunLiu I'll take your suggestion about using the mask_id and _pad_id from the embedder. We fixed the vocab bug by adding a label namespace of _tags to MaskedLanguageModelingTask. And good point about mask_id, i'll check for that. I guess I can add everything (except padding idx) in the inputs by 2 before passing it into the transformer

jiant/preprocess.py
pruksmhc: Ah yes will do

jiant/tasks/lm.py
pruksmhc: Yes, I think so. I don't think they should be two separate domains because the model is supposed to predict the correct original word that was replaced by MASK/a random word/kept as is.

jiant/__main__.py
pruksmhc: What do you mean? This should only be in pretrain_tasks because multitask learning is only in pretraining.

HaokunLiu: In that case, why will there be a case for early_stopping_method in target_names but not in pretrain_names?

jiant/models.py
pruksmhc: Yes that is supported

jiant/__main__.py
pruksmhc: Ah yeah, there wouldn't. I'll take it out

jiant/preprocess.py
```diff
@@ -661,7 +665,7 @@ def add_task_label_vocab(vocab, task):
         return
     log.info("\tTask '%s': adding vocab namespace '%s'", task.name, namespace)
 
-    if isinstance(task, SequenceGenerationTask):
+    if isinstance(task, SequenceGenerationTask) and not isinstance(task, LanguageModelingTask):
```
pruksmhc: This change is because of bad code design before (I think SequenceGenerationTask was initially created for machine translation and only supported non-BERT models). I think SequenceGenerationTask should be refactored in a separate PR, so just adding this here for now.

pruksmhc:
@pyeres  @sleepinyourhat  ready for review!

sleepinyourhat:
"Adding Masked Language Modeling Task for RoBERTa
What this version of MLM supports: Transformer-based embedders"
-> Clarify whether this covers all Transformers or just RoBERTa?

sleepinyourhat:
Could you share numbers for RoBERTa + CCG => Target? That's the evaluation that's most relevant to us.

jiant/models.py
```diff
@@ -160,16 +161,18 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
...
-                args.input_module == "elmo"
-                or args.input_module.startswith("bert")
-                or args.input_module.startswith("xlnet")
-            ),
-            f"Using input_module = {args.input_module} for language modeling is probably not a "
-            "good idea, since it allows the language model to use information from the right-hand "
-            "context.",
-        )
+        if not any(isinstance(task, MaskedLanguageModelingTask) for task in tasks):
+            # If an autoregressive LanguageModelingTask
```
sleepinyourhat: I'm not sure this is crucial, but it'd be nice to avoid this weird comment logic by explicitly introducing an AutoregressiveLMTask class.

jiant/models.py
```diff
@@ -746,6 +752,12 @@ def build_lm(task, d_inp, args):
     return hid2voc
 
 
+def build_mlm(embedder):
+    " Build MLM components " ""
```
sleepinyourhat: Fix style here.

jiant/models.py
```diff
@@ -1160,6 +1174,69 @@ def _lm_forward(self, batch, task, predict):
             pass
         return out
 
+    def _masked_lm_forward(self, batch, task, predict):
+        mlm_probability = 0.15
```
sleepinyourhat: I think hard-coded constants like this are only okay if we make it clear (in docs, method names, etc.) that we're _only_ supporting the exact RoBERTa setup. Otherwise, make this configurable?

jiant/models.py
```diff
@@ -1160,6 +1174,69 @@ def _lm_forward(self, batch, task, predict):
...
+        out = {}
+        sent_encoder = self.sent_encoder
+        tokenizer_name = self.sent_encoder._text_field_embedder.input_module
+        vocab_size = (
+            self.sent_encoder._text_field_embedder.model.embeddings.word_embeddings.num_embeddings
+        )
+        tokenizer = get_tokenizer(tokenizer_name)
+        input_key = self.sent_encoder._text_field_embedder.tokenizer_required
+        # mask_idx = self.sent_encoder._text_field_embedder._mask_id #
```
sleepinyourhat: ?

jiant/models.py
```diff
@@ -1160,6 +1174,69 @@ def _lm_forward(self, batch, task, predict):
...
+
+        probability_matrix = torch.full(labels.shape, mlm_probability, device=inputs.device)
+        padding_mask = labels.eq(0)
+        probability_matrix.masked_fill_(padding_mask, value=0.0)
+
+        masked_indices = torch.bernoulli(probability_matrix).to(
+            device=inputs.device, dtype=torch.uint8
+        )
+
+        labels[~masked_indices] = -100  # We only compute loss on masked tokens
```
sleepinyourhat: Why is this needed, then?

jiant/models.py
```diff
@@ -1160,6 +1174,69 @@ def _lm_forward(self, batch, task, predict):
...
+
+        # 80% of the time, we replace masked input tokens with tokenizer.mask_token ([MASK])
+        indices_replaced = (
+            torch.bernoulli(torch.full(labels.shape, 0.8)).to(
```
sleepinyourhat: Could you structure the code here in a way that's more similar to the above?

jiant/preprocess.py
```diff
@@ -261,6 +263,7 @@ def _build_vocab(args: config.Params, tasks: List[Task], vocab_path: str):
     for task in tasks:  # add custom label namespaces
         # TODO: surface more docs for add_task_label_vocab:
         add_task_label_vocab(vocab, task)
+
```
sleepinyourhat: ?

jiant/preprocess.py
sleepinyourhat: Reasonable, though this does raise a real conflict: The _normal_ LM task is a sequence generation task, but MLM isn't. It may be simpler to just make MLM inherit directly from some broad task class, rather than from LMTask.

jiant/models.py
```diff
@@ -1160,6 +1174,69 @@ def _lm_forward(self, batch, task, predict):
...
+                device=inputs.device, dtype=torch.uint8
+            )
+            & masked_indices
+            & ~indices_replaced
+        )
+        random_words = torch.randint(
+            len(tokenizer), labels.shape, dtype=torch.long, device=inputs.device
+        )
+        inputs[indices_random] = random_words[indices_random]
+        # Add 2 to all non-special tokens due to correct_sent logic in Transformer-based
```
sleepinyourhat: This seems right to the extent that I understand it, but I don't think I'm smart enough to be sure of that. This whole fn could really use tests.

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+class MaskedLanguageModelingTask(LanguageModelingTask):
+    """
+       Generic Masked Language Modeling Task
+    """
+
+    pass
+
+
+@register_task("mlm", rel_path="WikiText103/")
+class MLMTask(MaskedLanguageModelingTask):
```
sleepinyourhat: This is worrying. Could we rework this to avoid the distinction between "MaskedLanguageModeling" and "MLM"?

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+    """
+    Masked language modeling task on Toronto Books dataset
```
sleepinyourhat: Maybe add this to the task name? Or specify that we're using data _in the format of_ Toronto Books.

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+    Attributes:
+        max_seq_len: (int) maximum sequence length
+        min_seq_len: (int) minimum sequence length
+        files_by_split: (dict) files for three data split (train, val, test)
```
sleepinyourhat: split => splits

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+            labels.append(word)
+        for path in self.files_by_split:
+            for sent in self.get_data_iter(self.files_by_split[path]):
+                for tok in sent:
+                    if tok not in labels:
+                        labels.append(tok)
+        return labels
+
+    def update_metrics(self, out, batch=None):
+        # self.scorer1(logits,labels)
```
sleepinyourhat: ?

jiant/tasks/lm.py
sleepinyourhat: Also, what about Wikipedia text?

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+        """Loading data file and tokenizing the text
+        Args:
+            path: (str) data file path
+        """
+        import csv
+
+        f = open(path, "r")
+        reader = csv.reader(f)
+        text = list(reader)
+        moses_tokenizer = get_tokenizer("MosesTokenizer")
```
sleepinyourhat: Why do we need this?

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+        self, split, indexers, model_preprocessing_interface
+    ) -> Iterable[Type[Instance]]:
+        """Process a language modeling split by indexing and creating fields.
+        Args:
+            split: (list) a single list of sentences
+            indexers: (Indexer object) indexer to index input words
+        """
+
+        def _make_instance(sent_):
+            """ Forward targs adds <s> as a target for input </s>
```
sleepinyourhat: I'm confused—is this a leftover comment from the normal LM code? If this is true, it doesn't sound appropriate for an MLM.

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+                ),
+            }
+            return Instance(d)
+
+        for sent in split:
+            yield _make_instance(sent)
+
+
+@register_task("mlm_toronto", rel_path="toronto/")
+class TorontoLanguageModelling(MaskedLanguageModelingTask):
```
sleepinyourhat: Should probably add 'Masked' to the name. TorontoBooksMLMTask?

jiant/tasks/lm.py
```diff
@@ -172,3 +177,121 @@ def __init__(self, path, *args, **kw):
...
+    """ Language modeling on the Toronto Books dataset
+    See base class: LanguageModelingTask
```
sleepinyourhat: Last line seems redundant.

jiant/trainer.py
```diff
@@ -436,6 +435,9 @@ def get_sampling_weights(
             sample_weights = 1 / task_n_train_examples
         elif weighting_method == "inverse_log_example":
             sample_weights = 1 / np.log(task_n_train_examples)
+        elif "examples-proportional-mixing" in weighting_method:
+            K = int(weighting_method.split("K=")[1])
```
sleepinyourhat: At least under GitHub's syntax highlighting, it seems like 'K' is being treated as if it were reserved...

sleepinyourhat: Update defaults.conf to document this option.

jiant/preprocess.py
pruksmhc: This was from black. I tried reverting it and black just added it back in.

jiant/__main__.py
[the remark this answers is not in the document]
pruksmhc: It'll be reasonable to test. Adding that to to-do list.

jiant/config/base_mlm_roberta.conf
```diff
@@ -0,0 +1,19 @@
+// Base config file for mlm experiments wit roberta
+include "defaults.conf"
+
+early_stopping_method=auto // Early stopping method. Options: task_name to only do early stopping based
```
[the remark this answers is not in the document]
pruksmhc: Having it end with _task might be a bit misleading though, as the default right now is to do early stopping based on the macro_avg.

jiant/tasks/lm.py
pruksmhc: I think this was a copy-pasta error. Fixing now.

jiant/tasks/lm.py
pruksmhc: The wikipedia corpus available is preprocessed using MosesTokenizer, so we need to detokenize it before tokenizing it with our tokenizer of choice.

jiant/tasks/lm.py
pruksmhc: Yes it is, removing.

jiant/models.py
pruksmhc: @phu-pmh can explain this better than I.

phu-pmh: nn.CrossEntropy ignores the idices with value = -100 by default. Therefore, we replace non-masked indices with -100 not to consider them in loss computation.

jiant/trainer.py
```diff
@@ -436,6 +435,9 @@ def get_sampling_weights(
             sample_weights = 1 / task_n_train_examples
         elif weighting_method == "inverse_log_example":
             sample_weights = 1 / np.log(task_n_train_examples)
+        elif "examples-proportional-mixing" in weighting_method:
+            max_K = int(weighting_method.split("K=")[1])
```
pyeres: Heads up: this will throw a `ValueError` if a user follows the example in the `defaults.conf`:

"...`set weighting_method=examples-proportional-mixingK==104857`"

jiant/trainer.py
```diff
@@ -436,6 +435,9 @@ def get_sampling_weights(
             sample_weights = 1 / task_n_train_examples
         elif weighting_method == "inverse_log_example":
             sample_weights = 1 / np.log(task_n_train_examples)
+        elif "examples-proportional-mixing" in weighting_method:
```
pyeres: For consistency with the other `weighting_method` names, please use underscores.

jiant/tasks/lm.py
```diff
@@ -172,3 +178,102 @@ def __init__(self, path, *args, **kw):
             "val": os.path.join(path, "valid.sentences.txt"),
             "test": os.path.join(path, "test.sentences.txt"),
         }
+
+
+@register_task("mlm", rel_path="WikiText103/")
+class MaskedLanguageModelingTask(Task):
```
pyeres: It shouldn't be possible to resolve the reference to `Task` without importing it, or am I missing something?

jiant/tasks/lm.py
```diff
@@ -172,3 +178,102 @@ def __init__(self, path, *args, **kw):
...
+        super().__init__(name, **kw)
+        self.scorer1 = Average()
+        self.scorer2 = None
+        self._label_namespace = self.name + "_labels"
+        self.val_metric = "%s_perplexity" % self.name
+        self.val_metric_decreases = True
+        self.max_seq_len = max_seq_len
+        self.min_seq_len = 0
+        self.target_indexer = {"words": SingleIdTokenIndexer(namespace="tokens")}
+        self.files_by_split = {
```
pyeres: I'm a little confused by the versions of WikiText103 data that are floating around*. Here we're using a version with `train.txt`, `valid.txt` and `test.txt`; for the [`wiki103-classif` task](https://github.com/nyu-mll/jiant/blob/master/jiant/tasks/tasks.py#L2129-L2144) we depend on `train.sentences.txt`, `valid.sentences.txt`, and `test.sentences.txt`. Can you clarify what version of the data is expected here (in the filename/comments). Can you please also upload the final version of this data to the shared data repo?

\* Side note: there are two more versions of this data in the shared data repo with different naming patterns.

jiant/models.py
pyeres: Thanks for the info @phu-pmh. Requesting that this info gets captured here with a comment in the code.

jiant/preprocess.py
```diff
@@ -661,7 +663,9 @@ def add_task_label_vocab(vocab, task):
         return
     log.info("\tTask '%s': adding vocab namespace '%s'", task.name, namespace)
 
-    if isinstance(task, SequenceGenerationTask):
+    if isinstance(task, SequenceGenerationTask) and not isinstance(
+        task, MaskedLanguageModelingTask
```
pyeres: Looking for a comment here (in the code) to explain why this special case/handling is added.

jiant/tasks/lm.py
```diff
@@ -172,3 +178,102 @@ def __init__(self, path, *args, **kw):
             "val": os.path.join(path, "valid.sentences.txt"),
             "test": os.path.join(path, "test.sentences.txt"),
         }
+
+
+@register_task("mlm", rel_path="WikiText103/")
```
pyeres: Requesting rename: `mlm` task name seems overly broad (there can be other `mlm`-style tasks, right?). Borrow the naming pattern from [another task that uses WikiText103 data](https://github.com/nyu-mll/jiant/blob/61440732f4df2c54e68e59ead34b8656bf52af3b/jiant/tasks/tasks.py#L2129) where the name mentions the data and model-type?

pyeres:
Hi @pruksmhc — 

For the following items I gave preliminary feedback in the PR thread:
* adds a new config param `early_stopping_method`
* adds new sampling weights method `examples-proportional-mixing`
* renames `LanguageModelingTask` to `AutoregressiveLanguageModelingTask`

I'm still working on reviewing these items:
* adds new task class `MaskedLanguageModelingTask`, and task `mlm` (using `WikiText103` data)
* adds new task-specific module builder for `MaskedLanguageModelingTask`s
* adds new MLM forward method to `MultiTaskModel`
* updates `huggingface_transformers_interface`:
     * adds integer token indices for special mask tokens in `BertEmbedderModule`, `RobertaEmbedderModule`, `AlbertEmbedderModule` and `XLNetEmbedderModule` classes.
     * updates `lm_head` construction for `RobertaEmbedderModule`, `XLNetEmbedderModule`, `OpenAIGPTEmbedderModule`, and `GPT2EmbedderModule`

btw, if you think these bullets are accurate and possibly helpful to other reviewers, please add them to the PR description field.

jiant/models.py
```diff
@@ -1160,6 +1176,69 @@ def _lm_forward(self, batch, task, predict):
...
+            device=inputs.device, dtype=torch.uint8
+        )
+        indices_replaced = bernoulli_mask & masked_indices
+        inputs[indices_replaced] = mask_idx
+
+        # 10% of the time, we replace masked input tokens with random word
+        bernoulli_mask = torch.bernoulli(torch.full(labels.shape, 0.5)).to(
+            device=inputs.device, dtype=torch.uint8
+        )
+        indices_random = bernoulli_maz & masked_indices & ~indices_replaced
```
pyeres: Heads up: looks like this will fail with NameErrror on L1222.

pruksmhc:
Cool, thanks for the comments Phil. I'll take a look at this in a few hours.

jiant/tasks/lm.py
```diff
@@ -69,7 +66,9 @@ def get_metrics(self, reset=False):
 
     def load_data(self):
         # Data is exposed as iterable: no preloading
-        pass
+        self.examples_by_split = {}
+        for split in self.files_by_split:
+            self.examples_by_split[split] = list(self.get_data_iter(self.files_by_split[split]))
```
pyeres: These changes will impact `AutoregressiveLanguageModelingTask`'s subclasses. Before this change data was "...exposed as iterable: no preloading", with this change it looks like calls to `load_data()` will store data in a new `self.examples_by_split` field in the Task objects. 

Can you please to confirm that this change is intended? If this change is necessary, with the final PR please provide confirmation that any performance changes are acceptable.

jiant/tasks/lm.py
```diff
@@ -172,3 +179,134 @@ def __init__(self, path, *args, **kw):
...
+
+        f = open(path, "r", encoding="utf-8")
+        reader = csv.reader(f)
+        text = list(reader)
+        moses_tokenizer = get_tokenizer("MosesTokenizer")
+        for i in range(10):
+            row = text[i]
+            untokenized_toks = moses_tokenizer.detokenize(row)
+            toks = "".join(untokenized_toks)
+            yield tokenize_and_truncate(self._tokenizer_name, toks, self.max_seq_len)
```
pyeres: Please refactor to move any necessary imports to the top of the file, ensure that the file object is closed, and avoid converting the file reading generator to a list inside the generator you're building. You can find an example implementation of `get_data_iter()` in `AutoregressiveLanguageModelingTask`.

pyeres:
I added some comments regarding the changes to `jiant/tasks/lm.py` above.

I'm also watching for updates addressing [this comment](https://github.com/nyu-mll/jiant/pull/1030#discussion_r396130068) before looking closely at these changes:

> * adds new MLM forward method to `MultiTaskModel`

And I'm watching for updates in [this thread](https://github.com/nyu-mll/jiant/pull/1030/files#r396128574) before looking closely at these changes:

> * updates `huggingface_transformers_interface`:
>   
>   * adds integer token indices for special mask tokens in `BertEmbedderModule`, `RobertaEmbedderModule`, `AlbertEmbedderModule` and `XLNetEmbedderModule` classes.
>   * updates `lm_head` construction for `RobertaEmbedderModule`, `XLNetEmbedderModule`, `OpenAIGPTEmbedderModule`, and `GPT2EmbedderModule`

Please ping me in this thread when you're ready for me to take another look.

jiant/models.py
```diff
@@ -1160,6 +1176,70 @@ def _lm_forward(self, batch, task, predict):
...
+
+        probability_matrix = torch.full(labels.shape, mlm_probability, device=inputs.device)
+        padding_mask = labels.eq(0)
+        probability_matrix.masked_fill_(padding_mask, value=0.0)
+
+        masked_indices = torch.bernoulli(probability_matrix).to(
+            device=inputs.device, dtype=torch.uint8
+        )
+        tokenizer_name = self.sent_encoder._text_field_embedder.tokenizer_required
+        labels, _ = self.sent_encoder._text_field_embedder.correct_sent_indexing(
```
pruksmhc: It is necessary to correct_sent_indexing for labels here because the namespace for labels is padded by PAD and UNK. It was necessary to do this because the version we had was preprocessed with MosesTokenizer and there was one token that was being preprocessed wrongly.

pruksmhc: This might be no longer necessary once we use the non-preprocessed wikitext.

pruksmhc:
Phil and I were discussing about where to extract out the MLM dynamic masking logic to for easier testing. I suggested we put it under the MaskedLanguageModeling task, but Phil put down that he would prefer an implementation where the dynamic masking is pulled into a single-purpose, testable, well-documented function (and I’d be more comfortable with this being a function invoked by models doing RoBERTa-style MLM training, than a method of Task objects).
@pyeres  I think we're saying the same thing. I won't put the masking code under the Task class, but only the MLMTask class.

pyeres:
> Phil and I were discussing about where to extract out the MLM dynamic masking logic to for easier testing. I suggested we put it under the MaskedLanguageModeling task, but Phil put down that he would prefer an implementation where the dynamic masking is pulled into a single-purpose, testable, well-documented function (and I’d be more comfortable with this being a function invoked by models doing RoBERTa-style MLM training, than a method of Task objects).
> @pyeres I think we're saying the same thing. I won't put the masking code under the Task class, but only the MLMTask class.

Yeah, wherever this functionality is implemented we want it to be single-purpose, testable, and documented. I agree with you @pruksmhc — I don't think the differences between the approaches we were discussing are significant in the scope of this code review.

jiant/preprocess.py
pruksmhc: I think his was from some artifact old code from when LM was first put into the repo and test on non-transformer models. I'm actually going to remove this block, since our Transformer models add special tokens elsewhere to the vocab.

jiant/tasks/lm.py
pruksmhc: Yes, this change is intended. Before, we weren't saving the data anywhere, and we were re-loading the data everytime we wanted to count the number of examples of get the sentences in the dataset to create the vocab.

jiant/tasks/lm.py
pruksmhc: Phu and I are using a raw, unpreprocessed Wikipedia dataset

pruksmhc: We'll make that clear.

jiant/tasks/lm.py
```diff
@@ -172,3 +179,132 @@ def __init__(self, path, *args, **kw):
...
+
+@register_task("wikipedia_corpus_mlm", rel_path="wikipedia_corpus_small/")
+class MaskedLanguageModelingTask(Task):
+    """
+    Masked language modeling task on Wikipedia dataset
+    Attributes:
+        max_seq_len: (int) maximum sequence length
+        min_seq_len: (int) minimum sequence length
+        files_by_split: (dict) files for three data split (train, val, test)
+    We are currently using an unpreprocessed version of the Wikipedia corpus
```
pruksmhc, line 192: @phu-pmh  can you double check this description of what dataset we're currently using?

jiant/huggingface_transformers_interface/modules.py
pruksmhc: The  LM decoder weights being returned in the head was not initialized from the pretrained model, which was a bug.

jiant/tasks/lm.py
pyeres: Yes, before your change the data was streamed by inheriting subclasses of `AutoregressiveLanguageModelingTask` (and like you say, it looks like it was streamed twice: once to count examples, and once to create the vocab).

I would assume however that the decision to stream the data was made to avoid loading the data into memory (and it also looks like the `pass`-logic that your change replaces was added in an effort to [delay data-loading](https://github.com/nyu-mll/jiant/pull/662)).

It looks to me like this change could cause a breakage (if the modified tasks have data that won't fit in memory — or if they're used in experiments where memory is already tight). So, if you want to make this change please confirm that the subclasses that inherit this behavior still work, and the change in memory footprint is acceptable.

**TL;DR** I wouldn't recommend making this change because it requires work to confirm that it's safe.

jiant/tasks/lm.py
phu-pmh: We did hyperlink and xml removal etc, and did sentence-level tokenization. Each sentence is not processed.

jiant/tasks/lm.py
pruksmhc: Yeah, I guess it is a tradeoff between time and memory. Before I made this change, it took ~13 hours to do preprocessing, and making this change cut that to 1/3 the time. I see your point about memory though. 
In that case I'll just revert the change.

jiant/tasks/lm.py
```diff
@@ -181,14 +181,17 @@ def __init__(self, path, *args, **kw):
...
 class MaskedLanguageModelingTask(Task):
     """
     Masked language modeling task on Wikipedia dataset
     Attributes:
         max_seq_len: (int) maximum sequence length
         min_seq_len: (int) minimum sequence length
         files_by_split: (dict) files for three data split (train, val, test)
+    We are currently using an unpreprocessed version of the Wikipedia corpus
+    that consists of 5% of the data. Please reach out to jiant admin if you 
+    would like access to this dataset.
```
pyeres: Is the task registered as `wikipedia_corpus_mlm` (using dataset under relative path `/wikipedia_corpus_small`) an experiment with some research significance that users would want to reproduce (or is it perhaps a toy dataset being used to demo `MaskedLanguageModelingTask`)?

pyeres: Oops, looks like I commented on some outdated code. I'll look at the updated code and refile my comment if necessary.

phu-pmh: It's not toy data. The full Wikipedia is very large and jiant currently does not have ability to load the entire thing. So, I tried to extract a subset with the same size as Wikitext 103, which is around 5% of full Wikipedia.

jiant/huggingface_transformers_interface/modules.py
pyeres: Thanks for the clarification. Can you also spell out for me what the the motivation was for removing `nn.Sequential()` / `nn.LogSoftmax(dim=-1)` on the neighboring line? (related to [Sam's question](https://github.com/nyu-mll/jiant/pull/1030/files#r396128626))

jiant/tasks/lm.py
pyeres: Thanks for clarifying, @phu-pmh. Maybe relevant: there was a [recent discussion](https://github.com/nyu-mll/jiant/pull/1030#discussion_r398754495) and [change](https://github.com/nyu-mll/jiant/pull/1030/commits/8ca1eba4828362beda6f0436c2585f768ef2c28c) to another LM task to reduce its memory footprint. It may be relatively easy to make a similar change here to allow you to use the full dataset (if memory, not time, is the concern).

But if you want to introduce a modified dataset with this task, please also submit the code you used to construct the dataset. There's an example of a data preprocessing script under [`jiant/scripts`](https://github.com/nyu-mll/jiant/blob/master/scripts/winograd/preprocess_winograd.py). But all your script would need to do is document/link the data it takes as input, perform your preprocessing, and save the output as it's expected by your task.

jiant/huggingface_transformers_interface/modules.py
pyeres: Thanks — I see that after your changes `get_pretrained_lm_head()` will return `lm_head` without applying `nn.Sequential()` and `nn.LogSoftmax()` for these `HuggingfaceTransformersEmbedderModule`s:
- `RobertaEmbedderModule`
- `XLNetEmbedderModule`
- `OpenAIGPTEmbedderModule`
- `GPT2EmbedderModule`
- `TransfoXLEmbedderModule`

It looks like `get_pretrained_lm_head()` will continue to `return nn.Sequential(lm_head, nn.LogSoftmax(dim=-1))` for these `HuggingfaceTransformersEmbedderModule`s:
- `AlbertEmbedderModule`
- `BertEmbedderModule`
- `XLMEmbedderModule`

For consistency can you update the remaining `HuggingfaceTransformersEmbedderModule`s?

jiant/tasks/lm.py
```diff
@@ -172,3 +178,131 @@ def __init__(self, path, *args, **kw):
...
+        self.val_metric_decreases = True
+        self.max_seq_len = max_seq_len
+        self.min_seq_len = 0
+        self.target_indexer = {"words": SingleIdTokenIndexer(namespace="tokens")}
+        self.files_by_split = {
+            "train": os.path.join(path, "train.txt"),
+            "val": os.path.join(path, "valid.txt"),
+            "test": os.path.join(path, "test.txt"),
+        }
+        self.examples_by_split = {}
```
pruksmhc, line 217: I'll revert this in next commit

jiant/tasks/lm.py
pruksmhc: The preprocessing code we used was a slight modification of what is here (@phu-pmh 
can speak more to this), so it might not be the best idea to copy paste it into scripts. https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/LanguageModeling/BERT/data . Perhaps we could put instructions (maybe in the documentation in the Task class?) about how to generate it instead.

phu-pmh: For English Wikipedia, I think the nvidia code is ready to use. I only added modifications so that we can do cross-lingual experiments.

pyeres:
Thanks for your recent changes, @pruksmhc.

It looks like there are only a few open comments remaining:
1. Providing [data/script/instructions for the new task](https://github.com/nyu-mll/jiant/pull/1030#discussion_r404532750).
2. Is [correct_sent_indexing](https://github.com/nyu-mll/jiant/pull/1030#discussion_r399009285) necessary with the final configuration of the task/data?
3. I think we left [our discussion of the token masking functionality](https://github.com/nyu-mll/jiant/pull/1030#issuecomment-605753516) with a plan to extract the masking logic into a single-purpose/documented/testable function/method.
4. It looks like there are also a few smaller open comment threads ([here](https://github.com/nyu-mll/jiant/pull/1030#discussion_r404428109) and [here](https://github.com/nyu-mll/jiant/pull/1030#discussion_r396131248)).

Finally, as you suggested, we'll want to re-run your [validation experiments](https://github.com/nyu-mll/jiant/pull/1030#issue-385266255) after the changes are in.

pruksmhc:
@pyeres , for 3, I realized that 1. the masking code is a little hard to test because it _randomly_ masks, which makes any asserts hard and 2. code is from the well-maintained Transformers library. Thus, I decided to hold off on unit tests for that part of the code.  I'll still extract it out into its own function though, but just a heads up.

pruksmhc:
and 2. is still necessary.

jiant/models.py
```diff
@@ -1160,6 +1176,69 @@ def _lm_forward(self, batch, task, predict):
...
+        # Masking code from https://github.com/huggingface/transformers/blob/master/examples/run_language_modeling.py
+        probability_matrix = torch.full(labels.shape, mlm_probability, device=inputs.device)
+        padding_mask = labels.eq(0)
+        probability_matrix.masked_fill_(padding_mask, value=0.0)
+
+        masked_indices = torch.bernoulli(probability_matrix).to(
+            device=inputs.device, dtype=torch.uint8
+        )
+        tokenizer_name = self.sent_encoder._text_field_embedder.tokenizer_required
+        labels, _ = self.sent_encoder._text_field_embedder.correct_sent_indexing(
```
pyeres, line 1191: It looks like the labels will be modified (as a result of `correct_sent_indexing()`). It looks like the inputs aren't getting the same adjustment here. Is that intentional/correct?

pruksmhc: Yes because we do it here for inputs: https://github.com/nyu-mll/jiant/blob/master/jiant/huggingface_transformers_interface/modules.py#L401

jiant/tasks/lm.py
```diff
@@ -172,3 +179,173 @@ def __init__(self, path, *args, **kw):
...
+class MaskedLanguageModelingTask(Task):
+    """
+    Masked language modeling task on Wikipedia dataset
+    Attributes:
+        max_seq_len: (int) maximum sequence length
+        min_seq_len: (int) minimum sequence length
+        files_by_split: (dict) files for three data split (train, val, test)
+    We are currently using an unpreprocessed version of the Wikipedia corpus
+    that consists of 5% of the data. You can generate the data using code from
+    https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/LanguageModeling/BERT/data.
```
pruksmhc, line 194: @pyeres here's the instructions for data generation (point 2 you raised).

pyeres: Thanks — it looks like the link points to a directory with a number of scripts. Should these instructions be made more specific to help the user reproduce this task?

jiant/tasks/lm.py
```diff
@@ -172,3 +179,173 @@ def __init__(self, path, *args, **kw):
...
+        """Yield sentences, used to compute vocabulary.
+        """
+        for split in self.files_by_split:
+            # Don't use test set for vocab building.
+            if split.startswith("test"):
+                continue
+            for sent in self.get_data_iter(self.files_by_split[split]):
+                yield sent
+
+    def mlm_dynamic_masking(self, inputs, labels, mask_idx, tokenizer_name, sent_encoder):
```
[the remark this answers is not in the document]
pruksmhc: Added

pruksmhc:
Here's the current validation checks. 
|  Performance metric | CCG w/ RoBERTa     | CCG w/ ALBERT   |   QQP w/ RoBERTa  | QQP w/ ALBERT |
| ------------- | ------------- |:-------------:| :-------------:| :-------------:| 
| Perplexity | 4.1689 |   | | |
| Performance of interm task   | 0.9530  |  | | |

Issue #973, opened 2019-12-03
Question: Why both load_eval_checkpoint and load_target_train_checkpoint?


sleepinyourhat:
I think we may have worked our way into a bit of redundancy through historical accident: We have two flags that specify checkpoints to load upon launch.

They're loaded in different parts of the code, but in practice, I think you can always derive that decision from the other flags (if do_target_task_train, then check load_target_train_checkpoint). Is there a reason we shouldn't just merge them?

zphang:
*This is an automatically generated comment.*

As we update `jiant` to v2.x, `jiant` v1.x has been migrated to https://github.com/nyu-mll/jiant-v1-legacy. As such, we are closing all issues relating to `jiant` v1.x in this repository.

If this issue is still affecting you in `jiant` v1.x, please follow up at https://github.com/nyu-mll/jiant-v1-legacy/issues/973.

If this issue is still affecting you in `jiant` v2.x, reopen this issue or create a new one.

Issue #937, opened 2019-10-17
Weight Decay configuration for BERT?


zphang:
The Transformers/pytorch_pretrained_bert version of the BERT training code applies weight decay to only a subset of the parameters: https://github.com/huggingface/transformers/blob/0856a231c0d6f3abb193e99ecf52ff76edb75fb8/examples/run_classifier.py#L482-L509

(I am referencing pytorch_pretrained_bert v0.6 because AllenNLP uses BertAdam from that version.)

As far as I can tell, we do not have the equivalent setup in jiant. I believe the current setup applies the default weight decay of 0.01 (which is the correct weight decay for the parameters that should have it applied) to all parameters.

Given that we have been able to train performant models with the current setup, I'm not sure if this qualifies as a bug or if we are satisfied with this arrangement.

W4ngatang:
:+1:

There's also been some changes to optimizers that we haven't kept up with, e.g. `BertAdam` has an option for `amsgrad` that defaults to `False`.

zphang:
*This is an automatically generated comment.*

As we update `jiant` to v2.x, `jiant` v1.x has been migrated to https://github.com/nyu-mll/jiant-v1-legacy. As such, we are closing all issues relating to `jiant` v1.x in this repository.

If this issue is still affecting you in `jiant` v1.x, please follow up at https://github.com/nyu-mll/jiant-v1-legacy/issues/937.

If this issue is still affecting you in `jiant` v2.x, reopen this issue or create a new one.

Issue #918, opened 2019-09-27
If task is in pretrain_tasks and target_tasks, target_train_data_fraction always overrides pretrain_data_fraction


avnermay:
https://github.com/nyu-mll/jiant/blob/c36b74e315226334120639d16a7924fd43345a1f/jiant/preprocess.py#L380-L393

If a task is specified in pretrain_tasks and also in target_tasks, then the target_train_data_fraction always overrides the pretrain_data_fraction.  This happens at line 391 of jiant/preprocess.py (see link above).  This is true even if do_target_task_training = 0.  Note that in the instructions in defaults.conf, it says: "If you want to train and evaluate on a single task without doing any new pretraining, you should set target_tasks and pretraining_tasks to the same task, set do_pretrain to 1, and do_target_task_training to 0."

Thus, if I do what is recommended above, and set pretrain_data_fraction < 1, but don't override the default target_data_fraction, then the data fraction gets overriden to be 1.
As an example, this occurs if I pass in the following as overrides for defaults.conf: "pretrain_tasks = qnli, target_tasks = qnli, pretrain_data_fraction = 0.1, do_pretrain = 1, do_target_task_training = 0, do_full_eval = 1".

sleepinyourhat:
Good catch—thanks for writing this up!

@avnermay @pruksmhc @W4ngatang @iftenney @HaokunLiu - Any opinions on how best to address this? (And any interest in taking it on? I won't have the bandwidth for a few weeks.) 

We should probably preserve the ability to pretrain and fine-tune on the same task, which would mean that we'd need two different iterators with two potentially different fractions. It might be best to load the full dataset here, but create the actual iterators at the start of each training phase...

pyeres:
Fixed by PR https://github.com/nyu-mll/jiant/pull/1070.

Issue #917, opened 2019-09-23
Allow data_fraction args to be set for specific tasks


sleepinyourhat:
Ideally, we'd support something like `mnli += {pretrain_data_fraction = 0.5}, pretrain_tasks = {mnli,boolq}`. Currently, pretrain_data_fraction is just a global argument for all pretraining tasks.

zphang:
*This is an automatically generated comment.*

As we update `jiant` to v2.x, `jiant` v1.x has been migrated to https://github.com/nyu-mll/jiant-v1-legacy. As such, we are closing all issues relating to `jiant` v1.x in this repository.

If this issue is still affecting you in `jiant` v1.x, please follow up at https://github.com/nyu-mll/jiant-v1-legacy/issues/917.

If this issue is still affecting you in `jiant` v2.x, reopen this issue or create a new one.

PR #1035, opened 2020-03-10
[WIP] Trainer & __main__ interface refactor, some gpu related refactor

HaokunLiu:
In preparation for the incoming FP16 #472 , I found the required usage of apex conflict with our pipleline.

Since we only call Trainer.train once for each Trainer created. The conflict can be resolved by refactoring the Trainer, move some code setting up training process in Trainer.train() and some code making stop_metric and should_decrease earlier, so that they will be inside Trainer.__init__(), rather than Trainer.train().

In doing so, I found a few correct, but itching things in trainer. It may be a good opportunity to clear the history debts in trainer. Here are a few related issues in consideration.
#973 #937 #918 #917 #869 #847 #828 #759 #573 #537

It is not certain how many of them are easily solvable, this is just a temporary note. When this PR is ready, I'll update the issues actually got addressed, and left untouched.

### DIFF
--- a/cola_inference.py
+++ b/cola_inference.py
@@ -155,7 +155,7 @@ def main(cl_arguments):
     # Build or load model #
     model = build_model(args, vocab, word_embs, tasks, cuda_device)
     log.info("Loading existing model from %s...", cl_args.model_file_path)
-    load_model_state(model, cl_args.model_file_path, args.cuda, [], strict=False)
+    load_model_state(model, cl_args.model_file_path, [], strict=False)
 
     # Inference Setup #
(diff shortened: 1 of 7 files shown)


HaokunLiu:
The update will be pretty wide, and some tests that depend on the previous trainer also need to updates, some addition tests may also be necessary.

If anyone else would like to join, that will be very helpful.

pyeres:
> If anyone else would like to join, that will be very helpful.

I'd like to join (especially for the test development).

jeswan:

PR #1057, opened 2020-04-10
Taskmaster scripts

HaokunLiu:
People involved in the taskmaster project, please check the scripts and point out on things that don't match our plan.


HaokunLiu:
Taskmaster branch is not updated for a while, some of the changes are pulled from master branch

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+# Contributing to `jiant`
```
sleepinyourhat, line 1: Please rebase. Many of these changes are irrelevant mismatch junk.

scripts/taskmaster/optuna_hp_search/collect_trials.py
```diff
@@ -0,0 +1,111 @@
+import optuna
+import argparse
+import os
+import json
+import pandas
+import numpy
+
+RESULT_DIR = "/scratch/hl3236/jiant_results"
```
sleepinyourhat, line 8: If we'll have to reuse this at all, could you add a short header comment explaining what this is for?

scripts/taskmaster/optuna_hp_search/generate_scripts.py
```diff
@@ -0,0 +1,300 @@
...
+    task_names = list(set([task["task_name"] for task in task_metadata.values()]))
+    exp_names = [f"batch_size_{input_module}"] + [
+        f"exp_round{rid}_seed{seed}" for rid, seed in enumerate(RANDOM_SEEDS)
+    ]
+    for exp_name in exp_names:
+        target_tasks = ",".join(task_names)
+        override = f'exp_name={exp_name}, run_name=preprocess, target_tasks=\\"{target_tasks}\\"'
+        outputs.append(
+            f'JIANT_CONF="jiant/config/taskmaster/clean_roberta.conf" '
+            f"JIANT_OVERRIDES={override} sbatch ~/jp40.sbatch"
```
sleepinyourhat, line 44: Where is this sbatch file?

scripts/taskmaster/optuna_hp_search/generate_scripts.py
```diff
@@ -0,0 +1,300 @@
...
+        )
+    return outputs
+
+
+def run_batch_size_check(input_module):
```
sleepinyourhat, line 49: Neat!

CONTRIBUTING.md
HaokunLiu: > Did you mention having some tool/trick to run indexing/vocab-building only once per task-model pair?

Previously, if a task has task-specific vocabulary, but there are already vocabulary from previous runs in that experiment. The preprocessing will not create vocabulary for this task, despite it's missing in the saved vocabulary, hence causing an error. We add reload_vocab to force the preprocess always create vocabulary anew. 

However, if we initialize each experiment with a dummy run that just load every task. The saved vocabulary will have everything needed in the following runs. So every run will load correctly, and reload_vocab is no longer needed.

HaokunLiu:
The script is currently in taskmaster branch. Now it directly use DATA_RESULT_DIR and other env variables. sbatch files are also put in scripts/taskmaster/submit_sbatch

PR #1058, opened 2020-04-10
Adding Sentence Order Prediction [WIP]

pruksmhc:
Adding Sentence Order Prediction Task for ALBERT
What this version of MLM supports: ALBERT embedder. 
Below are the runs for MLM + SOP + intermediate task (which is how we intend to use SOP). 

| Task      | Task performance (best ALBERT finetuned on task) |  MLM perplexity | SOP performance|
| ------------- |:-------------:| -------------:|  -------------:| 
| CCG   | 0.944 |5.123   | 0.978 |
| QQP     | 0.8594    | 4.303 |0.8428 |


sleepinyourhat:
That was fast! Taking a look shortly...

jiant/models.py
```diff
@@ -186,7 +187,6 @@ def build_sent_encoder(args, vocab, d_emb, tasks, embedder, cove_layer):
             sep_embs_for_skip=args.sep_embs_for_skip,
             cove_layer=cove_layer,
         )
-        d_sent = 2 * args.d_hid
```
sleepinyourhat, line 189: What's this for?

jiant/models.py
```diff
@@ -541,6 +541,9 @@ def build_task_specific_modules(task, model, d_sent, d_emb, vocab, embedder, arg
             params=task_params,
         )
         setattr(model, "%s_mdl" % task.name, module)
+    elif isinstance(task, SentenceOrderTask):
```
sleepinyourhat, line 544: Is there a reason this isn't a PairClassificationTask? In any case, it probably makes more sense readability-wise to cluster this with the single-purpose task classes below, rather than with the broad classes here. (Just proposing to rearrange the else clauses, not change behavior.)

jiant/models.py
```diff
@@ -687,7 +690,33 @@ def build_single_sentence_module(task, d_inp: int, project_before_pooling: bool,
...
+    # exposed in huggingface for the pretrained weights, we set projection to true.
+    project_before_pooling = True
+    module = build_pair_sentence_module(task, d_inp, model, params, project_before_pooling)
+    # The huggingface implementation exposes the pretrained projection layer for the SOP task, which
+    # we use. See: https://github.com/huggingface/transformers/issues/2671 for more details.
+    module.pooler.project = model.sent_encoder._text_field_embedder.model.pooler
+    return module
+
+
+def build_pair_sentence_module(task, d_inp, model, params, project_before_pooling=None):
```
sleepinyourhat, line 719: `project_before_pooling=None` appears to be new. Why?

sleepinyourhat: Similar questions for the changes elsewhere in this fn.

jiant/models.py
```diff
@@ -687,7 +690,33 @@ def build_single_sentence_module(task, d_inp: int, project_before_pooling: bool,
     return module
 
 
-def build_pair_sentence_module(task, d_inp, model, params):
+def build_sop(task, d_inp, model, params):
+    """
+    Build and load the pretrained head for the sentence order prediction task.
```
sleepinyourhat, line 695: Mention ALBERT support here?

sleepinyourhat:
General sanity check before we run anything big/expensive:
- Make sure that we're getting better-than-chance SOP accuracy out of the box with ALBERT, before any training.

jiant/models.py
pruksmhc: Ah this is an unnecessary change and nothing to do with this PR. I will revert.

jiant/models.py
```diff
@@ -687,7 +690,33 @@ def build_single_sentence_module(task, d_inp: int, project_before_pooling: bool,
...
+    model: MultiTaskModel,
+    params: Params
+
+    Returns
+    -------
+    module: Pooler object, which is loaded with pretrained weights from ALBERT SOP
+    pretraining. 
+
+    """
+    # By default, we do not project before pooling. Since the projection weights is what is
```
sleepinyourhat, line 709: nit: 'weights is'

jiant/models.py
```diff
@@ -1097,6 +1127,7 @@ def _tagger_forward(self, batch: dict, task: TaggingTask, predict: bool) -> dict
         """
         out = {}
         # batch[inputs] only has one item
+
```
pruksmhc, line 1130: Will revert these extra spaces.

jiant/models.py
```diff
@@ -687,7 +690,33 @@ def build_single_sentence_module(task, d_inp: int, project_before_pooling: bool,
...
+    params: Params
+
+    Returns
+    -------
+    module: Pooler object, which is loaded with pretrained weights from ALBERT SOP
+    pretraining. 
+
+    """
+    # By default, we do not project before pooling. Since the projection weights is what is
+    # exposed in huggingface for the pretrained weights, we set projection to true.
```
sleepinyourhat, line 710: Clarify: Is this a projection _before pooling_ , or something else? I'm not aware of any kind of pooling operation in ALBERT.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
         """Get metrics specific to the task"""
         acc = self.scorer1.get_metric(reset)
         return {"accuracy": acc}
+
+
+@register_task("wikipedia_corpus_sop", rel_path="wikipedia_corpus_small")
+class SentenceOrderTask(PairClassificationTask):
+    """ Task class for Sentence Order Prediction (SOP) with wikipedia data.
+        See the ALBERT paper for details on SOP: https://arxiv.org/abs/1909.11942.
+        We are currently using an unpreprocessed version of the Wikipedia corpus
```
sleepinyourhat, line 3741: Say a tiny bit more about the data source here, even if the full docs live elsewhere—there is no one 'Wikipedia corpus'.

jiant/models.py
sleepinyourhat: Ah, I see that this is a PairClassificationTask, but it's the only one that has to preempt the generic build_sentence_pair_module, so it has to be here. This seems like it's making an already messy pile of if-elses a bit harder to follow. I don't see any obvious alternative that doesn't involve a larger refactor, but it'd be good to get @pyeres's input before a merge.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+        """Loading data file and tokenizing the text
+        Args:
+            path: (str) data file path
+        """
+        count = 0
+        with open(path, encoding="utf-8") as txt_fh:
+
+            for row in txt_fh:
+                count += 1
+                if count == 10:
```
sleepinyourhat, line 3770: What's special about 10? Add a comment.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+                    break
+                toks = row.strip()
```
sleepinyourhat, line 3772: This looks fishy...

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+                sentences = row.split(".")
```
sleepinyourhat, line 3773: This looks very fishy. What about other end-of-sentence punctuation like '!'? What does the ALBERT code do?

sleepinyourhat: (We should check on this before running any large experiments, since it could hurt performance in a way that's incremental/hard to detect.)

jiant/tasks/tasks.py
sleepinyourhat: It looks like it's never used...

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+                if len(sentences) <= 1:
+                    continue
+                else:
+                    for i in range(len(sentences) - 1):
+                        if random.uniform(0, 1) > 0.5:
```
sleepinyourhat, line 3778: Genuine question: Is there a more idiomatic way to do this? `random` doesn't have a Bernoulli method built-in. Is randint any faster or more standard?

jiant/models.py
pruksmhc: What they call a "pooler" is a projection: 
https://github.com/huggingface/transformers/blob/ddb6f9476b58ed9bf4433622ca9aa49932929bc0/src/transformers/modeling_albert.py#L446

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+            for row in txt_fh:
+                count += 1
+                if count == 10:
+                    break
+                toks = row.strip()
+                sentences = row.split(".")
+                if len(sentences) <= 1:
+                    continue
+                else:
+                    for i in range(len(sentences) - 1):
```
sleepinyourhat, line 3777: How does this block of code compare with code in HuggingFace/ALBERT?

jiant/models.py
pruksmhc: and this is used here to create the pooled_output (which is the head used in the huggingface code for SOP prediction) https://github.com/huggingface/transformers/blob/ddb6f9476b58ed9bf4433622ca9aa49932929bc0/src/transformers/modeling_albert.py#L570

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+                        )
+                        yield (sent_a_processed, sent_b_processed, is_right_order)
+
+    def load_data(self):
+        pass
+
+    def process_split(
+        self, split, indexers, model_preprocessing_interface
+    ) -> Iterable[Type[Instance]]:
+        """Process a sentence order prediction split by indexing and creating fields.
```
sleepinyourhat, line 3800: Clarify why this needs to override `process_split`, when most pair tasks don't. I assume it has something to do with making sure the label names match what's loaded from pretrained ALBERT? (This kind of information should be helpful for future cleanup of the structure of tasks...)

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+        Args:
+            split: (list) a single list of sentences
+            indexers: (Indexer object) indexer to index input words
+        """
+
+        def _make_instance(sent_):
```
sleepinyourhat, line 3806: I don't have much context for reading this part of the task logic. @W4ngatang? @HaokunLiu?

jiant/models.py
pruksmhc: basically, what the huggingface implementation of ALBERT does is, for the SOP head, takes the output of the ALBERT encoding and passes it through a linear layer and an activation function. The closest thing we have to that is this project_before_pooling attribute in the PairClassifier.

jiant/models.py
pruksmhc: This is because I wanted to be able to set project_before_pooling=True for this experiment (see comment in previous thread about why), but here it was set to False for transformer based models It seems as though you were the one who added this -why was this added? https://github.com/nyu-mll/jiant/blob/master/jiant/models.py#L839

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+                    model_preprocessing_interface.boundary_token_fn(sent_b), indexers
+                )
+                label = LabelField(is_right_order, label_namespace="labels", skip_indexing=True)
+                d = {"input1": inp1, "input2": inp2, "targs": label}
+            return Instance(d)
+
+        for sent in split:
+            yield _make_instance(sent)
+
+    def count_examples(self):
```
sleepinyourhat, line 3827: Same here—not sure what the history/intent is behind when this is/isn't overridden.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,121 @@ def get_metrics(self, reset=False):
...
+        """Computes number of samples
+        Assuming every line is one example.
+        """
+        example_counts = {}
+        for split, split_path in self.files_by_split.items():
+            example_counts[split] = sum(1 for _ in self.get_data_iter(split_path))
+        self.example_counts = example_counts
+
+    def get_split_text(self, split: str):
```
sleepinyourhat, line 3836: Same as above for the next two methods.

jiant/models.py
sleepinyourhat: Ah, got it. This could use a clarification in the code comments here, though it doesn't help that I've lost track of where our pooling logic lives.

jiant/models.py
sleepinyourhat: There's no reason to project (with a linear layer) _before_ pooling with BERT-style models, where pooling is just a first- or last-token selector. This kind of projection only really makes sense with max-pooling.

jiant/tasks/tasks.py
pruksmhc: @phu-pmh

pruksmhc:
I'm reworking this task entirely (basically not making it inherit from Pairclassificationtask) to make the code most cleaner/clearer, so the game plan is to incorporate comments from here. I'm going to start another PR, since most of these comments will be resolved with this change.

jiant/tasks/tasks.py
pruksmhc: Hmm I'm not sure. Taking a look.

jiant/tasks/tasks.py
pruksmhc: SOP isn't implemented in Huggingface.

RobinAlgayres:
Hi, is there a way to use the Sentence Order Prediction currently? From jiant or from huggingface? 
Thank you

jeswan:
@RobinAlgayres This is currently not supported in jiant v2. What tasks/models are you planning to use with Sentence Order Prediction. We can help point you in the right direction.

RobinAlgayres:
Well I would like to use a BERT model pretrained on SOP and use it on a dataset made of pairs of sentences. I do not wish to do finetuning so the model should output a boolean somewhere when I give it a pair of sentences and directly tell me if those sentences are following each other. 
For example, the BertForNextSentencePredictio from huggingface does that but it's trained on NSP.
Do you have any ideas?

PR #1059, opened 2020-04-10
Updating to Transformers v2.6.0

zphang:
Performance comparison on a set of representative tasks.

Task | albert-xxlarge-v2 ([#990](https://github.com/nyu-mll/jiant/pull/990#issuecomment-578271369)) | roberta-large ([#990](https://github.com/nyu-mll/jiant/pull/990#issuecomment-578271369)) | albert-xxlarge-v2 (v2.6.0) | roberta-large (v2.6.0)
---  |  ---  |  ---  |  ---  |  ---
wsc | 0.76 | 0.692 | 0.769 | 0.692
rte | 0.888 | 0.866 | 0.892 | 0.848
wic | 0.74 | 0.724 | 0.741 | 0.721
sst | 0.955 | 0.961 | 0.956 | 0.956
cosmosqa | 0.847 | 0.814 | 0.846 | 0.822
qa-srl | 0.641 | 0.664 | 0.641 | 0.666


environment.yml
```diff
@@ -39,8 +39,9 @@ dependencies:
         - sacremoses
 
         # Warning: jiant currently depends on *both* pytorch_pretrained_bert > 0.6 _and_
```
sleepinyourhat, line 41: Check if this is still true—is the old package still a dependency via Allen? If not, delete.

sleepinyourhat:
Looks good so far! Thanks!

sleepinyourhat:
Since you asked, this still looks good to me. If the testing infrastructure is all ready to go, though, it couldn't hurt to kick off tests with 2.8.0 now, too.

zphang:
Updated the full table. I think we should be good to merge. 

Given the upcoming deadlines, I recommend waiting till after EMNLP to update transformers again.

environment.yml
zphang: allennlp v0.8.4 still depends on pytorch-pretrained-bert, and we've not changed the allennlp requirement.

sleepinyourhat: Dang. I guess it's not worth the effort to update the Allen dependency, assuming that the 2.0 migration is coming up fairly soon.

sleepinyourhat:
@zphang Why delay the merge? If it we've vetted it to our usual degree, than we should get some additional speedups/options out of this PR. 

Of course, it's not great to make major changes after a round of experiments has already started, but the solution to that would just be to maintain a separate branch for each major experiment, which is a good idea in any case.

zphang:
@pyeres I've removed the commit concerting the XLMRoBERTaTokenizer. This PR should only update the requirements (transformers, and tokenizers).

@sleepinyourhat To clarify, I support merging in the update to v2.6.0 now, and putting off the update to v2.8.0.

zphang:
Unrelated: `test_dynamic_masking` failed on the first pass, and passed on rerunning.

PR #1061, opened 2020-04-11
Adding Sentence Order Prediction

pruksmhc:
Adding Sentence Order Prediction Task for ALBERT
What this version of MLM supports: ALBERT embedder.
Below are the runs for MLM + SOP + intermediate task (which is how we intend to use SOP)
The results (on ALBERT) are: 
| Experiment description   | Intermediate Task Performance  | SOP performance|
| ------------- |:-------------:| :-------------:| 
| SST + SOP   | 0.85 |  0.95 |
| QQP  + SOP   | 0.95     | 0.94 |


jiant/models.py
```diff
@@ -687,6 +692,30 @@ def build_single_sentence_module(task, d_inp: int, project_before_pooling: bool,
...
+    Returns
+    -------
+    module: Pooler object, which is loaded with pretrained weights from ALBERT SOP
+    pretraining. 
+
+    """
+    module = SOPClassifier(d_inp, task.n_classes, params)
+    # The huggingface implementation exposes the pretrained projection layer for the SOP task, which
+    # we use. See: https://github.com/huggingface/transformers/issues/2671 for more details.
+    module.pooler.project = model.sent_encoder._text_field_embedder.model.pooler
```
pruksmhc, line 715: Note: ALBERT does do pooling the way we do (pooling type = final): https://github.com/google-research/albert/blob/a41cf11700c1ed2b7beab0a2649817fa52c8d6e1/modeling.py#L261

jiant/models.py
```diff
@@ -541,6 +543,9 @@ def build_task_specific_modules(task, model, d_sent, d_emb, vocab, embedder, arg
             params=task_params,
         )
         setattr(model, "%s_mdl" % task.name, module)
+    elif isinstance(task, SentenceOrderTask):
```
sleepinyourhat, line 546: Like I said before, this should go at the bottom of this if/else block, just for style/readability.

jiant/models.py
sleepinyourhat: I still don't really understand what this is. Are these the final classifier layer parameters (d_hid x 2), or are those loaded somewhere else?

sleepinyourhat: Reading the rest of the PR, it doesn't look like those parameters are loaded anywhere, so this shouldn't work... am I missing something?

jiant/models.py
```diff
@@ -926,7 +957,6 @@ def _single_sentence_forward(self, batch, task, predict):
         logits = classifier(word_embs_in_context, sent_mask)
         out["logits"] = logits
         out["n_exs"] = get_batch_size(batch, self._cuda_device)
-
```
sleepinyourhat, line 929: Intended?

jiant/modules/simple_modules.py
```diff
@@ -1,5 +1,6 @@
 import torch
 import torch.nn as nn
+from jiant.utils.utils import assert_for_log
```
sleepinyourhat, line 3: We've been phasing this out—I don't think it's needed anywhere.

sleepinyourhat: (Just use an assert)

jiant/modules/simple_modules.py
```diff
@@ -94,6 +95,28 @@ def from_params(cls, d_inp, n_classes, params):
...
+    followed by Tanh activation layer, followed by another linear layer.
+    """
+
+    def __init__(self, d_inp, n_classes, params):
+        super(SOPClassifier, self).__init__()
+        self.activation = nn.Tanh()
+        self.pooler = Pooler(d_inp=d_inp, d_proj=d_inp, pool_type=params["pool_type"])
+        assert_for_log(
+            params["cls_type"] == "log_reg",
+            "Please set classifier=log_reg in your config to use" "SOP prediction task",
```
sleepinyourhat, line 110: Odd spacing.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,128 @@ def get_metrics(self, reset=False):
...
+    def get_data_iter(self, path):
+        """Loading data file and tokenizing the text
+        Args:
+            path: (str) data file path
+        """
+        import csv
+
+        f = open(path, "r")
+        reader = csv.reader(f)
+        text = list(reader)
```
sleepinyourhat, line 3783: This looks like a list of file lines, not sentences. Clarify where this gets segmented into sentences.

jiant/models.py
pruksmhc: They are used here: https://github.com/nyu-mll/jiant/blob/master/jiant/modules/simple_modules.py#L34 , which is called in the sop_forward (from the SOP Classifier)

jiant/tasks/tasks.py
pruksmhc: This is segmented in the source text.

sleepinyourhat:
@pruksmhc and I just got off a call, and it seems like there's a pretty deep problem here:
- For whole-input classification tasks, ALBERT passes the first hidden state (CLS) of the last layer through a pretrained Tanh layer before passing it to any task classifier, including the final classifier for SOP. That Tanh layer is not used for other types of tasks, like MLM or SQuAD, where separate outputs are needed for each token position. AFAIK, we don't generally implement or import that extra pretrained Tanh 'pooled_output' layer, so our implementations for MNLI/SST/etc. are incorrect. 
- This should have been handled in #990 (@HaokunLiu), but it seems like nobody caught it. This is understandable—I couldn't find any mention of this setup in the ALBERT paper, but it does seem to be there if you follow the code, and it's referenced in some discussion on a Transformers issue (If you want to chase this down yourself, start here: https://github.com/google-research/albert/search?q=get_pooled_output&unscoped_q=get_pooled_output https://github.com/huggingface/transformers/issues/2671). @HaokunLiu: Could you look into this a bit more fully and make an issue for this pooling layer?
- It'd be best to fix this, but it would probably mean a real non-trivial refactor of our Huggingface interface.
- In the mean time, we should figure out how much this hurts us. @pruksmhc @phu-pmh @HaokunLiu: Do any of you have a minute to do some comparisons _now_ between our ALBERT and the ALBERT paper on some standard classification tasks? We'd have to make sure we're doing test vs. test or dev vs. dev. If it doesn't hurt us much, it may be okay to stick with our current setup for now, as long as we clearly warn users.
- Turning back to this PR, it doesn't look like the HuggingFace version of ALBERT gives us access to the final classifier layer for SOP. This means that we can't use SOP out of the box—it's only usable after some training. This is different from MLM, where we automatically load pretrained weights, so the model can do MLM out of the box. Mind updating this PR to document this very clearly for potential users? 
- Also for this PR, the pretty clear takeaway is that SOP should be implemented the same way as other pair classification tasks. We need a new data loader, but we don't need custom model code. Please refactor.

sleepinyourhat:
Follow-up about this PR: It _may_ make sense to keep this PR more-or-less as is for now. If we don't add the pooler as a universal option, but we can add it here, that may help speed up the rate at which the SOP classifier is learned. The one thing we shouldn't do is initialize a new pooling layer anywhere: The layer is ~4096x4096, and that's too many parameters to try to train from scratch during intermediate training.

pruksmhc:
@Sam  on your update this means keeping it separate for now as a Pairclassificationtask? Even though it is pair classification, right now the head is different enough to other pair classification tasks that even if we have it inherit from paurclassificaitontask we would still have to add in extra logic (as per the last PR I had put up on SOP) that is specific only for this task.

HaokunLiu:
> @pruksmhc and I just got off a call, and it seems like there's a pretty deep problem here:
> 
> * For whole-input classification tasks, ALBERT passes the first hidden state (CLS) of the last layer through a pretrained Tanh layer before passing it to any task classifier, including the final classifier for SOP. That Tanh layer is not used for other types of tasks, like MLM or SQuAD, where separate outputs are needed for each token position. AFAIK, we don't generally implement or import that extra pretrained Tanh 'pooled_output' layer, so our implementations for MNLI/SST/etc. are incorrect.
> * This should have been handled in #990 (@HaokunLiu), but it seems like nobody caught it. This is understandable—I couldn't find any mention of this setup in the ALBERT paper, but it does seem to be there if you follow the code, and it's referenced in some discussion on a Transformers issue (If you want to chase this down yourself, start here: https://github.com/google-research/albert/search?q=get_pooled_output&unscoped_q=get_pooled_output [huggingface/transformers#2671](https://github.com/huggingface/transformers/issues/2671)). @HaokunLiu: Could you look into this a bit more fully and make an issue for this pooling layer?
> * It'd be best to fix this, but it would probably mean a real non-trivial refactor of our Huggingface interface.
> * In the mean time, we should figure out how much this hurts us. @pruksmhc @phu-pmh @HaokunLiu: Do any of you have a minute to do some comparisons _now_ between our ALBERT and the ALBERT paper on some standard classification tasks? We'd have to make sure we're doing test vs. test or dev vs. dev. If it doesn't hurt us much, it may be okay to stick with our current setup for now, as long as we clearly warn users.
> * Turning back to this PR, it doesn't look like the HuggingFace version of ALBERT gives us access to the final classifier layer for SOP. This means that we can't use SOP out of the box—it's only usable after some training. This is different from MLM, where we automatically load pretrained weights, so the model can do MLM out of the box. Mind updating this PR to document this very clearly for potential users?
> * Also for this PR, the pretty clear takeaway is that SOP should be implemented the same way as other pair classification tasks. We need a new data loader, but we don't need custom model code. Please refactor.

1. About the linear layer in pooling. Yes. That seems to be the case. They have this pooling layer on top of the first embedding out of the transformer. But we didn't use it in our implementation.
2. Having a single linear layer pretrained before the classifier seems to be an odd but to some extend understandable design. Maybe they think even this layer can benefit from pretraining.
3. Our results does not fall short against theirs, some times even better. And according to wider discussion from other papers, parameters on the top can not be as easily transferred from pretrain to target task. My guess is, this does not hurt us in any significant way.
4. If possible, I would incline to stay with out current setting.
5. About this SOP PR itself, I agree SOP is not different from sentence pair classification. Pooling layer is not a difference from SOP, but from how ALBERT people think classification should be done.

sleepinyourhat:
@HaokunLiu Clarification Q: What are you referring to with "wider discussion from other papers"?

pruksmhc:
I _can_ revert this task back to Pair classification task, but there will be some cons to that as well. We'll have to add yet another task specific if statement inside of the pair classification forward and build_pair_classification_module functions. @HaokunLiu @sleepinyourhat

HaokunLiu:
> @HaokunLiu Clarification Q: What are you referring to with "wider discussion from other papers"?

"To tune or not to tune" found that the performance of diagnostic classifier at 8 ~ 9 has reach the performance at 12.
Some mainly visualization paper I don't remember the title, found that parameter in the lower layers doesn't change much in the finetuning, while higher layers are where most of the changes happens.


They do not directly suggest adding another pretrained linear layer at the very end will not be beneficial. But it is fairly close to that.

HaokunLiu:
> I _can_ revert this task back to Pair classification task, but there will be some cons to that as well. We'll have to add yet another task specific if statement inside of the pair classification forward and build_pair_classification_module functions. @HaokunLiu @sleepinyourhat

Wait, I don't understand. What will the task-specific if statement do?

pruksmhc:
@HaokunLiu  Well, right now Transformers based models have project_before_pooling=False for all, so we'd have to hard-code it to true for only SOP, and then load the SOP linear layer weights from huggingface to  the pooler, which is currently not being done for the rest of the Sentence classification tasks (and I don't think we have bandwidth right now to run _all_ the relevant test experiments to do that).

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,125 @@ def get_metrics(self, reset=False):
         """Get metrics specific to the task"""
         acc = self.scorer1.get_metric(reset)
         return {"accuracy": acc}
+
+
+@register_task("wikipedia_corpus_sop", rel_path="wikipedia_corpus_small")
+class SentenceOrderTask(Task):
+    """ Task class for Sentence Order Prediction (SOP) with Wikipedia data (more specifically,
+        the dump from Wikimedia). See the ALBERT paper for details on SOP: https://arxiv.org/abs/1909.11942.
+        We are currently using an unpreprocessed version of the Wikipedia corpus
```
sleepinyourhat, line 3741: Please fix the comment on this from the earlier PR>

jiant/tasks/tasks.py
sleepinyourhat: Make it clear how to prepare the source text, then. Also, make sure it's a separate file from the MLM source—MLM requires full paragraphs, while SOP requires sentences.

sleepinyourhat:
> I can revert this task back to Pair classification task, but there will be some cons to that as well. We'll have to add yet another task specific if statement inside of the pair classification forward and build_pair_classification_module functions. @HaokunLiu @sleepinyourhat

We should revert the task class to inherit from a the normal pair classification task iff we use the same modeling code that we do for other pair classification tasks. I don't have strong feelings either way, but I think those two decisions should go together.

sleepinyourhat:
There are a bunch of comments on the old PR that are still valid here, BTW. Please go through those again too before asking for review next.

sleepinyourhat:
> "To tune or not to tune" found that the performance of diagnostic classifier at 8 ~ 9 has reach the performance at 12.
That's useful precedent. @HaokunLiu Could you add a link to this in the issue for future reference?

pruksmhc:
Alright, let's just stick with the current implementation, since there are differences between SOP and other PairClassificationTasks as of this moment.

sleepinyourhat:
Roger. It is arguably a weird inconsistency, so we should only do it (use the current impl.) if we're pretty sure it'll help rather than hurt. Probably a good idea to check in with the group.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,132 @@ def get_metrics(self, reset=False):
...
+
+    def get_data_iter(self, path):
+        """Loading data file and tokenizing the text
+        Args:
+            path: (str) data file path
+        """
+        f = open(path, "r")
+        text = [line.strip() for line in f]
+        for i in range(len(text) - 1):
+            if random.uniform(0, 1) > 0.5:
```
pruksmhc, line 3789: On your point from the last PR, I'm not sure if there's a significant difference between random.randint and random.uniform. My bet is that it won't significantly decrease speed in preprocessing.

pruksmhc:
@sleepinyourhat I went through the last PR and think I have all the comments addressed. A lot of the comments were on why we override the functions for the SOP task here I think is now pretty obvious, because it is because this SOP task inherits from Task and a lot of the functions are not implemented in the Task class.

pruksmhc:
@sleepinyourhat I went through the last PR and think I have all the comments addressed. A lot of the comments were on why we override the functions for the SOP task here I think is now pretty obvious, because it is because this SOP task inherits from Task and a lot of the functions are not implemented in the Task class.

pruksmhc:
We should have the validation runs tomorrow

jiant/models.py
```diff
@@ -687,6 +692,30 @@ def build_single_sentence_module(task, d_inp: int, project_before_pooling: bool,
     return module
 
 
+def build_sop(task, d_inp, model, params, args):
+    """
+    Build and load the pretrained head for the sentence order prediction task.
+    Right now, there is only support for ALBERT.
```
sleepinyourhat, line 698: We should add an assert or warning for this. Do we expect reasonable (non-standard) behavior with other pretrained models?

jiant/models.py
```diff
@@ -1009,6 +1040,21 @@ def _span_prediction_forward(self, batch, task, predict):
             out["preds"] = {"span_start": pred_span_start, "span_end": pred_span_end}
         return out
 
+    def _sop_forward(self, batch, task, predict):
```
sleepinyourhat, line 1043: Do we actually need this? Add a comment to clarify what happens here that isn't already done by _pair_sentence_forward. (The model structure is different, but it seems like that difference lives entirely inside self._get_classifier, right?)

jiant/modules/simple_modules.py
```diff
@@ -94,6 +94,26 @@ def from_params(cls, d_inp, n_classes, params):
...
+    Task head for sentence order prediction task. We approximate the pooled output from ALBERT
+    via a linear layer followed by Tanh activation layer, which is then fed into the
+    classification linear layer.
+    """
+
+    def __init__(self, d_inp, n_classes, params):
+        super(SOPClassifier, self).__init__()
+        self.activation = nn.Tanh()
+        self.pooler = Pooler(d_inp=d_inp, d_proj=d_inp, pool_type=params["pool_type"])
+        assert params["cls_type"] == "log_reg"
```
sleepinyourhat, line 108: Add an informative error here.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,132 @@ def get_metrics(self, reset=False):
...
         acc = self.scorer1.get_metric(reset)
         return {"accuracy": acc}
+
+
+@register_task("wikipedia_corpus_sop", rel_path="wikipedia_corpus_small")
+class SentenceOrderTask(Task):
+    """ Task class for Sentence Order Prediction (SOP). See the ALBERT paper for details on SOP:
+        https://arxiv.org/abs/1909.11942.
+        We are currently using an unpreprocessed version of the Wikipedia corpus
+        (more specifically, the Wikidump data) that consists of 5% of the data. You can generate
```
sleepinyourhat, line 3742: Question: Where do we specify the specific dump that we're downloading? Does the NVidia code always download the latest dump? If so, that's a problem for reproducibility.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,132 @@ def get_metrics(self, reset=False):
...
+        the data by following the instructions from jiant/scripts/mlm.
+        There are several key things to note about our SOP ALBERT implementation.
+            1. We do not load the pretrained weights for the SOP head beacuse they are
+            unavailable in Huggingface. We only use the pretrained weights of the linear layer
+            from ALBERT that creates the pooled output used in SOP.
+            2. jiant supports a multitask version of the ALBERT pretraining regime, which requires
+            at each training step for each piece of text running that text through MLM and SOP head
+            at the same time.
```
sleepinyourhat, line 3750: How? I thought you said that we decided *not* to support running both heads at the same time.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,132 @@ def get_metrics(self, reset=False):
...
+        _, preds = logits.max(dim=1)
+        self.acc_scorer(preds, labels)
+
+    def get_data_iter(self, path):
+        """Loading data file and tokenizing the text
+        Args:
+            path: (str) data file path
+        """
+        f = open(path, "r")
+        text = [line.strip() for line in f]
```
sleepinyourhat, line 3787: Asking again (!!!): Where does this get split into sentences? SOP needs sentences. MLM needs paragraphs or longer.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,132 @@ def get_metrics(self, reset=False):
...
+            )
+            sent_b_processed = tokenize_and_truncate(
+                self._tokenizer_name, sent_b, self.max_seq_len // 2
+            )
+            yield (sent_a_processed, sent_b_processed, is_right_order)
+
+    def load_data(self):
+        pass
+
+    def process_split(
```
sleepinyourhat, line 3808: The methods below still need a look from someone else.

jiant/tasks/tasks.py
phu-pmh: We actually use the same sentence level data for MLM. We've tried with both sentence level and paragraph level for MLM and we didn't see noticable difference in performance.

pruksmhc:
Alright, I'll have it inherit from the PairClassificationTask, since the main differences between it and other pairclassification tasks now is in its classifier.

jiant/tasks/tasks.py
pruksmhc: We don't. We support a "multitask version," so _not_ the original version in the ALBERT implementation. Maybe that's not necessary to put in here, just in the paper.

jiant/tasks/tasks.py
pruksmhc: The file here takes in sentences. I'll clarify.

jiant/models.py
pruksmhc: We do load the weights from the pooling layer which only makes sense for ALBERT, but there shouldn't be anything here that breaks for any other model. I'm hesitant to say we support other models before we run experiments with other models, so I'll just add an assert.

jiant/models.py
pruksmhc: I merged it into pair sentence forward.

jiant/tasks/tasks.py
pruksmhc: Seems like it downloads from here: https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2. @phu-pmh  do you know if that's static or if it's overwritten with the latest dump?

pruksmhc: Seems like it does keep changing. Phu is working on a modified script that reads a dump from a particular date.

phu-pmh: Yes, seems like NVIDIA code always download the latest version. The dumb version we used is `https://dumps.wikimedia.org/enwiki/20200301/`. We can edit their code to download the same version as ours but it seems that there is copyright issues. 
According to @pyeres "NVIDIA files are Apache 2.0 licensed, and (depending on whether we’re redistributing or modifying+redistributing, we’d need to add some notices"

jiant/tasks/tasks.py
pruksmhc: @zphang mind taking a look?

jiant/tasks/tasks.py
pruksmhc: @sleepinyourhat

sleepinyourhat: If the script is easy to edit, one option would be to give instructions here for how to edit it to specify a version.

zphang:
Following up on my note on multiple sentences, here's my read of happens in BERT/ALBERT. 

1. BERT and ALBERT seem to share the same/similar preprocessing code. [BERT](https://github.com/google-research/albert/blob/a41cf11700c1ed2b7beab0a2649817fa52c8d6e1/create_pretraining_data.py#L267-L387. [ALBERT](https://github.com/google-research/bert/blob/eedf5716ce1268e56f0a50264a88cafad334ac61/create_pretraining_data.py#L223-L335)
2. The original input files consist of one sentence per line, and an empty line to delineate a document boundary. (This prevents taking the last sentence of one document and the first sentence of the next document as a valid pair).
3. However, in constructing the training instances, they actually do take multiple consecutive sentences as part of "sentence a" or "sentence b". Let's call these SequenceA and SequenceB for clarity. The logic is surprisingly complex, but basically they do load multiple sentences for both SequenceA and Sequence B. See the above links to BERT and ALBERT code.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,113 @@ def get_metrics(self, reset=False):
...
+        the step of reading in the data for SOP is different than other
+        PairClassificationTasks.
+
+        Args:
+            path: (str) data file path
+        """
+        f = open(path, "r")
+        #  The dataset comes with one sentence per line, thus we split by
+        #  line here.
+        text = [line.strip() for line in f]
```
zphang, line 3775: This looks like it loads the whole files into memory. Since we're ultimately yielding, it would be better to use a generator here as well. E.g.:

```python
prev_sent = None
this_sent = None
for line in f:
    prev_sent = this_sent
    this_sent = line.strip()
    if prev_sent is None:
        continue
    is_right_order = random.uniform(0, 1) > 0.5
    if is_right_order:
        sent_a, sent_b = prev_sent, this_sent
    else:
        sent_a, sent_b = this_sent, prev_sent
    sent_a_processed = ...
    sent_b_processed = ...
    yield ...
```

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,113 @@ def get_metrics(self, reset=False):
...
+        """Process a sentence order prediction split by indexing and creating fields.
+        We override the PairClassificationTask process_split because our data split
+        is different from the typical PairClassificationTask due to the more memory-efficient
+        generator way of loading data we employ for SOP due to the dataset size.
+        Args:
+            split: (list) a single list of sentences
+            indexers: (Indexer object) indexer to index input words
+        """
+
+        def _make_instance(sent_):
```
zphang, line 3808: Use a different variable name than `sent`/`sent_`, since it's a tuple of two sentences and a boolean.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,113 @@ def get_metrics(self, reset=False):
...
+    def get_sentences(self) -> Iterable[Sequence[str]]:
+        """Yield sentences, used to compute vocabulary.
+        """
+        for split in self.files_by_split:
+            # Don't use test set for vocab building.
+            if split.startswith("test"):
+                continue
+            for sent in self.get_data_iter(self.files_by_split[split]):
+                yield sent[0]
+                yield sent[1]
```
zphang, line 3835: If I'm reading this correctly, use something like
```python
for sent_a, sent_b, is_right_order in self.get_data_iter(self.files_by_split[split]):
    yield sent_a
    yield_sent_b
```
(Also it looks like almost every sentences will be yielded twice, right? This would increase memory/storage. Arguably, for tasks like MLM and NSP/SOP, we don't necessarily care about making full use of every sentence since raw text is so plentiful.)

jiant/tasks/tasks.py
zphang: If we now the full length of the dataset (i.e. if we don't use an iterator as suggested below), it should be faster request a whole array of random booleans once, than per iteration.
In either case, my sense is that the RNG time will be minor compared to the `tokenize_and_truncate` function calls each iteration.

zphang:
Besides the NSP/SOP sentences issue above, and the in-memory list of lines, I don't see any areas of concern.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+
+    def get_target_seq_length(self):
+        target_is_max = random.random() > 0.1
+        if target_is_max:
+            target_seq_length = self.max_seq_len
+        else:
+            target_seq_length = random.randint(2, self.max_seq_len)
+        return target_seq_length
+
+    def get_data_iter(self, path):
```
pruksmhc, line 3771: @zphang  @sleepinyourhat  would like your input on this function. I'm in the middle of writing tests right now, but if you see a way to make this clearer/less riddled with nested if/else statements let me know!

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+            "test": os.path.join(path, "test.txt"),
+        }
+        self._label_namespace = self.name + "_labels"
+
+    def get_target_seq_length(self):
+        target_is_max = random.random() > 0.1
+        if target_is_max:
+            target_seq_length = self.max_seq_len
+        else:
+            target_seq_length = random.randint(2, self.max_seq_len)
```
zphang, line 3768: Worth double checking `max_seq_len` includes adjustment for the additional tokens (CLS/SEP).

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+                -50% of the time, the first and second segments are switched. 
+            There is also some logic here to make sure that we always split by the sentence instead of 
+            in the middle of a sentence.
+        Args:
+            path: (str) data file path
+        """
+        f = open(path, "r")
+        #  The dataset comes with one sentence per line, thus we split by
+        #  line here.
+        current_chunk = [tokenize_and_truncate(self._tokenizer_name, next(f), -1)]
```
zphang, line 3792: Worth double checking the usage of `tokenize_and_truncate`. It looks like `tokenize_and_truncate(..., ..., -1)` may be problematic. From the function:

```python
...
max_seq_len -= 2
...
tokenizer.tokenize(sent)[:max_seq_len]
```

Which will chop off tokens at the end.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+            segment = tokenize_and_truncate(self._tokenizer_name, segment, -1)
+            if "END OF ARTICLE" in segment or current_length >= target_seq_length:
+                for_next_chunk = []
+                if "END OF ARTICLE" not in segment:
+                    for_next_chunk.append(segment)
+                if current_length > target_seq_length:
+                    # Since the most current sentence added to the chunk exceeds the target
+                    # length, we save it for the next chunk (next example).
+                    for_next_chunk.append(current_chunk[-1])
+                    current_chunk.pop(-1)
```
zphang, line 3807: Combine to 
```
for_next_chunk.append(current_chunk.pop())
```

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+                    for j in range(a_end):
+                        tokens_a.extend(current_chunk[j])
+                    tokens_b = []
+                    for j in range(a_end, len(current_chunk)):
+                        tokens_b.extend(current_chunk[j])
+                    if random.random() < 0.5:
+                        in_order = 0
+                        tokens_a, tokens_b = tokens_b, tokens_a
+                    yield (tokens_a, tokens_b, in_order)
+                    in_order = 1
```
zphang, line 3825: This is confusing.
```python
in_order = random.random() < 0.5
if in_order:
    yield (tokens_a, tokens_b, in_order)
else:
    yield (tokens_b, tokens_a, in_order)
```

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+                    if len(for_next_chunk) > 0:
+                        current_chunk = for_next_chunk
+                    else:
+                        # We find the next sentence for the next example.
+                        try:  # Might run into StopIterationError
+                            current_chunk = [next(f)]
+                        except:
```
zphang, line 3832: `except StopIterationError:`

jiant/modules/simple_modules.py
```diff
@@ -94,6 +94,30 @@ def from_params(cls, d_inp, n_classes, params):
         )
 
 
+class SOPClassifier(nn.Module):
+    """
+    Task head for sentence order prediction task. We approximate the pooled output from ALBERT
```
sleepinyourhat, line 99: approximate => implement 
The data and checkpoint loaders aren't quite correct, but this is the exact right model, no?

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+
+
+@register_task("wikipedia_corpus_sop", rel_path="wikipedia_corpus_small")
+class SentenceOrderTask(PairClassificationTask):
+    """ Task class for Sentence Order Prediction (SOP). See the ALBERT paper for details on SOP:
+        https://arxiv.org/abs/1909.11942.
+        We are currently using an unpreprocessed version of the Wikipedia corpus
+        (more specifically, the Wikidump data) that consists of 5% of the data. You can generate
+        the data by following the instructions from jiant/scripts/mlm.
+        One thing to note about our SOP ALBERT implementationn is that we do not load the pretrained
```
sleepinyourhat, line 3744: typo

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+            segment = next(f)
+            segment = tokenize_and_truncate(self._tokenizer_name, segment, -1)
+            if "END OF ARTICLE" in segment or current_length >= target_seq_length:
+                for_next_chunk = []
+                if "END OF ARTICLE" not in segment:
+                    for_next_chunk.append(segment)
+                if current_length > target_seq_length:
+                    # Since the most current sentence added to the chunk exceeds the target
+                    # length, we save it for the next chunk (next example).
+                    for_next_chunk.append(current_chunk[-1])
```
zphang, line 3806: Should be done in the reverse order. Put the part that's too long from the end of `current_chunks` into `for_next_chunk`, then add the current `segment` to `for_next_chunk`.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+
+        ALBERT does SOP classification by, for each document:
+            For each example:
+                -90% of the time, getting sentences of max_seq_length number of tokens, and
+                10% of the time, getting sentences of a random number of tokens between 2 and max_seq_length.
+                -Given the sampled sentences, randomly sample N such that the first N sentences in the 
+                sampled go to the first segment, and the rest go to the second. 
+                -50% of the time, the first and second segments are switched. 
+            There is also some logic here to make sure that we always split by the sentence instead of 
+            in the middle of a sentence.
```
sleepinyourhat, line 3785: If you're going to give this level of detail, spell out how this works. How can you sample sentences of 2 tokens if there are few/no such sentences in the corpus, for example? Do you ever truncate?

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+                    # length, we save it for the next chunk (next example).
+                    for_next_chunk.append(current_chunk[-1])
+                    current_chunk.pop(-1)
+                target_seq_length = self.get_target_seq_length()
+                if len(current_chunk) >= 2:
+                    # Make sure we have at least 2 sentences to distribute between the two
+                    # segments.
+                    a_end = 1
+                    if len(current_chunk) > 2:
+                        a_end = random.randint(1, len(current_chunk) - 1)
```
zphang, line 3814:
```python
a_end = random.randint(1, len(current_chunk) - 1)
```` should suffice, right?

We know from the above that `len(current_chunk) >= 2`. If `len(current_chunk)==2`, we get `random.randint(1, 1)` which is always 1.

File: jiant/tasks/tasks.py [hunk 30]
```diff
@@ -3731,3 +3732,160 @@ def get_metrics(self, reset=False):
...
+                for_next_chunk = []
+                if "END OF ARTICLE" not in segment:
+                    for_next_chunk.append(segment)
+                if current_length > target_seq_length:
+                    # Since the most current sentence added to the chunk exceeds the target
+                    # length, we save it for the next chunk (next example).
+                    for_next_chunk.append(current_chunk[-1])
+                    current_chunk.pop(-1)
+                target_seq_length = self.get_target_seq_length()
+                if len(current_chunk) >= 2:
```
zphang, line 3809: What happens when `len(current_chunk)==1`?

jiant/tasks/tasks.py
phu-pmh: I'll update this in MLM branch's preprocessing script `jiant/scripts/mlm`.

jiant/tasks/tasks.py
phu-pmh: Fixed. This happens when the next sentence is "END OF ARTICLE".  We simply ignore that example in that case.

jiant/tasks/tasks.py
phu-pmh: Looks ok to me now. Jason left a few comments and I've fixed it according to his comments.

jiant/tasks/tasks.py
phu-pmh: We truncate based on `target_seq_length`, which is defined in `get_target_seq_length()`.

sleepinyourhat: I'm asking for this to be clarified in the comment—I still don't understand the code comment.

jiant/tasks/tasks.py
sleepinyourhat: @zphang Mind taking another look? I don't fully follow this.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,165 @@ def get_metrics(self, reset=False):
...
+            tokenizer = get_tokenizer(tokenizer_name)
+            return tokenizer.tokenize(sent)[:]
+            
+        f = open(path, "r")
+        #  The dataset comes with one sentence per line, thus we split by
+        #  line here.
+        current_chunk = [_tokenize(self._tokenizer_name, next(f))]
+        current_length = 0
+        target_seq_length = self.get_target_seq_length()
+        in_order = 1
```
sleepinyourhat, line 3800: Unused?

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,165 @@ def get_metrics(self, reset=False):
...
+        """
+        def _tokenize(tokenizer_name, sent):
+            tokenizer = get_tokenizer(tokenizer_name)
+            return tokenizer.tokenize(sent)[:]
+            
+        f = open(path, "r")
+        #  The dataset comes with one sentence per line, thus we split by
+        #  line here.
+        current_chunk = [_tokenize(self._tokenizer_name, next(f))]
+        current_length = 0
```
sleepinyourhat, line 3798: This never gets reset or reduced later on. Is that right?

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,167 @@ def get_metrics(self, reset=False):
...
+                    in_order = random.random() < 0.5
+                    if in_order:
+                        yield (tokens_a, tokens_b, in_order)
+                    else:
+                        yield (tokens_b, tokens_a, in_order)
+                # if len(current_chunk) >=2, we will yield and reinitialize
+                # if len(current_chunk) ==1, we will not yeild, and reinitialize
+                if len(for_next_chunk) > 0:
+                    current_chunk = for_next_chunk
+                    current_length
```
zphang, line 3832: Is this line incomplete?

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,167 @@ def get_metrics(self, reset=False):
...
+                sampled go to the first segment, and the rest go to the second. 
+                -50% of the time, the first and second segments are switched. 
+            There is also some logic here to make sure that we always split by the sentence instead of 
+            in the middle of a sentence.
+        Args:
+            path: (str) data file path
+        """
+        def _tokenize(tokenizer_name, sent):
+            tokenizer = get_tokenizer(tokenizer_name)
+            return tokenizer.tokenize(sent)[:]
```
zphang, line 3792: What is the purpose of the `[:]` here?

jiant/tasks/tasks.py
pruksmhc: @phu-pmh  would be great if you updated this in the MLM branch.

pyeres:
Hi @phu-pmh & @pruksmhc — I see this is a "[WIP]". What remains to be done or reviewed?

phu-pmh:
> Hi @phu-pmh & @pruksmhc — I see this is a "[WIP]". What remains to be done or reviewed?

I just pushed the instructions for data generation. Other than that, nothing left to be done or reviewed, I think.

jiant/tasks/tasks.py
sleepinyourhat: This seems like a contradiction—if it's unpreprocessed, why do we have a data preparation script.

tests/tasks/test_sop.py
```diff
@@ -0,0 +1,39 @@
...
+class TestSOP(unittest.TestCase):
+    def setUp(self):
+        cls, _, kw = REGISTRY["wikipedia_corpus_sop"]
+        self.temp_dir = tempfile.mkdtemp()
+        self.max_seq_len = 24
+        self.SOPTask = cls(
+            os.path.join("wikipedia_corpus_sop"),
+            max_seq_len=self.max_seq_len,
+            name="wikipedia_corpus_sop",
+            tokenizer_name="roberta-large",
```
sleepinyourhat, line 16: Will this cause the model to download RoBERTa unnecessarily when running tests?

pruksmhc: No it won't since it only sets the self._tokenizer_name variable to be used later.

jiant/tasks/tasks.py
sleepinyourhat: Also, like we talked about, it should be clear exactly what dataset you get when you run these scripts. Make it clear somewhere (maybe in the scripts readme) whether you're getting a specific dump, or just whatever is most recent.

phu-pmh: I've updated it.

sleepinyourhat:
Okay—my major concerns have been addressed, though I'd still prefer Phil do the final review.

pruksmhc:
@pyeres  if you could take a look that would be great.

pyeres:
> Okay—my major concerns have been addressed, though I'd still prefer Phil do the final review.

Per conversation with @sleepinyourhat — my review will be focused on `SentenceOrderTask`.

scripts/sop/README.md
```diff
@@ -0,0 +1,30 @@
+# Downloading Wikipedia Corpus
+
+We use the preprocessing code from https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/LanguageModeling/BERT#getting-the-data
+and the bash scripts provided here is used to help with streamlining the data generation in the NVIDIA repository.
+
+First, git clone https://github.com/NVIDIA/DeepLearningExamples.git.
+Then, move create_wiki_data.sh and get_small_english_wiki.sh into DeepLearningExamples/PyTorch/LanguageModeling/BERT/data.
```
pyeres, line 7: I can't find `create_wiki_data.sh`, I'm guessing this is meant to refer to `scripts/sop/create_wiki_sop_data.sh`.

scripts/sop/README.md
```diff
@@ -0,0 +1,30 @@
...
+In `DeepLearningExamples/PyTorch/LanguageModeling/BERT/data/TextSharding.py`, replace line 55:
+`self.articles[global_article_count] = line.rstrip()` with `self.articles[global_article_count] = line.rstrip() + "\n ========THIS IS THE END OF ARTICLE.========"`.
+This is because SOP requires a signal for the end of each Wikipedia article.
+
+Run `bash create_wiki_sop_data.sh $lang $save_directory`
+The NVIDIA code supports English (en) and Chinese (zh) wikipedia.
+
+For example, to download and process English Wikipedia and save it in `~/Download` directory, run
+`bash create_wiki_sop_data.sh en ~/Download`
+
```
pyeres, line 25: Heads up — there are some steps that I needed to take that I don't think are documented here. Here are my notes from my debugging process (they're rough but maybe they'll help with updating these instructions):

> `create_wiki_sop_data.sh` takes cmd line args, `create_wiki_sop_data.sh` exports the second arg as `BERT_PREP_WORKING_DIR`, `create_wiki_sop_data.sh` checks if dir `wikiextractor` exists and if not clones it. Later, `create_wiki_sop_data.sh` calls `bertPrep.py`, and `bertPrep.py` (called with default `args.skip_wikiextractor == 0`) expects `WikiExtractor.py` to be at `/workspace/wikiextractor/WikiExtractor.py`, but `create_wiki_sop_data.sh` didn't put it under `/workspace/`.

So, after that I modified `bertPrep.py` at L80 to point to where `create_wiki_sop_data.sh` cloned `wikiextractor`.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,171 @@ def get_metrics(self, reset=False):
...
+        return target_seq_length
+
+    def get_data_iter(self, path):
+        """Loading data file and tokenizing the text. We override the
+        this function and all functions that call this function because
+        the step of reading in the data for SOP is different than other
+        PairClassificationTasks.
+
+        ALBERT does SOP classification by, for each document:
+            For each example, we first fetch target_seq_length number of sentences from the dcoument:
```
pyeres, line 3779: I'm having trouble matching this description to what's happening in the code:

This line says we're getting "target_seq_length number of sentences from the dcoument".

At a few points later in the code `target_seq_length` is compared to token (not sentence) counts.

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,171 @@ def get_metrics(self, reset=False):
...
+                -90% of the time, this target_seq_length is equal to max_seq_length,  and
```
pyeres, line 3780: Can you clarify where this 90% logic comes from? I tried to find it by skimming the [article cited in the docstring](https://arxiv.org/abs/1909.11942).

jiant/tasks/tasks.py
```diff
@@ -3731,3 +3732,171 @@ def get_metrics(self, reset=False):
...
+                        current_chunk = [_tokenize(self._tokenizer_name, next(f))]
+                        current_length = len(current_chunk[0])
+                    except:
+                        print("Done loading data for SOP")
+                        current_chunk = []
+                        current_length = 0
+                        pass
+            else:
+                current_chunk.append(segment)
+                current_length += len(segment)
```
pyeres, line 3849: I've been trying to understand this logic, and I'm struggling to understand the corner cases.

Based on the description in the docstring I tried to replicate the logic as I understand it. This is what I came up with:

```
        tokenizer = get_tokenizer(self._tokenizer_name)
        with open(path, "r") as f:
            document = []
            for line in f:
                line = line.strip()
                if "END OF ARTICLE" not in line:
                    tokenized_line = tokenizer.tokenize(line)
                    document.append(tokenized_line)
                else:
                    if len(document) < 2:
                        continue
                    for first_token_seq, second_token_seq in zip(document, document[1:]):
                        yield first_token_seq, second_token_seq, True
                        yield second_token_seq, first_token_seq, False
                    document = []
```

This doesn't implement the "90%" logic I asked about [above](https://github.com/nyu-mll/jiant/pull/1061/files#r409929289), but it passes the test provided in your PR.

Can you help me understand what I'm missing that's implemented in L3789-L3849?

scripts/sop/README.md
pruksmhc: will make that change

jiant/tasks/tasks.py
pyeres: OK, since last night this has come into focus — this code is mostly ported [from TensorFlow](https://github.com/google-research/albert/blob/4135001fb1a84491bb86bebb2646523495d36f74/create_pretraining_data.py#L273-L350).

Instead of copying this over and rewriting components, please ~either import (probably not worth the overhead of making this a dependency), or~ directly copy the TensorFlow implementation into a utility file (and provide an attribution/license/modification notice for the copied code), then make the minimal set of changes to that encapsulated code to make compatible with your task (commenting your modifications).

pruksmhc: There's a few things incorrect. First, with 50% chance, you yield first_token_seq, second_token_seq, True and 50% chance you yield second_token_seq, first_token_seq, False. Second, you need to get, at each example, as many sentences as you can and still be under the target_seq_length limit (which 90% of the time is max_seq_len and 10% of the time is a random number between 2 and max_seq_len).

pyeres: Sorry, I still don't understand — I need to set up time with you to discuss. I'll ping you.

pyeres:
Capturing some notes from conversation with @pruksmhc out-of-band:
* There are intentional differences between our implementation of SOP and the TF implementation — inline code comments will be added to call out these differences.
* The docstrings will be updated to fully outline the version of the SOP implemented here.
* @pruksmhc considers the SOP implementation to have been reviewed for correctness (already, by other reviewers).

@phu-pmh, @sleepinyourhat, @zphang for visibility as contributors/reviewers on this PR — I'll plan to resume review once the docstrings and comments (mentioned above) are added.

sleepinyourhat:
@pruksmhc @phu-pmh I talked through this PR with Phil earlier today, and I think it'll be ready to go with two more smallish fixes:

- Fix the docstring issue mentioned here: https://github.com/nyu-mll/jiant/pull/1061#discussion_r409929236
- Fix the credit assignment issue for the sections of code that use snippets of ALBERT code.

Try to get to these before you do anything else non-emergency on jiant, and we can get this merged in. Thanks!

scripts/sop/README.md
pruksmhc: Included.

pyeres:
> Below are the runs for MLM + SOP + intermediate task (which is how we intend to use SOP)

@pruksmhc — I'm not able to locate the run results you mentioned in the PR description (maybe they're buried somewhere else in the thread?). Please repost them in the PR description.

PR #1064, opened 2020-04-12
Increasing Length for each Example in MLM

pruksmhc:
Previously, we had done sentence-level MLM (https://github.com/nyu-mll/jiant/pull/1030). In this PR, we modify each example to be longer sequences.


jiant/tasks/lm.py
```diff
@@ -246,16 +246,25 @@ def update_metrics(self, out, batch=None):
         return
 
     def get_data_iter(self, path):
-        """Loading data file and tokenizing the text
+        """
+        Loading data file and tokenizing the text. We treat the Wikipedia corpus as a
+        long sequence, and we take each slice of 510 tokens as an example. The dataset
```
sleepinyourhat, line 251: Is it always 510, or does it depend on the specified max seq length?

pruksmhc: Ah.

jiant/tasks/lm.py
```diff
@@ -246,16 +246,25 @@ def update_metrics(self, out, batch=None):
...
         """
+        seq_len = self.max_seq_len - 2
+        total_tokens = []
         with open(path, "r", encoding="utf-8") as txt_fh:
             for row in txt_fh:
                 toks = row.strip()
                 if not toks:
                     continue
-                yield tokenize_and_truncate(self._tokenizer_name, toks, self.max_seq_len)
+                total_tokens += toks
```
pyeres, line 264: This change has me confused* (so I kicked off runs to compare the behavior of this code pre/post change):
* pre-change `tokenize_and_truncate()` was called and `sent_` (at L270) was a list of word-level tokens.
* post-change `tokenize_and_truncate()` is not called, and instead `total_tokens += toks` is performed, and this results in a list of characters (and later on `sent_` (at L279) is also list of characters).

Is the move from word- to character-level tokens intentional?

\* please let me know if there's other context I'm overlooking that would help explain this change.

pruksmhc: You are completely correct. I caught this before we ran the full set of K sweeps yesterday, but we were head over heels with other things so didn't update this PR yesterday.

pruksmhc: Let me push now what we're currently using.

jiant/tasks/lm.py
```diff
@@ -246,16 +246,26 @@ def update_metrics(self, out, batch=None):
...
         """
+        seq_len = self.max_seq_len - 2
+        total_tokens = []
         with open(path, "r", encoding="utf-8") as txt_fh:
             for row in txt_fh:
                 toks = row.strip()
                 if not toks:
                     continue
-                yield tokenize_and_truncate(self._tokenizer_name, toks, self.max_seq_len)
+                toks = tokenize_and_truncate(self._tokenizer_name, toks, -1)
```
pyeres, line 264: Why drop the last 3 tokens from every line?

jiant/tasks/lm.py
```diff
@@ -246,16 +246,26 @@ def update_metrics(self, out, batch=None):
...
+                total_tokens += toks
```
pyeres, line 265: Are the datasets small enough that loading them into memory (`total_tokens += toks`) isn't a memory issue?

pruksmhc: It hasn't been an issue so far.

jiant/tasks/lm.py
pruksmhc: Good point, that's not supposed to happen.

pruksmhc:
base_roberta.conf should not be added- - I removed it.

jiant/tasks/lm.py
pyeres: Please update to something like the below to ease memory load:

```
                token_buffer += toks
                if len(token_buffer) > seq_len:
                    token_sequence = token_buffer[:seq_len]
                    token_buffer = token_buffer[seq_len:]
                    yield token_sequence
```

(`token_buffer` replaces your `total_tokens`, otherwise you should be able to drop this in, replacing L266-269)

pruksmhc: Done.

Issue #180, opened 2018-07-17
training_data_fraction is slow.


sleepinyourhat:
If you're using a small fraction (10% or less), you spend most of your time hashing examples, GPU usage drops, and samples per second drops dramatically. The lazy solution for now is to use a cumbersome two different setups: training_data_fraction when training on >1% of the training data, and a custom data file and new task when training on less.

zphang:
*This is an automatically generated comment.*

As we update `jiant` to v2.x, `jiant` v1.x has been migrated to https://github.com/nyu-mll/jiant-v1-legacy. As such, we are closing all issues relating to `jiant` v1.x in this repository.

If this issue is still affecting you in `jiant` v1.x, please follow up at https://github.com/nyu-mll/jiant-v1-legacy/issues/180.

If this issue is still affecting you in `jiant` v2.x, reopen this issue or create a new one.

PR #1069, opened 2020-04-16
Faster data fraction

HaokunLiu:
#180 
This add a one-time cost before training, but the code should be more efficient during training.

There isn't a test case for read_record, but I wrote some temporary code to compare the result from previous code and new code. You can check in the file changes. After the PR is approved, I will remove those temporary code.

Ideally, if we can get this through soon, we will be able to save some time and cost in the incoming taskmaster experiments.


sleepinyourhat:
Leaving this to @pyeres, but a couple of high-level comments:
- Check that logging is correct.
- When you write permanent tests, one of them should check situations where you use a limited data fraction in one stage of training, but the full dataset in the other stage of training. IIRC, this was a source of some extra complexity previously, and I believe it has come up as something we've found useful.

pyeres:
@HaokunLiu, thanks for the regression tests. Can you give an estimate of time savings w/ these changes? (useful for prioritizing review and any additional work that might be necessary for this PR)

jiant/utils/serialize.py
```diff
@@ -57,7 +58,48 @@ def bytes_to_float(b):
...
+        with open(filename, "rb") as fd:
+            for line in fd:
+                blob = base64.b64decode(line)
+                example = pkl.loads(blob)
+                yield example
+
+    return RepeatableIterator(_iter_fn) if repeatable else _iter_fn()
+
+
+# temporary backup code, remove before merging
```
pruksmhc, line 101: Please remove.

jiant/preprocess.py
```diff
@@ -399,8 +400,10 @@ def build_tasks(
             if force_reindex or not cache_found:
```
pruksmhc, line 400: Can you add a comment here about what's happening?

jiant/preprocess.py
```diff
@@ -399,8 +400,10 @@ def build_tasks(
...
                 record_file = _get_serialized_record_path(task.name, split, preproc_dir)
-                if os.path.exists(record_file) and os.path.islink(record_file):
-                    os.remove(record_file)
+                all_record_files = "%s*" % record_file
+                for one_record_file in glob.glob(all_record_files):
+                    if os.path.exists(one_record_file) and os.path.islink(one_record_file):
+                        os.remove(one_record_file)
 
                 _index_split(
                     task, split, indexers, vocab, record_file, model_preprocessing_interface
```
pruksmhc, line 409: Can you put a comment about what is happening here?

pruksmhc:
I'm _pretty_ sure I understand what's going on here, but not 100% sure - basically, you're adding functionality of loading the vocabulary for all the tasks before running each experiment for the one-time cost. A few requests. 
1) Please document more clearly what's going on, and flesh out what the "1-time cost" is so a stranger can understand what this PR is for. 
2) Please move the test to tests folder and make it a unit test.

jiant/utils/serialize.py
HaokunLiu: In case other reviewer would like to check this, I will remove it once the PR is approved

HaokunLiu:
> I'm _pretty_ sure I understand what's going on here, but not 100% sure - basically, you're adding functionality of loading the vocabulary for all the tasks before running each experiment for the one-time cost. A few requests.
> 
> 1. Please document more clearly what's going on, and flesh out what the "1-time cost" is so a stranger can understand what this PR is for.
> 2. Please move the test to tests folder and make it a unit test.

Errrr, this PR is not about that thing...  This PR is meant to address low gpu usage when data_fraction is very low, like 1%. Instead of iterating over all the instances and rejecting most of them, this create an additional preprocessed file that contains all the instances that will actually be used.

This is why when reload_index is True, we need to use glob to find out both the normally preprocessed file and our additional preprocessed files, and delete them all

HaokunLiu:
> Leaving this to @pyeres, but a couple of high-level comments:
> 
> * Check that logging is correct.
> * When you write permanent tests, one of them should check situations where you use a limited data fraction in one stage of training, but the full dataset in the other stage of training. IIRC, this was a source of some extra complexity previously, and I believe it has come up as something we've found useful.

The issue you mentioned will be addressed in #1070 . This PR is orthogonal to it.

jiant/utils/serialize.py
```diff
@@ -57,7 +58,48 @@ def bytes_to_float(b):
...
+      fraction: if set to a float between 0 and 1, load only the specified percentage
+        of examples. Hashing is used to ensure that the same examples are loaded each
+        epoch.
+
+    Returns:
+      iterable, possible repeatable, yielding deserialized Python objects
+    """
+
+    if fraction < 1.0:
+        frac_filename = f"{filename}__fraction_{fraction}"
```
pyeres, line 77: Because with this change the fraction argument is going to generate new (possibly big) files, we should consider giving the user a heads up. Can you add a note to the related options in the config mentioning this new behavior?

scripts/temporary_test_read_records.py
```diff
@@ -0,0 +1,57 @@
+from jiant.utils import serialize
```
pruksmhc, line 1: Can you please make this into a unit test?

HaokunLiu: I think this is more like a regression test. I don't see an easy way to convert it into a unit test.

pyeres: Hey @HaokunLiu — I tried to write something similar to the tests you wrote, but tried to make them more like unit tests. These tests end up involving `write_records`, but they'll add some protection from changes going forward: 

```
class TestReadRecords(unittest.TestCase):

    def test_read_records_without_data_fraction(self):
        """write records then read records (with no data fraction arg), check that records match"""
        data_file_name = "data.pb64"
        fake_example_count = 100
        fake_examples = [i for i in range(fake_example_count)]
        with tempfile.TemporaryDirectory() as tmp_dir_path:
            fake_filepath = os.path.join(tmp_dir_path, data_file_name)
            serialize.write_records(fake_examples, fake_filepath)
            fake_examples_read = serialize.read_records(fake_filepath)
            self.assertCountEqual(fake_examples, fake_examples_read)

    def test_read_records_with_data_fraction(self):
        """write examples, read without fraction, read with fraction, check expected files exist"""
        filename = "data.pb64"
        frac = 0.1
        fake_example_count = 100
        fake_examples = [i for i in range(fake_example_count)]
        with tempfile.TemporaryDirectory() as tmp_dir_path:
            fake_filepath = os.path.join(tmp_dir_path, filename)
            serialize.write_records(fake_examples, fake_filepath)
            fake_examples_read = serialize.read_records(fake_filepath)
            fake_examples_read_frac = serialize.read_records(fake_filepath, fraction=frac)
            file_list = os.listdir(tmp_dir_path)
            self.assertLess(len(list(fake_examples_read_frac)), len(list(fake_examples_read)))
            self.assertTrue(set(fake_examples_read_frac).issubset(set(fake_examples_read)))
            self.assertCountEqual(file_list, [filename, filename + "__fraction_" + str(frac)])

    def test_read_records_with_data_fraction_without_prior_full_data_read(self):
        """write examples, then read with fraction (no full read before fractional read)"""
        filename = "data.pb64"
        frac = 0.1
        fake_example_count = 100
        fake_examples = [i for i in range(fake_example_count)]
        with tempfile.TemporaryDirectory() as tmp_dir_path:
            fake_filepath = os.path.join(tmp_dir_path, filename)
            serialize.write_records(fake_examples, fake_filepath)
            fake_examples_frac = serialize.read_records(fake_filepath, fraction=frac)
            self.assertLess(len(list(fake_examples_frac)), len(list(fake_examples)))
            self.assertTrue(set(fake_examples_frac).issubset(set(fake_examples)))

    def test_read_records_with_data_fraction_from_cache(self):
        """write examples, read with fraction, then read with fraction again (from cache)"""
        filename = "data.pb64"
        frac = 0.1
        fake_example_count = 100
        fake_examples = [i for i in range(fake_example_count)]
        with tempfile.TemporaryDirectory() as tmp_dir_path:
            fake_filepath = os.path.join(tmp_dir_path, filename)
            serialize.write_records(fake_examples, fake_filepath)
            fake_examples_frac = serialize.read_records(fake_filepath, fraction=frac)
            os.remove(fake_filepath)  # remove full file make sure we get frac examples from cache
            fake_examples_frac_cache = serialize.read_records(fake_filepath, fraction=frac)
            self.assertCountEqual(list(fake_examples_frac), list(fake_examples_frac_cache))
```

pyeres:
Based on logs in some recent experiments it looks like these changes speed up data reading, however there's a bottleneck elsewhere (with these changes alone the speedup is <10%). 

Planning to close this PR. Issue "training_data_fraction is slow." #180 remains open to track issue. If this issue becomes a priority again, the changes in this PR should be considered as part of a broader fix.

Issue #1066, opened 2020-04-14
Results with pretrain_data_fraction args


claravania:
I'm trying to train MNLI in low-data regime (~750 examples) on RoBERTa, using the pretrain_data_fraction argument. I'm using default RoBERTa hyperparameter setup, but set the num epochs higher to allow the model trained longer. However, I got quite high performance 89.02 macro_avg, just slightly lower than using full data, 90.2 macro_avg. Could there possibly be some bugs?

Attaching log file, config file, and my command.

Command:
`python main.py --config_file jiant/config/nli-roberta_conf.txt -o "run_name=run1, random_seed=123456"`

[nli-roberta_conf.txt](https://github.com/nyu-mll/jiant/files/4477438/nli-roberta_conf.txt)
[log.log](https://github.com/nyu-mll/jiant/files/4477412/log.log)

(changed .conf to _conf.txt since I can't upload the format to GitHub)

pyeres:
It looks like the task's `task.train_data`, which is an instance generator with a pretraining data fraction arg, is being overwritten with an instance generator with the _target_ data fraction arg when task name appears in both the `pretrain_task_names` and `target_task_names`:

https://github.com/nyu-mll/jiant/blob/de3c44a6c2aad2bdfc5bc3047e063fd03c6c4c12/jiant/preprocess.py#L421-L436

pyeres:
Update: I'm working on a PR to address this.

PR #1070, opened 2020-04-16
Fix target_train_data_fraction overriding pretrain_data_fraction

pyeres:
This PR addresses the bug described in issue #1066 ("Results with pretrain_data_fraction args").

This PR...
* Adds a new private field to the `Task` object (`_instance_generators`) to store instance generators with the appropriate phase-specific data fraction settings.
* Adds new methods, getter `get_instance_generator` and setter `set_instance_generator`, to `Task` to access the `_instance_generators`.
* Updates preprocessing, training, and evaluation code to use the new getter and setter.
* Updates a test that uses evaluation logic to use the new getter and setter.
* Adds tests for `Task`'s that an exception is raised when training instance generator is requested without specifying the phase (because phase determines data fraction).

Validation: the config provided in issue #1066 was run with these code changes. Results logged for the task (MNLI) using a pretraining data fraction were more in line with expectation: after 25k steps training accuracy was ~100%, while val and test accuracy were ~75%.


jiant/tasks/tasks.py
```diff
@@ -293,6 +295,41 @@ def handle_preds(self, preds, batch):
...
         return preds
 
+    def set_instance_generator(
+        self, dataset_name: str, instance_generator: Iterable, phase: str = None
+    ):
+        """Takes a data instance generator and stores it in a private field of this Task instance
+
+        Parameters
+        ----------
+        dataset_name : string
```
sleepinyourhat, line 305: I'm more familiar with 'dataset' referring to a full collection, not a specific split subset of the data. How about something like 'split_name' instead?

sleepinyourhat: Otherwise, it's plausible that future code editors would try setting something like dataset_name = "mnli".

jiant/trainer.py
```diff
@@ -322,7 +322,12 @@ def _setup_training(
                 os.mkdir(os.path.join(self._serialization_dir, task.name))
 
             # Adding task-specific smart iterator to speed up training
```
sleepinyourhat, line 324: It's okay to leave this be, but do you know what 'smart' is supposed to mean here?

sleepinyourhat: I suspect it refers to some contrast that's no longer relevant...

jiant/trainer.py
```diff
@@ -335,7 +340,9 @@ def _setup_training(
                 biggest_batch_first=True,
             )
             task_info["iterator"] = iterator
-            task_info["tr_generator"] = iterator(task.train_data, num_epochs=None)
+            task_info["tr_generator"] = iterator(
```
sleepinyourhat, line 343: The naming seems a bit confusing here. If you're calling 'get_instance_generator', shouldn't that return something that can be used as a 'generator'?

jiant/trainer.py
pyeres: [Looks like](https://github.com/nyu-mll/jiant/pull/87/files) it was added to describe the addition of a BucketIterator. Dropping w/ a44d4d5.

jiant/tasks/tasks.py
pyeres: 👍, done w/ c5d6822.

jiant/trainer.py
pyeres: Thanks — resolved some naming confusion in my changes w/ 72a5bfe.

PR #1074, opened 2020-04-16
Remove --remote_log from exp_fns.sh

iftenney:
Allow running these if not on GCP. The log should still be in a text file in the console.

If running on Kubernetes, this flag doesn't matter anyway since k8s does its own log management.


sleepinyourhat:
LGTM. Approving as edge probing-specific.

jeswan:

Issue #1076, opened 2020-04-20
Transient test failures from MLM/#1030


sleepinyourhat:
It looks like there's a probabilistic component to at least one of the tests introduced by #1030—it is failing and passing sporadically. As of now, CircleCI says that `jiant` is broken at head, since the recent README-only PR got unlucky and caused a random failure.

Please fix ASAP—by whatever means seem best in the context of the original PR—so that we can trust the results of the tests.

pyeres:
Closing as addressed by @pruksmhc's changes w/ #1077 (now in master). Thanks @pruksmhc.

PR #1077, opened 2020-04-21
Increasing sample size for MLM test

pruksmhc:
This PR addresses https://github.com/nyu-mll/jiant/issues/1076
Because this is a statistical test, this PR increases the length of input passed in and number of examples to average over in order, as well as slightly increase the error boundary around the masking ratio in order to decrease the amount of failed tests. 
I tested this by running the MLM test 15 times with all successes.


zphang:
How about making the test deterministic with seeds?

From a quick look, setting the random seed of `random` and `torch` should cover all sources of randomness.

pruksmhc:
I think torch.bernoulli could be made more deterministic with the seed, but I just took out the random.randint that produced the length (just because I don't think there's much you lose by setting it to a large number).

pyeres:
@pruksmhc — @zphang's suggestion should only require adding two lines [at line L77](https://github.com/nyu-mll/jiant/blob/67238afad48b3f71463a54debe7e9d2047636d92/tests/tasks/test_mlm.py#L77):

```
torch.manual_seed(0)
random.seed(0)
```

With the additional length/iteration in your PR lines L79-L80 this one test takes a couple seconds on my laptop — people generally try to keep the total time to run all their unit tests together to less than a second. If you want to add this additional time, I want to check that we're getting something beyond what we can accomplish with the seeds alone.

pruksmhc:
Alright, I just increased the length of each sample instead of increasing the number of samples.

Issue #658, opened 2019-05-03
Turn off AllenNLP's parameter value logging


sleepinyourhat:
AllenNLP dumps a bunch of redundant parameter values to the log by default, and we can't easily access the code that would disable this. Either:
– Find a way to pass AllenNLP a decoy log.
– Patch AllenNLP to turn off verbose logging.

Sub-issue of #602

pruksmhc:
Did some preliminary research into this, seems relevant
https://github.com/allenai/allennlp/issues/2506

pruksmhc:
Addressed with https://github.com/nyu-mll/jiant/commit/7be8fdb9b8b24296b559bf307d7cacbe9a669b33

sleepinyourhat:
I'm still seeing a bunch of relevant-looking logs junk: 

06/24 04:03:12 PM: Number of trainable parameters: 334649361
06/24 04:03:12 PM: Converting Params object to dict; logging of default values will not occur when dictionary parameters are used subsequently.
06/24 04:03:12 PM: CURRENTLY DEFINED PARAMETERS: 
06/24 04:03:12 PM: lr = 2e-05
06/24 04:03:12 PM: t_total = 29500
06/24 04:03:12 PM: warmup = 0.1
06/24 04:03:12 PM: type = reduce_on_plateau
06/24 04:03:12 PM: Converting Params object to dict; logging of default values will not occur when dictionary parameters are used subsequently.
06/24 04:03:12 PM: CURRENTLY DEFINED PARAMETERS: 
06/24 04:03:12 PM: mode = max
06/24 04:03:12 PM: factor = 0.5
06/24 04:03:12 PM: patience = 4
06/24 04:03:12 PM: threshold = 0.0001
06/24 04:03:12 PM: threshold_mode = abs
06/24 04:03:12 PM: verbose = True
06/24 04:03:12 PM: type = bert_adam
06/24 04:03:12 PM: parameter_groups = None
06/24 04:03:12 PM: Number of trainable parameters: 334649361
06/24 04:03:12 PM: Converting Params object to dict; logging of default values will not occur when dictionary parameters are used subsequently.
06/24 04:03:12 PM: CURRENTLY DEFINED PARAMETERS: 
06/24 04:03:12 PM: lr = 2e-05
06/24 04:03:12 PM: t_total = 29500
06/24 04:03:12 PM: warmup = 0.1
06/24 04:03:12 PM: type = reduce_on_plateau
[...]

PR #1080, opened 2020-04-28
Suppress verbose INFO-level logging by AllenNLP components in trainer.py

pyeres:
Addresses https://github.com/nyu-mll/jiant/issues/658 by temporarily increasing (and then restoring) the log level around [this code](https://github.com/nyu-mll/jiant/blob/master/jiant/trainer.py#L510-L513) from `INFO` to `WARNING` (`WARNING` is the level above `INFO`, and warnings will still be logged).

### DIFF
--- a/jiant/trainer.py
+++ b/jiant/trainer.py
@@ -507,10 +507,15 @@ class SamplingMultiTaskTrainer:
             else:
                 val_limit = self._max_vals
             optimizer_params["t_total"] = val_limit * self._val_interval
+
+        # temporarily increase the log level to avoid some verbose INFO-level logging from AllenNLP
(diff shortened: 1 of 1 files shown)


pyeres:
> Looks reasonable to me! Would you mind showing a diff of the logs for a before-and-after pair of runs on demo.conf (maybe excluding any random diff in the actual loss values), just to show the impact of the change?

Sure, here's a diff (logs for demo.conf w/ PR branch on the left and master branch on the right):

Screen Shot 2020-04-28 at 9 37 09 PM

And [here's the full log comparison](https://www.diffchecker.com/Z0lUd32L) (again PR branch on the left, master branch on the right).

Issue #1082, opened 2020-04-29
load_tsv skips "bad lines"


pyeres:
**Background**: an issue was raised describing "Potentially corrupts SNLI data" (https://github.com/nyu-mll/jiant/issues/1073). The header of the SNLI data downloaded by jiant contains 11 fields, however the following data has as many as 15 tab-separated values.

**This issue**: by default the function that reads task data into tasks ([`load_tsv(...)`](https://github.com/nyu-mll/jiant/blob/master/jiant/utils/data_loaders.py#L139-L148)) skips "bad lines" (see `error_bad_lines` in [pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html)). This removes a layer of protection that may be helpful when processing potentially corrupted task data files.

Issue #1073, opened 2020-04-16
Potentially corrupts SNLI data


Apsod:
After running the download_glue_data script, the train.tsv file has an inconsistent number of fields: 

e.g. 
line 1171 has 11 fields: `'1171\t1974336555.jpg#0\t1974336555.jpg#0r1c\t( ( ( A ( young man ) ) ( in ( his ( mid twenties ) ) ) ) ( ( is ( ( kicking ( his ( left foot ) ) ) ( ( ( about two ) feet ) ( off ( ( the leaf ) ( ( covered ( ( ground , ) ( with ( paved ( asphalt ( and ( green ( ( plants and ) trees ) ) ) ) ) ) ) ) ( in ( the background ) ) ) ) ) ) ) ) . ) )\t( ( two women ) ( play soccer ) )\t(ROOT (S (NP (NP (DT A) (JJ young) (NN man)) (PP (IN in) (NP (PRP$ his) (JJ mid) (NNS twenties)))) (VP (VBZ is) (VP (VBG kicking) (NP (PRP$ his) (JJ left) (NN foot)) (PP (NP (QP (RB about) (CD two)) (NNS feet)) (IN off) (NP (NP (DT the) (NN leaf)) (VP (VBN covered) (NP (NP (NN ground)) (, ,) (PP (IN with) (NP (JJ paved) (NN asphalt) (CC and) (JJ green) (NNS plants) (CC and) (NNS trees)))) (PP (IN in) (NP (DT the) (NN background)))))))) (. .)))\t(ROOT (NP (NP (CD two) (NNS women)) (NP (NN play) (NN soccer))))\tA young man in his mid twenties is kicking his left foot about two feet off the leaf covered ground, with paved asphalt and green plants and trees in the background.\ttwo women play soccer\tcontradiction\tcontradiction\n'`
[...]

PR #1083, opened 2020-04-30
Update load_tsv to stop if "bad lines" are found

pyeres:
This PR is motivated by issue #1082 (please see that issue for important context).

This PR would change the behavior of `load_tsv()`: if pandas finds a "bad line" in the data passed to `load_tsv()`, a `pandas.errors.ParserError` will be thrown and the run will stop.

**How was this PR validated/tested?** For all tasks where I had task data on hand (10+ tasks including SNLI), on this PR branch (i.e., w/ `error_bad_lines=True`) I ran code to trigger a call to `load_tsv()`, and for all tasks except SNLI `load_tsv()` worked without an exception. For SNLI (where [we know there's some inconsistency](https://github.com/nyu-mll/jiant/issues/1073) in the format of the data file) this (expected) exception is raised:

```
pandas.errors.ParserError: Error tokenizing data. C error: Expected 11 fields in line 13, saw 15
```

**What will this break**: this change will cause SNLI task data loading to fail with exception like above until another fix (e.g., an upstream fix to rely on SNLI data without "bad lines") is applied.

### DIFF
--- a/jiant/utils/data_loaders.py
+++ b/jiant/utils/data_loaders.py
@@ -139,7 +139,7 @@ def load_tsv(
     rows = pd.read_csv(
         data_file,
         sep=delimiter,
-        error_bad_lines=False,
+        error_bad_lines=True,
         header=None,
         skiprows=skip_rows,
         quoting=quote_level,


sleepinyourhat:
@W4ngatang - Could you look into the SNLI issue? It sounds like that's specific to the GLUE-downloader version that `jiant` uses by default, and if that's easy to fix, it's probably better to fix it before we merge this PR.

Issue #1087, opened 2020-05-15
Incompatibility in model checkpoints between multi-GPU/single-GPU


zphang:
**Summary**: If you train a model with multi-GPU, then try to load it with single-GPU, the weights will not be successfully loaded. The run will print warnings but will not fail.

Background:
* When a PyTorch module is wrapped in `DataParallel`, the original module will be stored as `multi_gpu_model.module`. 
* This means that calling `state_dict` on `multi_gpu_model` will add a "module." prefix to every key in the `state_dict`.
* In other words, `state_dict`s are not compatible between single-gpu and multi-gpu models.

What this affects:
* Two-phase training runs that are run from two separate commands (via `load_target_train_checkpoint`) will incorrectly load the model weights, reverting to the pretrained weights **if one phase uses multi-GPU and the other uses single-GPU**
* E.g. If intermediate task training is on multi-GPU and target task training is on single-GPU, the target task run will fail to load the intermediate-trained weights, and revert to pretrained weights. The logs will be filled with "parameter missed" warnings, **but will not fail**

Proposed solution:
* Don't call `state_dict` on a DataParallel model. Write a function that determines whether a model is DaraParallel, and call `model.state_dict` or `model.module.state_dict` correspondingly

```python
def get_state_dict_for_saving(model: nn.Module) -> nn.Module:
    if isinstance(model, nn.DataParallel):
        return model.module.state_dict()
    else:
        return model.state_dict()
```

sleepinyourhat:
Eek—does anyone have bandwidth to put together a fix?

pruksmhc:
I can take a look at this this weekend

pruksmhc:
Update: It's actually not _quite_ as simple as just putting the above function to modify saving, since modifying saving will not work if we want to restart a job from multi-GPU (because the saved model will now always have model.state_dict() while the model in multi-GPU expects module.model.state_dict()

zphang:
Could you instead call `model.module.load_state_dict`?

zphang:
*This is an automatically generated comment.*

As we update `jiant` to v2.x, `jiant` v1.x has been migrated to https://github.com/nyu-mll/jiant-v1-legacy. As such, we are closing all issues relating to `jiant` v1.x in this repository.

If this issue is still affecting you in `jiant` v1.x, please follow up at https://github.com/nyu-mll/jiant-v1-legacy/issues/1087.

If this issue is still affecting you in `jiant` v2.x, reopen this issue or create a new one.

PR #1091, opened 2020-05-17
Fixing model checkpoints to be robust to mutli -> single GPU usage

pruksmhc:
This is a fix to https://github.com/nyu-mll/jiant/issues/1087. I decided to make the change in the model loading portion because making the change in model saving as suggested in the https://github.com/nyu-mll/jiant/issues/1087 will fix multi -> single GPU model loading, but will break multi -> multi GPU model loading (if we want to reload a checkpoint that was trained in multi-GPU on a multi-GPU machine). 
Additionally, I also did some light cleanup of model loading in trainer to not be redundant, and also deleted an unused parameter. 

**Tests**
Multi -> Single GPU: I tested by training a roberta-large model on SST on multi-GPU, and then loading that checkpoint in a single-GPU for further training. 
Multi -> Multi GPU: This is implicitly already done in jiant, specifically we load the best checkpoint before doing evaluation, so this was tested when I trained the roberta-large SST model the first time on multi-GPU.


jiant/utils/utils.py
```diff
@@ -323,22 +325,48 @@ def select_task_specific_args(exp_args, diff_args):
...
+    final_model_state = collections.OrderedDict()
+
+    def get_key(name):
+        key = name
+        if isinstance(model, nn.DataParallel):
+            if "module" not in key:
+                key = "module.%s" % key
+        else:
+            if "module" in key:
+                key = key.replace("module.", "")
```
zphang, line 346: Make this check explicitly for prefix with `.startswith`, and drop the first n characters (in case module appears somewhere else in the parameter name).

jiant/utils/utils.py
```diff
@@ -323,22 +325,48 @@ def select_task_specific_args(exp_args, diff_args):
...
+        return key
+
+    for name, weights in model_state.items():
+        key = get_key(name)
+        final_model_state[key] = model_state[name]
```
zphang, line 351: ` = weights`

jiant/trainer.py
```diff
@@ -1217,15 +1217,7 @@ def _restore_checkpoint(self, phase, tasks=None):
...
-        model_state = torch.load(model_path)
-
-        for name, param in self._model.named_parameters():
-            if param.requires_grad and name not in model_state:
-                log.error("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-                log.error("Parameter missing from checkpoint: " + name)
-                log.error("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-
-        self._model.load_state_dict(model_state, strict=False)
+        load_model_state(self._model, model_path)
```
zphang, line 1220: Is this warning being disabled?

pruksmhc: No it's not. This warning is also inside load_model_state function

pyeres:
Hi @zphang & @HaokunLiu — are either of you available to provide the substantial review for this PR? The core concerns seem to be 1) whether this addresses issue #1087, and 2) whether these changes introduce new risks/regressions.

jiant/utils/utils.py
```diff
@@ -323,22 +325,49 @@ def select_task_specific_args(exp_args, diff_args):
...
+        multi-GPU and single-GPU use cases
+
+    Parameters
+    ----------
+    model: The model object to populate with loaded parameters.
+    model_state: collections.OrderdDict of model state
+    """
+    final_model_state = collections.OrderedDict()
+
+    def get_key(name):
```
sleepinyourhat, line 339: Add a comment explaining why we need this logic.

jiant/trainer.py
```diff
@@ -1217,15 +1217,7 @@ def _restore_checkpoint(self, phase, tasks=None):
...
 
-        model_state = torch.load(model_path)
-
-        for name, param in self._model.named_parameters():
-            if param.requires_grad and name not in model_state:
-                log.error("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-                log.error("Parameter missing from checkpoint: " + name)
-                log.error("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-
-        self._model.load_state_dict(model_state, strict=False)
```
sleepinyourhat, line 1228: You're no longer setting strict=False here. It's debatable whether that's the ideal behavior here, but it was intentional, and I believe it has had some real experimental uses. Why the change?

jeswan:

PR #1100, opened 2020-06-28
Update __init__.py so that MiniBERTas can be probed

YianZhang:
Update from the inductive-bias-pretraining team (Sam, Alex, Haokun, Haau-Sing, Yian). Four new roberta model names are added to _transformer\_input\_module\_to\_tokenizer\_name_ so that our new models can be used as input_modules in jiant.


sleepinyourhat:
@YianZhang Thanks! 

- Have you confirmed that these work on a clean installation of `jiant`? In particular, make sure that the models auto-download when you specify these names.
- Please update defaults.conf to mention these options.

YianZhang:
@sleepinyourhat 

This update is for the somewhat hacky solution we agreed on on slack: In the tutorial we will require the user to download the checkpoint themselves, rename it and put it under the root directory. I did not look into auto-download function, but I can have a try. Do you think this is necessary?

sleepinyourhat:
Ah, got it. If I understand correctly, this won't be an advertised/core model in jiant, but there will be a tutorial on an external page explaining how to use it. That sounds reasonable to me, since this is just a pointer to another checkpoint file for the same model—we can pretty much guarantee that no future PR will break support for these models without also breaking support for RoBERTa.

@pyeres: Any objection to something like this? Anything we should keep in mind?

YianZhang:
@sleepinyourhat @pyeres Sam: Yes, you understand it correctly. 

I checked the code again and I think it won't be crazy to include them as an advertised/core model in jiant, though it might take a ittle longer for debugging. Another option is to start officially supporting these models in jiant2.0 if you find it worthwhile.

YianZhang:
Please ignore this PR. I will make a new one where MiniBERTas can be used just like other input_modules.

PR #1101, opened 2020-06-29
Updates to support using MiniBERTas as input_module

YianZhang:
We pretrained RoBERTa on limited amounts of data, and released the model checkpoints on Huggingface Transformers. This PR will allow jiant to use MiniBERTas as input_module and will support edge probing with MiniBERTas.


YianZhang:
@sleepinyourhat I tested nyu-mll/roberta-base-1B-3, roberta-base, and bert-base-uncased on the edge probing dependency labeling task. All experiments can be run with no errors; all models achieve reasonable f1 (~95). The miniberta model downloads automatically.

YianZhang:
@sleepinyourhat Thanks for the feedback! What about now? I fixed all the style issues reported by pep8speaks except for those that are not caused by my commits.

sleepinyourhat:
Merging. FWIW, We're hoping to talk more soon about how to do things like this in jiant 2, which should come out in some form (alpha?) next week, since we'd like to treat it as less of a core feature, but still leave it usable.

YianZhang:
@sleepinyourhat Great! Let me know then how I can be of help.

sleepinyourhat:
Possibly now relevant to @phu-pmh, CC @jeswan.

sleepinyourhat:
(As a jiant 1.x example of something we might need to do in jiant 2.x)

PR #1121, opened 2020-09-24
Proposed citation for v2

sleepinyourhat:
Trying to list v2 contributors in rough order of contribution, then v1-only contributors. Not at all sure I got the v2 order right. Suggestions?


jeswan (approved):
Most familiar with the first four contributors. Order for those looks good to me.

PR #1135, opened 2020-10-07
Adding SingleTaskConfigurator, some cleanup

zphang:


jiant/proj/main/scripts/configurator.py
```diff
@@ -38,10 +38,175 @@ def cap_examples(num_examples, cap):
         return min(num_examples, cap)
 
 
+@Registry.register
+@zconf.run_config
+class SingleTaskConfigurator(zconf.RunConfig):
```
jeswan, line 43: Would prefer to not duplicate everything here. Can the configurators be combined?

zphang: This is meant to be a much more streamlined configurator since it only supports one task - the multi-task configurator has a lot more complexity because it needs to deal with potentially different tasks in different phases.

jeswan (approved):

PR #1146, opened 2020-10-12
Irt tasks wh

wh629:
Including Winogrande, ARCT, and MCScript.


sleepinyourhat:
Some PR basics:

- When changing ML functionality, you need benchmark results of some kind. For a new task, this usually involves finding a previously published dev or test number with some `jiant` model (usually BERT or RoBERTa), showing commands that allow you to reproduce that result, and explicitly comparing your number against the original numbers.
- Make the smallest changes you can in each PR. Unless these tasks are closely tied (i.e., they share a lot of overlapping new code), they should probably be separate PRs. 
- Make sure every line changed in the PR corresponds to the purpose of the PR. It looks like something went wrong with Black or your editor, so there are tons of changes visible in the diff view that have nothing to do with these tasks. Revert or fix that before requesting review.
- Make sure tests pass.
- Resolve the 'out-of-date' error above.

PR #1203, opened 2020-10-18
winogrande

wh629:
Adding winogrande task. Achieves 79-80% on validation accuracy across multiple restarts using LRs of 3e-6, 5e-6. Original paper reports 79.3% (https://arxiv.org/pdf/1907.10641.pdf). Used attached scripts to run. @keisuke-sakaguchi

[winogrande.zip](https://github.com/nyu-mll/jiant/files/5398063/winogrande.zip)


keisks:
Hello, William (@wh629).

Thank you for adding WinoGrande dataset into `jiant`!
If you have any questions about the dataset, feel free to ask me (@keisks)!!

jiant/scripts/download_data/datasets/files_tasks.py
```diff
@@ -222,3 +226,40 @@ def download_qasrl_data_and_write_config(
...
+                "train_xs": os.path.join(task_data_path, "train_xs.jsonl"),
+                "train_xs_labels": os.path.join(task_data_path, "train_xs-labels.lst"),
+                "train_s": os.path.join(task_data_path, "train_s.jsonl"),
+                "train_s_labels": os.path.join(task_data_path, "train_s-labels.lst"),
+                "train_m": os.path.join(task_data_path, "train_m.jsonl"),
+                "train_m_labels": os.path.join(task_data_path, "train_m-labels.lst"),
+                "train_l": os.path.join(task_data_path, "train_l.jsonl"),
+                "train_l_labels": os.path.join(task_data_path, "train_l-labels.lst"),
+                "train_xl": os.path.join(task_data_path, "train_xl.jsonl"),
+                "train_xl_labels": os.path.join(task_data_path, "train_xl-labels.lst"),
```
zphang, line 257: Are these files being used? (Are they for alternative versions of the datasets?)

wh629: Only the XL training set is being used. I kept the other training set sizes as reference. Submission to the leaderboard requires predictions from models trained on each of the training sets (XS, S, M, L, XL).

jiant/scripts/download_data/datasets/files_tasks.py
```diff
@@ -33,6 +33,10 @@ def download_task_data_and_write_config(task_name: str, task_data_path: str, tas
         download_qasrl_data_and_write_config(
             task_name=task_name, task_data_path=task_data_path, task_config_path=task_config_path
         )
+    elif task_name == "winogrande":
+        download_winogrande_data_and_write_config(
+            task_name=task_name, task_data_path=task_data_path, task_config_path=task_config_path
+        )
```
jeswan, line 39: It looks like `datasets` supports [Winogrande](https://huggingface.co/nlp/viewer/?dataset=winogrande&config=winogrande_xs). Can you switch to downloading the dataset using `datasets` instead of a direct download?

jeswan (changes requested):

jiant/scripts/download_data/datasets/files_tasks.py
wh629: I tried this originally, and there were error statements from assertions checking dataset size. It seems like this might be a mismatch between the dataset size constant used in `datasets` and the source file, but I'm not sure.

jeswan: I'll take a look at these errors. Thanks!

jiant/tasks/lib/winogrande.py
```diff
@@ -0,0 +1,123 @@
...
+    def _create_examples(cls, lines, set_type):
+        examples = []
+
+        for i, line in enumerate(lines):
+            choice_pre, option = line["sentence"].split("_")
+
+            examples.append(
+                Example(
+                    guid="%s-%s" % (set_type, i),
+                    prompt=option,
```
HaokunLiu, line 118: {"qID": "3FCO4VKOZ4BJQ6IFC0VAIBK4KTWE7U-2", "sentence": "Sarah was a much better surgeon than Maria so _ always got the easier cases.", "option1": "Sarah", "option2": "Maria", "answer": "2"}

I can tell the outcome will be the same, but is the choice and prompt in TokenizedExample and here both reversed? And the variable name option is not very precise.

wh629: Yes, they have been reversed in TokenizedExample as well. This is true about the variable name option... Let me know if I should change this in another push.

jeswan:
Hi Will, can you merge your changes in `jiant/scripts/download_data/datasets/files_tasks.py` to jiant/scripts/download_data/dl_datasets/files_tasks.py

jeswan (approved):

PR #1205, opened 2020-10-20
Adding ability to resume run in Simple API

zphang:


guides/tutorials/quick_start_main.md
```diff
@@ -52,24 +52,66 @@ python jiant/proj/main/scripts/configurator.py \
...
+
+```
+    --save_every_steps 500
+```
+
+will save a pickle of model weights every 500 training steps.
+
+### Early Stopping
+
+To do early stopping, we can perform validation evaluation at regular intervals over the course of training, and select the best model weights based on validation performance. For expedience, we often do not want to evaluate on the whole validaiton set, but only a subset. To do early stopping, use the following arguments as an example:
```
jeswan, line 107:
```suggestion
To do early stopping, we can perform validation evaluation at regular intervals over the course of training, and select the best model weights based on validation performance. For expedience, we often do not want to evaluate on the whole validation set, but only a subset. To do early stopping, use the following arguments as an example: 
```

jiant/proj/simple/runscript.py
```diff
@@ -234,148 +243,19 @@ def run_simple(args: RunConfiguration):
...
 def main():
     mode, cl_args = zconf.get_mode_and_cl_args()
     args = RunConfiguration.default_run_cli(cl_args=cl_args)
     if mode == "run":
-        run_simple(args)
-    elif mode == "dry_run":
-        dry_run(args)
+        run_simple(args, with_continue=False)
+    if mode == "run_with_continue":
+        run_simple(args, with_continue=True)
```
jeswan, line 258: Do you have the bandwidth to add a CI test to use this functionality? I can add it after as well.

zphang: Let's do it after? Potentially we can alter the way it's called to make it easier to isolate+test.

jeswan (approved):

PR #1215, opened 2020-10-25
Adds Fever NLI task and data downloader

angie-chen55:


jeswan (changes requested):
Thanks for this PR! Please add step 4 of https://github.com/nyu-mll/jiant/blob/master/guides/tasks/adding_tasks.md and address the formatting error.

jeswan (approved):

PR #1216, opened 2020-10-26
PIQA 

wh629:
Using RoBERTa-large, achieved 77.8% validation accuracy. Original paper achieves 79.2% validation accuracy (https://arxiv.org/pdf/1911.11641.pdf). Used the following scripts for testing.
[piqa.zip](https://github.com/nyu-mll/jiant/files/5439434/piqa.zip)

@ybisk


jeswan (approved):

ybisk:
A. Thanks for this!
B. I haven't had a chance to look more carefully but the reproduction requires the following: roughly these settings -- lr 1e-5, 8 ep, 4/batch/GPU (4x V100), max seq 150

PR #1220, opened 2020-10-29
Fix save-last behavior

zphang:
This commit makes explicit separate save-best and save-last behaviors, into `best_model.p` and (new) `last_model.p`.

### DIFF
--- a/jiant/proj/main/metarunner.py
+++ b/jiant/proj/main/metarunner.py
@@ -7,6 +7,7 @@ import jiant.proj.main.runner as jiant_runner
 import jiant.proj.main.components.task_sampler as jiant_task_sampler
 from jiant.shared.runner import (
     save_model_with_metadata,
+    save_state_dict_with_metadata,
     compare_steps_max_steps,
 )
 from jiant.utils.python.datastructures import ExtendedDataClassMixin
@@ -66,6 +67,7 @@ class JiantMetarunner(AbstractMetarunner):
         verbose: bool = True,
         save_best_model: bool = True,
         load_best_model: bool = True,
+        save_last_model: bool = True,
         log_writer: BaseZLogger = PRINT_LOGGER,
     ):
         self.runner = runner
@@ -78,6 +80,7 @@ class JiantMetarunner(AbstractMetarunner):
         self.verbose = verbose
         self.save_best_model = save_best_model
         self.load_best_model = load_best_model
+        self.save_last_model = save_last_model
         self.log_writer = log_writer
 
         self.best_val_state = None
@@ -122,6 +125,27 @@ class JiantMetarunner(AbstractMetarunner):
             file_name=f"model__{self.train_state.global_steps:09d}",
         )
 
+    def save_last_model_with_metadata(self):
+        save_model_with_metadata(
+            model=self.model,
+            metadata={
+                "train_state": self.train_state.to_dict(),
+            },
+            output_dir=self.output_dir,
+            file_name="last_model",
+        )
+
+    def save_best_model_with_metadata(self, val_metrics_dict):
+        save_state_dict_with_metadata(
+            state_dict=self.best_state_dict,
+            metadata={
+                "val_state": self.best_val_state.to_dict(),
+                "val_metrics": self.best_val_state.metrics,
+            },
+            output_dir=self.output_dir,
+            file_name="best_model",
+        )
+
     def should_save_checkpoint(self) -> bool:
         if self.save_checkpoint_every_steps == 0:
             return False
@@ -159,6 +183,7 @@ class JiantMetarunner(AbstractMetarunner):
         return False
 
     def done_training(self):
+        self.save_last_model_with_metadata()
         self.eval_save()
         if self.load_best_model and self.best_state_dict is not None:
             if self.verbose:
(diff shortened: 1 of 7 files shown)


jiant/proj/main/metarunner.py
```diff
@@ -122,6 +125,27 @@ def save_model(self):
...
+        save_state_dict_with_metadata(
+            state_dict=self.best_state_dict,
+            metadata={
+                "val_state": self.best_val_state.to_dict(),
+                "val_metrics": self.best_val_state.metrics,
+            },
+            output_dir=self.output_dir,
+            file_name="best_model",
+        )
+
```
jeswan, line 148: These functions seem like they are doing the same thing in different ways. Can they be combined?

jeswan: This also looks like a potential case of excessive delegation.

jeswan:
Please add a test for these additions or create an issue for a test and assign it to me.

jeswan (changes requested):
Please add a test for these additions or create an issue for a test and assign it to me.

jiant/proj/main/metarunner.py
zphang: It looks excessive, but the specific reason for this is to allow these methods to be overridden. (E.g. for adapters, we don't want to save all the model's weights, so we need a custom save function). Happy to implement another way, but I've found this helpful.

zphang: Side benefit: can override for testing

jeswan: I still think there is room for refactoring here. The function signature should similar to:
```python
def save_model(self, model, file_name, metadata_dict)
```

jiant/shared/runner.py
```diff
@@ -55,20 +56,16 @@ def get_eval_dataloader_from_cache(
...
+        state_dict = model_or_state_dict
+    else:
+        state_dict = torch_utils.get_model_for_saving(model_or_state_dict).state_dict()
 
-def save_state_dict_with_metadata(
-    state_dict: dict, metadata: dict, output_dir: str, file_name="model"
-):
     torch.save(state_dict, os.path.join(output_dir, f"{file_name}.p"))
-    py_io.write_json(metadata, os.path.join(output_dir, f"{file_name}.metadata.json"))
+    if metadata is not None:
```
jeswan, line 71:
```suggestion
    if metadata:
```

jiant/shared/runner.py
```diff
@@ -55,20 +56,16 @@ def get_eval_dataloader_from_cache(
...
+        py_io.write_json(metadata, os.path.join(output_dir, f"{file_name}.metadata.json"))
```
jeswan, line 72:
```suggestion
    if metadata:
        py_io.write_json(metadata, os.path.join(output_dir, f"{file_name}.metadata.json"))
```

jiant/shared/runner.py
```diff
@@ -55,12 +56,20 @@ def get_eval_dataloader_from_cache(
...
 
 
-def save_model_with_metadata(model: nn.Module, metadata: dict, output_dir: str, file_name="model"):
-    torch.save(
-        torch_utils.get_model_for_saving(model).state_dict(),
-        os.path.join(output_dir, f"{file_name}.p"),
-    )
-    py_io.write_json(metadata, os.path.join(output_dir, f"{file_name}.metadata.json"))
+def save_model_with_metadata(
+    model_or_state_dict: Union[nn.Module, dict],
```
jeswan, line 60:
```suggestion
    state_dict: dict,
```

jiant/shared/runner.py
```diff
@@ -55,12 +56,20 @@ def get_eval_dataloader_from_cache(
...
+    output_dir: str,
+    file_name="model",
+    metadata: Optional[dict] = None,
+):
+    if isinstance(model_or_state_dict, dict):
+        state_dict = model_or_state_dict
+    else:
+        state_dict = torch_utils.get_model_for_saving(model_or_state_dict).state_dict()
```
jeswan, line 68:
```suggestion
```

jiant/shared/runner.py
jeswan, line 60:
```suggestion
def save_state_dict_metadata(
    state_dict: dict,
```

jiant/proj/main/metarunner.py
```diff
@@ -218,20 +239,12 @@ def eval_save(self):
...
-                    },
-                    output_dir=self.output_dir,
-                    file_name="best_model",
-                )
             del self.best_state_dict
             self.best_state_dict = copy_state_dict(
                 state_dict=get_model_for_saving(self.model).state_dict(), target_device=CPU_DEVICE,
             )
+            if self.save_best_model:
+                self.save_best_model_with_metadata(val_metrics_dict=val_metrics_dict)
```
jeswan, line 247:
```suggestion
                self.save_state_dict(state_dict=self.best_state_dict, output_dir=self.output_dir, metadata=val_metrics_dict, file_name="best_model")
```

jiant/shared/runner.py
```diff
@@ -55,12 +56,20 @@ def get_eval_dataloader_from_cache(
...
-    torch.save(
-        torch_utils.get_model_for_saving(model).state_dict(),
-        os.path.join(output_dir, f"{file_name}.p"),
-    )
-    py_io.write_json(metadata, os.path.join(output_dir, f"{file_name}.metadata.json"))
+def save_model_with_metadata(
+    model_or_state_dict: Union[nn.Module, dict],
+    output_dir: str,
+    file_name="model",
+    metadata: Optional[dict] = None,
```
jeswan, line 63:
```suggestion
    model: Optional,
    file_name="model",
    metadata: Optional[dict] = None,
```

jeswan:
Planning to present a refactored alternative to `save_state_dict_with_metadata` is conflated with `save_model_with_metadata` in runner. There is not a standard format for saving a model vs state dict currently there is excessive delegation ie: `save_model_with_metadata` in runner.py calls `save_state_dict_with_metadata`.

zphang:
Mind taking a look at the changes again? I think the updates addressed some of your comments.

jeswan (approved):
Thanks for addressing my comments! Looks cleaner!

HarshTrivedi:
@zphang @jeswan It seems that `self.save_last_model` flag isn't being used anywhere. The done_training method *always* calls this method: `save_last_model_with_metadata`.

zphang:
Thanks! https://github.com/nyu-mll/jiant/pull/1243

PR #1225, opened 2020-11-01
Fix guids handling

zphang:


jiant/tasks/evaluate/core.py
```diff
@@ -73,7 +73,7 @@ def update(self, batch_logits, batch_loss, batch, batch_metadata):
         self.logits_list.append(batch_logits)
         batch_guid = batch_metadata.get("guid")
         if batch_guid is not None:
-            self.guid_list.extend(batch_guid)
+            self.guid_list.append(batch_guid)
```
jeswan, line 76: @HaokunLiu changed this line a few days ago. Did this change result in a bug?

jeswan: https://github.com/nyu-mll/jiant/pull/1197

zphang: Yes. `self.guid_list` should be a list of lists (which gets converted to an array with `np.concatenate`), and Haokun's change leads to an error when writing predictions.
Let me think about how best to test this.

jeswan: @HaokunLiu what error were you seeing which caused you to change this line?

HaokunLiu: Sorry, I misunderstood the design of this variable. 
Could you replace guid = accumulator.guid_list in MCTACOEvaluationScheme with guid = accumulator.get_guids() ?

jeswan (approved):

PR #1232, opened 2020-11-11
SQuAD update

zphang:
1. Fixes SQuAD v2 task config to properly supply kwargs
2. Update SQuAD tokenization logic based on https://github.com/huggingface/transformers/pull/7387 (and subsequently refactored in https://github.com/huggingface/transformers/pull/7970)

The summary for 2 is that previously (including in the Hugging Face implementation), the context was being tokenized per-word rather than as a string. This is problematic for tokenizers that treat start-of-word tokens differently. 

Although the cited PR states that there is a large performance improvement from this fix, I did not observe this myself in my testing (both in `jiant` as well as in the `transformers` examples). Both versions appeared to perform comparably.

For `jiant` users, this will impact tokenization, but training a fresh model for either old or new tokenization should both work. I recommended deleting and recreating any SQuAD-based caches. However, if you consistently use only the old tokenization caches, that should be fine too.

### DIFF
--- a/jiant/tasks/lib/templates/squad_style/core.py
+++ b/jiant/tasks/lib/templates/squad_style/core.py
@@ -21,6 +21,9 @@ from jiant.utils.display import maybe_tqdm
 
 import logging
 
+# Store the tokenizers which insert 2 separators tokens
+MULTI_SEP_TOKENS_TOKENIZERS_SET = {"roberta", "camembert", "bart"}
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


jiant/tasks/lib/templates/squad_style/core.py
```diff
@@ -21,6 +21,9 @@
 
 import logging
 
+# Store the tokenizers which insert 2 separators tokens
+MULTI_SEP_TOKENS_TOKENIZERS_SET = {"roberta", "camembert", "bart"}
```
jeswan, line 25: Thoughts on moving tokenizer specific wrapping constants to a dedicated file (ext/tokenizers/constants.py or similar)? Worried about the logic being scattered throughout files.

zphang: This constant comes from Transformers: https://github.com/huggingface/transformers/blob/969859d5f67c7106de4d1098c4891c9b03694bbe/src/transformers/data/processors/squad.py#L17

Unfortunately, it comes in a later version than what we're currently requiring, so we can't currently import it (unless we bump the version, I'd like to wait until after Fall to do it because of the v3.0.2 issue). The upside is that it's from the same file that all the rest of the SQuAD code comes from, so it'll be consistent for anyone looking up update the SQuAD implementation in the future.

jeswan (approved):

Issue #1193, opened 2020-10-14
Add delegate_load option for *only* loading the encoder


zphang:

jeswan:
Only load shared encoder and do not load any of the heads.

PR #1240, opened 2020-11-28
Load only encoder weights

zphang:
#1193

I also found that loading only encoder weights (and randomly initializing the task heads), because of the final learned representation, it is not uncommon even for randomly initialized task heads to have very strong task performance.


jeswan (approved):
Thanks! How were these changes tested?

zphang:
This was tested by loading an RTE model with `all` and `encoder_only`. The former has reproducible performance, the latter is random (though occasionally scores very well, see note in PR description).

PR #1246, opened 2020-12-15
Namespace default cache dir by model_type.

HarshTrivedi:
The default cache directory should be namespaced by model-type so that when 2 different models are sequentially run on the same task, the latter doesn't use cached inputs from previous model.


jeswan:
Thanks for opening this! I need to run the example notebooks before merging.

jeswan (approved):

PR #1256, opened 2021-01-04
Truncate MCQ inputs from the start and not end

HarshTrivedi:
In MCQ setting, the default truncation shouldn't be from the end. It has a risk of removing the question altogether. It's a better default to truncate from start so that if anything part of the context gets truncated but not the question.


zphang:
I am okay with this. Ideally, I would prefer this be configurable at the task level, but I'm not seeing a currently easy way to do this without greater code duplication / changing the current abstractions, we are more likely than not to redesign the abstractions altogether anyway. I do, however, agree that this is the sensible default. @jeswan thoughts?

(@HarshTrivedi run `black` to pass the tests.)

jeswan:
I agree that this is a sensible default. Thanks @HarshTrivedi!

jiant/tasks/lib/templates/multiple_choice.py
```diff
@@ -65,6 +65,7 @@ def featurize(self, tokenizer, feat_spec):
             prompt, choice = truncate_sequences(
                 tokens_ls=[self.prompt, choice],
                 max_length=feat_spec.max_seq_length - special_tokens_count,
+                truncate_end=False
```
jeswan, line 68:
```suggestion
                truncate_end=False,
```

jeswan (approved):

PR #1260, opened 2021-01-05
Switch export_model to use AutoModel and AutoTokenizer

jeswan:
This PR refactors `export_model` to use `AutoModel` and `AutoTokenizer`. It must be approved _after_ the tests for `export_model` have been merged (#1259). This PR reduces is part of the effort to reduce the `if-else` statements regarding model types in our codebase. `model_tokenizer_path` has been removed and replaced by `hf_pretrained_model_name_or_path`.


zphang:
If I'm reading this right, this PR needs to be updated to pass the new tests, right? `export_model` has a different signature I think.

jiant/shared/model_setup.py
```diff
@@ -5,43 +5,6 @@
...
-        ModelArchitectures.BART,
-        ModelArchitectures.MBART,
-        ModelArchitectures.ELECTRA,
-    ]:
-        do_lower_case = False
-    elif model_arch in [ModelArchitectures.ALBERT]:
-        do_lower_case = True
-    else:
-        raise RuntimeError(str(tokenizer_class))
-    tokenizer = tokenizer_class.from_pretrained(tokenizer_path, do_lower_case=do_lower_case)
```
jeswan, line 41: @zphang I am concerned I am not handling the `do_lower_case` gracefully. Do you think relying on AutoTokenizers is fine in this situation?

jiant/proj/main/modeling/model_setup.py
```diff
@@ -17,7 +17,8 @@
 
 
 def setup_jiant_model(
-    model_type: str,
+    model_type,
+    hf_pretrained_model_name: str,
```
zphang, line 21: Worth clarifying why we need both (see below)

jiant/proj/main/modeling/model_setup.py
```diff
@@ -38,7 +39,7 @@ def setup_jiant_model(
     """
     model_arch = ModelArchitectures.from_model_type(model_type)
     transformers_class_spec = TRANSFORMERS_CLASS_SPEC_DICT[model_arch]
-    tokenizer = model_setup.get_tokenizer(model_type=model_type, tokenizer_path=tokenizer_path)
+    tokenizer = transformers.AutoTokenizer.from_pretrained(hf_pretrained_model_name)
```
zphang, line 42: - Is `tokenizer_path` still being used?
- In some rare cases, we use tokenizers that aren't the pretrained HF ones. One option is the use `.from_pretrained`, but ensure that we also handle paths, like HF's `.from_pretrained` functions do.
- (On that note, consider streamlining the model loading/exporting to skip exporting if we're using the basic case of fine-tuning from the standard pretrained models)

jiant/shared/model_setup.py
zphang: I wonder if it is necessary at all, actually. `.from_pretrained` ought to handle this.

jiant/proj/main/modeling/model_setup.py
jeswan: - `tokenizer_path` is no longer being used. Thanks for pointing this out.
- This should already be supported but missing documentation. I'll add that in.
- I completely agree we need to do this...was planning to do that in a different PR since the refactoring was getting convoluted when including that change in this PR

jiant/proj/main/modeling/model_setup.py
jeswan: Good eye. This was an awkward area of the code. I'll take another look at the documentation but the situation is as follows: 
- We can get `model_type` from HF's config object
- We can generate the config object from `hf_pretrained_model_name`
- We cannot get the `model_type` from HF's model object (I need to check this again)
- It seemed awkward to generate a config object with `AutoConfig` every time I wanted `model_type`
It seems like the cleanest way forward it to just pass in `hf_pretrained_model_name` and get the model_type from an AutoClass method.

jeswan: Changed to only use `hf_pretrained_model_name_or_path`

zphang (approved):
Nothing jumps out to me, also relatively low risk to merge now, I think.

Potentially one thing you might want to look at is if updating to transformers 4+ will be complicated by any of this.

jeswan:
@zphang A few notes on how this PR will be affected by Transformers v4:
- AutoTokenizer will default to using the fast (rust) tokenizers. I am currently planning on switching to fast tokenizers as part of the update to Transformers v4.
- slow versions of the following tokenizers will not be available since SentencePiece is removed from required dependencies:
  - XLNetTokenizer
  - AlbertTokenizer
  - CamembertTokenizer
  - MBartTokenizer
  - PegasusTokenizer
  - T5Tokenizer
  - ReformerTokenizer
  - XLMRobertaTokenizer
- Side note: Model forward will have `return_dict` set to `True` as default and will require `jiant` forward passes to use `False` if using tuple unpacking.

zphang:
Sounds good, moving remaining discussion to other PR.

PR #1268, opened 2021-01-20
Switch to task model/head factories instead of embedded if-else statements

jeswan:
The goal of these changes is to streamline adding a model to `jiant`. This PR implements factories to create task models and task heads. In addition, `JiantTransformersModelWrapper` has been added to abstract and inconsistencies in HuggingFace's Transformers library.


jiant/proj/main/modeling/heads.py
```diff
@@ -4,14 +4,84 @@
...
+from jiant.tasks.core import TaskTypes
+
 
 """
 In HuggingFace/others, these heads differ slightly across different encoder models.
 We're going to abstract away from that and just choose one implementation.
 """
 
 
+class JiantHeadFactory:
```
jeswan, line 19: New task heads added here

zphang: Thinking aloud:
- If we're going to make this a class method, we might as well add a registry as well (to support potentially user-created heads that can be dynamically added)
- Would it be better to combine the head+taskmodels, rather than have separate factories for head and task model?
- The EmbeddingModel logic can be split off into its own factory, potentially. Fine leaving it for now.

jiant/proj/main/modeling/heads.py
```diff
@@ -205,3 +275,46 @@ def forward(self, unpooled):
...
+                hidden_size=hidden_size,
+                vocab_size=encoder.config.vocab_size,
+                layer_norm_eps=encoder.config.layer_norm_eps,
+            )
+        elif model_arch in (
+            ModelArchitectures.BART,
+            ModelArchitectures.MBART,
+            ModelArchitectures.ELECTRA,
+        ):
+            raise NotImplementedError()
```
zphang, line 316: f"MLM not supported for {model_arch}"

jiant/proj/main/modeling/model_setup.py
```diff
@@ -162,7 +156,7 @@ def load_encoder_from_transformers_weights(
     remainder_weights_dict = {}
     load_weights_dict = {}
     model_arch = ModelArchitectures.from_encoder(encoder=encoder)
-    encoder_prefix = MODEL_PREFIX[model_arch] + "."
+    encoder_prefix = model_arch.value + "."
```
zphang, line 159: `f"{model_arch.value}."

jiant/proj/main/modeling/primary.py
```diff
@@ -85,3 +89,31 @@ def wrap_jiant_forward(
...
+class JiantBERTModelWrapper(JiantTransformersModelWrapper):
+    def __init__(self, baseObject):
+        super().__init__(baseObject)
+        self.hf_pretrained_encoder_with_pretrained_head = transformers.BertForPreTraining
+
+    def get_hidden_size(self):
+        return self.config.hidden_size
+
+    def get_hidden_dropout_prob(self):
+        return self.config.hidden_dropout_prob
```
zphang, line 119: Can you add some docstrings to these? Their purpose isn't immediately clear.

zphang (changes requested):
Some comments for organization of logic, but overall no opposition.

jiant/proj/main/modeling/heads.py
jeswan: - Great idea to have dynamically supported user heads. I'll look into the details of this (will set up a meeting for feedback).
- I was debating this as well. I am not set on this implementation but I chose this since it seemed better to isolate the components to prevent unwanted coupling (although they are closely linked already)
- Thanks

jiant/proj/main/modeling/model_setup.py
jeswan: seems a less readable (the period could be overlooked)

jiant/proj/main/modeling/primary.py
jeswan: Can do. Just wanted to run this idea by you. This basically wraps Transformer models so that any inconsistencies can be abstracted from the model. For example, `hidden_size` has different member variable names in BERT and BART.

jiant/shared/model_resolution.py
```diff
@@ -27,86 +27,25 @@ def from_model_type(cls, model_type: str):
             Model architecture associated with the provided shortcut name.
 
         """
-        if model_type.startswith("bert"):
+        if model_type == "bert":
```
jeswan, line 30: reminder to self to change this to use `BERT` enum

tests/tasks/lib/test_mlm_premasked.py
```diff
@@ -35,7 +35,7 @@ def test_tokenization_and_featurization():
     data_row = tokenized_example.featurize(
         tokenizer=tokenizer,
         feat_spec=model_resolution.build_featurization_spec(
-            model_type="roberta-base", max_seq_length=16,
+            model_type="roberta", max_seq_length=16,
```
jeswan, line 38: reminder to self to change this to use `ROBERTA` enum

jiant/proj/main/modeling/model_setup.py
zphang: Ah, fair enough.

jiant/proj/main/modeling/heads.py
zphang: My sense is that as long as the heads have their own standlone implementations, that's sufficiently isolated and tying the whole task-model construction should be fine.

jiant/proj/main/modeling/primary.py
zphang: I think abstracting the hidden_size/dropout prob is headed in the right direction. One thing I'd like to throw in is that it might not be worth making a different class for each model type. Can discuss more tomorrow.

jiant/proj/main/modeling/taskmodels.py
```diff
@@ -9,44 +11,94 @@
 from jiant.proj.main.components.outputs import LogitsOutput, LogitsAndLossOutput
 from jiant.utils.python.datastructures import take_one
 from jiant.shared.model_resolution import ModelArchitectures
+from jiant.tasks.core import TaskTypes
+from jiant.utils.python.datastructures import take_one
```
jeswan, line 15:
```suggestion
```

PR #1283, opened 2021-02-25
Refactor get_output_from_encoder() to be member of JiantTaskModel

jeswan:
These changes move the if-else statements (for `ModelArchitectures`) in `get_output_from_encoder()` to the newly added `JiantTransformersModel` when using `__call__`.

### DIFF
--- a/jiant/proj/main/modeling/primary.py
+++ b/jiant/proj/main/modeling/primary.py
@@ -1,9 +1,12 @@
 import abc
 
+from dataclasses import dataclass
+from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import Union
-from typing import Callable
 
+import torch
 import torch.nn as nn
 import transformers
 
@@ -12,6 +15,13 @@ import jiant.tasks as tasks
 
 from jiant.proj.main.components.outputs import construct_output_from_dict
 from jiant.shared.model_resolution import ModelArchitectures
+
+
+@dataclass
+class JiantModelOutput:
+    pooled: torch.Tensor
+    unpooled: torch.Tensor
+    other: Any = None
 
 
 class JiantModel(nn.Module):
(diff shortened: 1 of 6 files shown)


jiant/proj/main/modeling/primary.py
```diff
@@ -154,6 +164,19 @@ def get_hidden_size(self):
     def get_hidden_dropout_prob(self):
         return self.config.hidden_dropout_prob
 
+    def __call__(self, input_ids, segment_ids, input_mask, output_hidden_states=True):
```
zphang, line 167: I might have missed this in the previous PR - what's the reasons for modifying the `__class__` and `__call__` of the object? I would prefer a more transparent interface (e.g. `.encode`)

jiant/proj/main/modeling/taskmodels.py
```diff
@@ -1,22 +1,22 @@
...
 from jiant.utils.python.datastructures import take_one
-from jiant.shared.model_resolution import ModelArchitectures
+
 from jiant.tasks.core import TaskTypes
 
 
 class JiantTaskModelFactory:
     """This factory is used to create task models bundling the task,
-       encoder, and task head within the task model.
+       jiant_transformers_model, and task head within the task model.
```
zphang, line 19: Likewise, is there a reason for changing from `encoder` to `jiant_transformers_model`? (It looks like it's still using the encoder, not the whole model, so it's a little unclear to me)

jeswan: The naming change was meant to clarify that it is a `JiantTransformersModel` and not HF's `encoder`. I thought the new name would be more descriptive.

jiant/proj/main/modeling/primary.py
jeswan: It was meant to mimic HF Transformer's forward but I am not attached to any specific interface.

PR #1290, opened 2021-03-11
Move tokenization logic to central JiantModelTransformers method

jeswan:
This PR refactors `normalize_tokenizations()` by moving the functions to `JiantTransformerModel`s. Although the location of this logic is not ideal, it was the easiest way to have a central location for editing new models/tokenizers. This PR also adds `resolve_model_arch_tokenizer(tokenizer)` to map the tokenizers to model_archs.

### DIFF
--- a/jiant/shared/model_resolution.py
+++ b/jiant/shared/model_resolution.py
@@ -64,7 +64,6 @@ def resolve_model_arch_tokenizer(tokenizer):
         ModelArchitecture
 
     """
-    print(TOKENIZER_CLASS_DICT.inverse)
     assert len(TOKENIZER_CLASS_DICT.inverse[tokenizer.__class__]) == 1
     return TOKENIZER_CLASS_DICT.inverse[tokenizer.__class__][0]
 
--- a/jiant/shared/model_resolution.py
+++ b/jiant/shared/model_resolution.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
+from jiant.utils.python.datastructures import BiDict
 
 import transformers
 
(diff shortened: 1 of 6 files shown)


jiant/shared/model_resolution.py
```diff
@@ -51,6 +54,21 @@ def resolve_tokenizer_class(model_type):
...
+    """Get the model architecture for a given tokenizer.
+
+    Args:
+        tokenizer
+
+    Returns:
+        ModelArchitecture
+
+    """
+    print(TOKENIZER_CLASS_DICT.inverse)
```
jeswan, line 67:
```suggestion
```

jiant/utils/tokenization_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    """Applies a beginning of word (BoW) marker to every token in the tokens sequence."""
+    return [bow_tag + t for t in tokens]
+
+
+def eow_tag_tokens(tokens: Sequence[str], eow_tag: str = "</w>"):
+    """Applies a end of word (EoW) marker to every token in the tokens sequence."""
+    return [t + eow_tag for t in tokens]
+
+
+def _process_wordpiece_tokens(tokens: Sequence[str]):
```
zphang, line 16: Let's make these public

zphang (commented):
1) I noticed that ELECTRA doesn't have a normalize tokenizations implementation. Can we add an abstractmethod to `JiantTransformersModel` and a `raise NotImplemented` for ELECTRA? (It should be possible to add the implementation for ELECTRA, but I'm okay with leaving it as a future to-do)
2) I was wondering if it might be a better idea to pull the normalize tokenization logic out into individual functions (in some cases, similar models have the same logic), but I can also see how that might be excessive delegation. I don't feel strongly either way, what do you think?

jeswan:
> * I noticed that ELECTRA doesn't have a normalize tokenizations implementation. Can we add an abstractmethod to `JiantTransformersModel` and a `raise NotImplemented` for ELECTRA? (It should be possible to add the implementation for ELECTRA, but I'm okay with leaving it as a future to-do)
> * I was wondering if it might be a better idea to pull the normalize tokenization logic out into individual functions (in some cases, similar models have the same logic), but I can also see how that might be excessive delegation. I don't feel strongly either way, what do you think?

Comments implemented! For point 2 above, discussed offline to refactor if there is additional duplicate code in the future.

PR #1296, opened 2021-03-19
Fix tasks circular import

jeswan:
Importing `BatchMixin` in primary previously imported all tasks resulting in a circular import with `tokenization_normalization()`. This PR fixes the circular imports by removing `import *` in tasks `__init__.py`.

### DIFF
--- a/jiant/tasks/evaluate/core.py
+++ b/jiant/tasks/evaluate/core.py
@@ -1,26 +1,35 @@
 import itertools
 import json
+
 from dataclasses import dataclass
 
 import numpy as np
 import pandas as pd
 import seqeval.metrics as seqeval_metrics
 import torch
-from sklearn.metrics import f1_score, matthews_corrcoef
-from scipy.stats import pearsonr, spearmanr
-from typing import Dict, List
+
+from scipy.stats import pearsonr
+from scipy.stats import spearmanr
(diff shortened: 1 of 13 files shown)


jiant/tasks/evaluate/core.py
```diff
@@ -1,26 +1,116 @@
...
-import jiant.tasks.lib.templates.squad_style.utils as squad_style_utils
-import jiant.tasks.lib.mlqa as mlqa_lib
+
 import jiant.tasks.lib.bucc2018 as bucc2018_lib
+import jiant.tasks.lib.mlqa as mlqa_lib
 import jiant.tasks.lib.tatoeba as tatoeba_lib
+import jiant.tasks.lib.templates.squad_style.core as squad_style
+import jiant.tasks.lib.templates.squad_style.utils as squad_style_utils
+
+from jiant.tasks.retrieval import AbductiveNliTask
```
zphang, line 26: As discussed, could we remove the individual imports and just do `import jiant.tasks.retrieval as tasks retrieval`?

zphang (changes requested):

jiant/tasks/evaluate/core.py
jeswan: Switched to this style of import.

Issue #1348, opened 2022-03-25
Unable to execute run_simple() with different models of the same type


TimDettmers:
**Describe the bug**

When one uses run_simple() with different models of the same type `roberta-base` and `roberta-large` the run crashes because the code assumes they are the same model because weights are saved under `hf_config.model_type` (instead of `args.hf_pretrained_model_name_or_path.`). As such, the code tries to load incompatible weights and crashes.

**To Reproduce**
1. Install jiant
2. Run the simple example in README
3. Change the model in the sample from 'roberta-base` to `roberta-large`

**Expected behavior**
One should be able to run run_simple() with different models of the same type.

**Screenshots**

**Additional context**
Solution: The `hf_config.model_type` should be used for caching tokenizer / tasks. The `args.hf_pretrained_model_name_or_path` for  the weights.

PR #1349, opened 2022-03-25
Bugfix for simple_run() multiple models with the same type.

TimDettmers:
This is a fix for #1348.

While the tokenization etc is still cached for each model type, now multiple models can be run for the same model type, for example caching for `roberta` while both `roberta-base` and `roberta-large` can be run.


zphang:
Thanks!

